PR #2, opened 2016-03-09
T10617

joaquimrocha:
Adds a plugin to view, launch and remove EOS app bundles to the App Center.

### DIFF
--- a/src/plugins/Makefile.am
+++ b/src/plugins/Makefile.am
@@ -132,7 +132,7 @@ libgs_plugin_xdg_app_reviews_la_CFLAGS = $(GS_PLUGIN_CFLAGS) $(WARN_CFLAGS)
 endif
 
 if HAVE_EAM
-libgs_plugin_eos_legacy_bundles_la_SOURCES = gs-plugin-eos-legacy-bundles.c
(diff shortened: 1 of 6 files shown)


src/plugins/Makefile.am
```diff
@@ -223,6 +234,23 @@ libgs_plugin_packagekit_proxy_la_LIBADD = $(GS_PLUGIN_LIBS)
...
+eos-app-manager-service.c: eos-app-manager-service.h
+    @: # generated as a side-effect
+
+EXTRA_DIST = $(top_srcdir)/data/eam-service-interface.xml
+CLEANFILES = eos-app-manager-service.c eos-app-manager-service.h
+BUILT_SOURCES = eos-app-manager-service.c eos-app-manager-service.h
+endif
+
 check_PROGRAMS =					\
 	gs-self-test
```
ebassi: You removed `gs-self-test` from `TESTS`, but I think you want to remove it from `check_PROGRAMS` as well, since that's the target used for `make check` and `make distcheck`.

ebassi:
Looks generally good to me. I think you could use `g_autoptr()` in various places, but it's not something blocking.

PR #3, opened 2016-03-11
Depend on a new enough json-glib to drop the compat defines

joaquimrocha:
Cherry-pick from upstream so g-s compiles with json-glib >= 1.1.2.


ebassi:
CR pass.

PR #4, opened 2016-03-17
Disable reviews and shell-extensions when building

joaquimrocha:
Reviews and GNOME Shell Extensions should not be part of the UI in Endless's use of GNOME Software.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+gnome-software (3.19.92-0) master; urgency=medium
+
+  * New upstream release
+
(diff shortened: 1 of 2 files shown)


dbnicholson:
This looks fine. Go ahead and tag this commit and push.

dbnicholson:
FYI, I suggested pushing this branch to debian-master rather than merging it. Then github doesn't create a merge commit and artificially move you off of your tag.

joaquimrocha:
Hi @dbnicholson , I didn't merge this through GitHub. I merged it in the terminal and the branch was rebased so no merge commit was made. FTR @cosimoc found out that the new tag 3.19.92 was not in upstream's *master* branch, which I hadn't noticed so I removed the tags and the debian changelog entry.

PR #8, opened 2016-04-28
Fix built plugins when generating the Debian package

joaquimrocha:
I have noticed that the xdg-app plugin is not shipped in the deb so I have enabled in debian/rules to see if that makes a difference (it should be built if xdg-app is available, which is a deb build dependency).

I have also disabled the ODRS plugin as we do not want to worry about ratings ATM.

https://phabricator.endlessm.com/T11467

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -4,6 +4,7 @@ GS_CONFIGURE_FLAGS = --libdir=/usr/lib \
 		--disable-silent-rules \
 		--disable-reviews      \
 		--disable-shell-extensions 	\
+		--disable-odrs			\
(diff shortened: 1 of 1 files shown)


debian/patches/fix-arch-on-32-bit.patch
```diff
@@ -0,0 +1,49 @@
+From 4611b33f9393214a39295cdde3c1ec5ad410b10d Mon Sep 17 00:00:00 2001
```
cosimoc: This patch should be applied in the main branch, instead of going through the debian packaging indirection.

PR #9, opened 2016-04-29
build: Set arch to 32 bit directly when the Debian architecture match…

joaquimrocha:
Fix xdg-app and appstream-glib getting the arch of the kernel which results in problems when the system has a 32 bit user space on top of a 64 bit kernel.


dbnicholson:
I know this is a downstream change, so it doesn't matter too much, but I don't think using dpkg-architecture from configure is appropriate.

Better would be to use `host_cpu` from `AC_CANONICAL_HOST` with a case on `i?86`. See http://www.gnu.org/software/autoconf/manual/autoconf.html#Canonicalizing.

joaquimrocha:
Thanks @dbnicholson , I will look into it.

PR #15, opened 2016-05-25
T11687

joaquimrocha:
This PR should be applied on top of:
https://github.com/endlessm/gnome-software/pull/14

See https://phabricator.endlessm.com/T11687


cosimoc:
@joaquimrocha It feels to me that there should be a better default - why does gnome-software think that it could remove those applications in the first place? Either way this works for us, so I merged it.

joaquimrocha:
@cosimoc , g-s assumes it can remove an installed application unless it's marked as compulsory. I tried removing one of the input methods, it disappears but just as I suspected, after relaunching g-s it is there again.

PR #17, opened 2016-06-10
T11731

joaquimrocha:
See https://phabricator.endlessm.com/T11731

### DIFF
--- a/src/gs-utils.c
+++ b/src/gs-utils.c
@@ -357,4 +357,60 @@ gs_utils_unlink (const gchar *filename, GError **error)
 	return TRUE;
 }
 
+static gboolean
+gs_utils_rmtree_real (const gchar *directory, GError **error)
+{
+	const gchar *filename;
+	g_autoptr(GDir) dir = NULL;
+
+	/* try to open */
+	dir = g_dir_open (directory, 0, error);
+	if (dir == NULL)
+		return FALSE;
+
+	/* find each */
+	while ((filename = g_dir_read_name (dir))) {
+		g_autofree gchar *src = NULL;
+		src = g_build_filename (directory, filename, NULL);
+		if (g_file_test (src, G_FILE_TEST_IS_DIR) &&
+		    !g_file_test (src, G_FILE_TEST_IS_SYMLINK)) {
+			if (!gs_utils_rmtree_real (src, error))
+				return FALSE;
+		} else {
+			if (g_unlink (src) != 0) {
+				g_set_error (error,
+					     GS_PLUGIN_ERROR,
+					     GS_PLUGIN_ERROR_FAILED,
+					     "Failed to delete: %s", src);
+				return FALSE;
+			}
+		}
+	}
+
+	if (g_rmdir (directory) != 0) {
+		g_set_error (error,
+			     GS_PLUGIN_ERROR,
+			     GS_PLUGIN_ERROR_FAILED,
+			     "Failed to remove: %s", directory);
+		return FALSE;
+	}
+	return TRUE;
+}
+
+/**
+ * gs_utils_rmtree:
+ * @directory: A full directory pathname to delete
+ * @error: A #GError, or %NULL
+ *
+ * Deletes a directory from disk and all its contents.
+ *
+ * Returns: %TRUE for success
+ **/
+gboolean
+gs_utils_rmtree (const gchar *directory, GError **error)
+{
+	g_debug ("recursively removing directory '%s'", directory);
+	return gs_utils_rmtree_real (directory, error);
+}
+
 /* vim: set noexpandtab: */
--- a/src/gs-utils.h
+++ b/src/gs-utils.h
@@ -63,7 +63,8 @@ GPermission	*gs_utils_get_permission	(const gchar	*id);
 gboolean	 gs_utils_strv_fnmatch		(gchar		**strv,
 						 const gchar	*str);
 GDesktopAppInfo *gs_utils_get_desktop_app_info	(const gchar	*id);
-
+gboolean	 gs_utils_rmtree		(const gchar	*directory,
+						 GError		**error);
 G_END_DECLS
 
 #endif /* __GS_UTILS_H */
--- a/src/plugins/Makefile.am
+++ b/src/plugins/Makefile.am
@@ -73,6 +73,7 @@ endif
 if HAVE_FLATPAK
 plugin_LTLIBRARIES += libgs_plugin_flatpak-system.la
 plugin_LTLIBRARIES += libgs_plugin_flatpak-user.la
+plugin_LTLIBRARIES += libgs_plugin_external-apps.la
 endif
 
 if HAVE_OSTREE
@@ -192,6 +193,20 @@ libgs_plugin_flatpak_user_la_SOURCES =			\
 libgs_plugin_flatpak_user_la_LIBADD = $(GS_PLUGIN_LIBS) $(FLATPAK_LIBS)
 libgs_plugin_flatpak_user_la_LDFLAGS = -module -avoid-version
 libgs_plugin_flatpak_user_la_CFLAGS = $(GS_PLUGIN_CFLAGS) $(WARN_CFLAGS)
+
+libgs_plugin_external_apps_la_SOURCES =		\
+	gs-appstream.c				\
+	gs-appstream.h				\
+	gs-flatpak.c				\
+	gs-flatpak.h				\
+	gs-flatpak-symlinks.c			\
+	gs-flatpak-symlinks.h			\
+	gs-plugin-external-apps.c
+
+libgs_plugin_external_apps_la_LIBADD = $(GS_PLUGIN_LIBS) $(FLATPAK_LIBS) $(JSON_GLIB_LIBS)
+libgs_plugin_external_apps_la_LDFLAGS = -module -avoid-version
+libgs_plugin_external_apps_la_CFLAGS = $(GS_PLUGIN_CFLAGS) $(JSON_GLIB_CFLAGS) $(WARN_CFLAGS)
+
 endif
 
 if HAVE_OSTREE
--- a/src/plugins/gs-flatpak.c
+++ b/src/plugins/gs-flatpak.c
@@ -1658,3 +1658,27 @@ gs_flatpak_is_installed (GsFlatpak *self,
 
 	return ref != NULL;
 }
+
+GPtrArray *
+gs_flatpak_get_remotes_names (GsFlatpak *self,
+			      GCancellable *cancellable,
+			      GError **error)
+{
+	GPtrArray *names = NULL;
+	guint i;
+	g_autoptr(GPtrArray) xremotes = NULL;
+
+	xremotes = flatpak_installation_list_remotes (self->installation,
+						      cancellable,
+						      error);
+	if (!xremotes)
+		return NULL;
+
+	names = g_ptr_array_new_full (xremotes->len, g_free);
+	for (i = 0; i < xremotes->len; i++) {
+		FlatpakRemote *xremote = g_ptr_array_index (xremotes, i);
+		g_ptr_array_add (names, g_strdup (flatpak_remote_get_name (xremote)));
+	}
+
+	return names;
+}
--- a/src/plugins/gs-flatpak.c
+++ b/src/plugins/gs-flatpak.c
@@ -1630,3 +1630,31 @@ gs_flatpak_get_prefix (GsFlatpak *self)
 
 	return NULL;
 }
+
+gboolean
+gs_flatpak_is_installed (GsFlatpak *self,
+			 GsApp *app,
+			 GCancellable *cancellable,
+			 GError **error)
+{
+	g_autoptr(FlatpakInstalledRef) ref;
+	FlatpakRefKind kind;
+	const char *name;
+	const char *arch;
+	const char *branch;
+
+	kind = gs_app_get_flatpak_kind (app);
+	name = gs_app_get_flatpak_name (app);
+	arch = gs_app_get_flatpak_arch (app);
+	branch = gs_app_get_flatpak_branch (app);
+
+	ref = flatpak_installation_get_installed_ref (self->installation,
+						      kind,
+						      name,
+						      arch,
+						      branch,
+						      cancellable,
+						      error);
+
+	return ref != NULL;
+}
--- a/src/plugins/gs-flatpak.c
+++ b/src/plugins/gs-flatpak.c
@@ -1617,3 +1617,16 @@ gs_flatpak_new (GsPlugin *plugin, GsFlatpakScope scope)
 	self->plugin = g_object_ref (plugin);
 	return GS_FLATPAK (self);
 }
+
+const char *
+gs_flatpak_get_prefix (GsFlatpak *self)
+{
+	if (self->scope == GS_FLATPAK_SCOPE_USER)
+		return GS_FLATPAK_USER_PREFIX;
+	else if (self->scope == GS_FLATPAK_SCOPE_SYSTEM)
+		return GS_FLATPAK_SYSTEM_PREFIX;
+
+	g_assert_not_reached ();
+
+	return NULL;
+}
--- a/src/plugins/gs-flatpak.c
+++ b/src/plugins/gs-flatpak.c
@@ -1682,3 +1682,9 @@ gs_flatpak_get_remotes_names (GsFlatpak *self,
 
 	return names;
 }
+
+gboolean
+gs_flatpak_app_is_runtime (GsApp *app)
+{
+	return gs_app_get_flatpak_kind (app) == FLATPAK_REF_KIND_RUNTIME;
+}
--- a/src/plugins/gs-flatpak.c
+++ b/src/plugins/gs-flatpak.c
@@ -1208,6 +1208,7 @@ gs_flatpak_app_install (GsFlatpak *self,
 			GError **error)
 {
 	g_autoptr(FlatpakInstalledRef) xref = NULL;
+	GError *local_error = NULL;
 
 	/* only process this app if was created by this plugin */
 	if (g_strcmp0 (gs_app_get_management_plugin (app),
@@ -1276,7 +1277,8 @@ gs_flatpak_app_install (GsFlatpak *self,
 							    gs_app_get_local_file (app),
 							    gs_flatpak_progress_cb,
 							    app,
-							    cancellable, error);
+							    cancellable,
+							    &local_error);
 	} else {
 		g_debug ("installing %s", gs_app_get_id (app));
 		xref = flatpak_installation_install (self->installation,
@@ -1286,9 +1288,20 @@ gs_flatpak_app_install (GsFlatpak *self,
 						     gs_app_get_flatpak_arch (app),
 						     gs_app_get_flatpak_branch (app),
 						     gs_flatpak_progress_cb, app,
-						     cancellable, error);
+						     cancellable, &local_error);
 	}
 	if (xref == NULL) {
+		/* We may have failed to install because the app is already
+		 * installed */
+		if (g_error_matches (local_error, FLATPAK_ERROR,
+				     FLATPAK_ERROR_ALREADY_INSTALLED)) {
+			gs_app_set_state (app, AS_APP_STATE_INSTALLED);
+			return TRUE;
+		}
+
+		if (local_error)
+			g_propagate_error (error, local_error);
+
 		gs_app_set_state_recover (app);
 		return FALSE;
 	}
--- a/src/plugins/gs-plugin-eos.c
+++ b/src/plugins/gs-plugin-eos.c
@@ -26,6 +26,8 @@
 #include <glib/gi18n.h>
 #include <gs-plugin.h>
 
+#include "gs-flatpak.h"
+
 /*
  * SECTION:
  * Plugin to improve GNOME Software integration in the EOS desktop.
@@ -203,10 +205,10 @@ gs_plugin_eos_update_app_shortcuts_info (GsPlugin *plugin,
 static gboolean
 gs_plugin_eos_app_is_flatpak (GsApp *app)
 {
-	const char *app_manager;
-	app_manager = gs_app_get_management_plugin (app);
+	const char *id = gs_app_get_id (app);
 
-	return g_strcmp0 (app_manager, "flatpak") == 0;
+	return id && (g_str_has_prefix (id, GS_FLATPAK_SYSTEM_PREFIX ":") ||
+		      g_str_has_prefix (id, GS_FLATPAK_USER_PREFIX ":"));
 }
 
 static void
--- a/src/plugins/gs-plugin-external-apps.c
+++ b/src/plugins/gs-plugin-external-apps.c
@@ -1,0 +1,782 @@
+/* -*- Mode: C; tab-width: 8; indent-tabs-mode: t; c-basic-offset: 8 -*-
+ *
+ * Copyright (C) 2016 Endless Mobile, Inc
+ *
+ * Author: Joaquim Rocha <jrocha@endlessm.com>
+ *
+ * Licensed under the GNU General Public License Version 2
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include <config.h>
+#include <flatpak.h>
+#include <glib.h>
+#include <gnome-software.h>
+#include <json-glib/json-glib.h>
+#include <string.h>
+
+#include "gs-flatpak.h"
+
+#define EXTERNAL_ASSETS_SPEC_VERSION 1
+#define JSON_SPEC_KEY "spec"
+#define JSON_RUNTIME_KEY "runtime"
+#define JSON_RUNTIME_NAME_KEY "name"
+#define JSON_RUNTIME_URL_KEY "url"
+
+#define METADATA_URL "GnomeSoftware::external-app::url"
+#define METADATA_HEADLESS_APP "GnomeSoftware::external-app::headless-app"
+#define METADATA_BUILD_DIR "GnomeSoftware::external-app::build-dir"
+#define METADATA_EXTERNAL_ASSETS "flatpak-3rdparty::external-assets"
+
+#define TMP_ASSSETS_PREFIX "gs-external-apps"
+
+struct GsPluginData {
+	GsFlatpak	*flatpak;
+	char		*runtimes_build_dir;
+};
+
+static gboolean flatpak_remote_delete (const char *repo_name, GError **error);
+
+static void
+remove_runtimes_build_dir (GsPlugin *plugin)
+{
+	g_autoptr(GError) error = NULL;
+	GsPluginData *priv = gs_plugin_get_data (plugin);
+
+	if (gs_utils_rmtree (priv->runtimes_build_dir, &error))
+		return;
+
+	if (!g_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_NOENT)) {
+		g_debug ("Cannot remove previously created external apps build "
+			 "dir '%s': %s", priv->runtimes_build_dir,
+			 error->message);
+	}
+}
+
+static void
+remove_ext_apps_remotes (GsPlugin *plugin)
+{
+	GsPluginData *priv = gs_plugin_get_data (plugin);
+	guint i;
+	GPtrArray *names;
+
+	names = gs_flatpak_get_remotes_names (priv->flatpak, NULL, NULL);
+
+	if (!names)
+		return;
+
+	for (i = 0; i < names->len; i++) {
+		const char *name = g_ptr_array_index (names, i);
+		if (g_str_has_prefix (name, TMP_ASSSETS_PREFIX)) {
+			flatpak_remote_delete (name, NULL);
+		}
+	}
+
+	g_ptr_array_free (names, TRUE);
+}
+
+void
+gs_plugin_initialize (GsPlugin *plugin)
+{
+	GsPluginData *priv = gs_plugin_alloc_data (plugin, sizeof(GsPluginData));
+	const char *ext_apps_build_dir = g_get_user_cache_dir ();
+
+	priv->flatpak = gs_flatpak_new (plugin, GS_FLATPAK_SCOPE_USER);
+	priv->runtimes_build_dir = g_build_filename (ext_apps_build_dir,
+						     TMP_ASSSETS_PREFIX,
+						     NULL);
+
+	/* Run plugin before the flatpak ones because we need them to install
+	 * the app's headless part first */
+	gs_plugin_add_rule (plugin, GS_PLUGIN_RULE_RUN_BEFORE, "flatpak-system");
+	gs_plugin_add_rule (plugin, GS_PLUGIN_RULE_RUN_BEFORE, "flatpak-user");
+}
+
+void
+gs_plugin_destroy (GsPlugin *plugin)
+{
+	GsPluginData *priv = gs_plugin_get_data (plugin);
+
+	/* Remove the runtimes build directory to clean any contents eventually
+	 * left from previous builds */
+	remove_runtimes_build_dir (plugin);
+
+	g_clear_object (&priv->flatpak);
+	g_clear_object (&priv->runtimes_build_dir);
+}
+
+static gboolean
+app_is_flatpak (GsApp *app)
+{
+	const gchar *id = gs_app_get_id (app);
+	return id && (g_str_has_prefix (id, GS_FLATPAK_USER_PREFIX ":") ||
+		      g_str_has_prefix (id, GS_FLATPAK_SYSTEM_PREFIX ":"));
+}
+
+void
+gs_plugin_adopt_app (GsPlugin *plugin,
+		     GsApp *app)
+{
+	if (!app_is_flatpak (app) ||
+	    !gs_app_get_metadata_item (app, METADATA_EXTERNAL_ASSETS))
+		return;
+
+	g_debug ("Adopt '%s' as an external app", gs_app_get_id (app));
+
+	gs_app_set_management_plugin (app, gs_plugin_get_name (plugin));
+}
+
+gboolean
+gs_plugin_setup (GsPlugin *plugin,
+		 GCancellable *cancellable,
+		 GError **error)
+{
+	GsPluginData *priv = gs_plugin_get_data (plugin);
+	gboolean ret = gs_flatpak_setup (priv->flatpak, cancellable, error);
+
+	if (ret) {
+		/* Remove the runtimes build directories and remotes to clean
+		 * any contents eventually left from previous builds */
+		remove_runtimes_build_dir (plugin);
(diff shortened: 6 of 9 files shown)


src/plugins/gs-flatpak.c
```diff
@@ -1617,3 +1617,14 @@ gs_flatpak_new (GsPlugin *plugin, GsFlatpakScope scope)
...
+
+const char *
+gs_flatpak_get_prefix (GsFlatpak *self)
+{
+	if (self->scope == GS_FLATPAK_SCOPE_USER)
+		return GS_FLATPAK_USER_PREFIX;
+	else if (self->scope == GS_FLATPAK_SCOPE_SYSTEM)
+		return GS_FLATPAK_SYSTEM_PREFIX;
+
+	return "";
```
cosimoc: If there are no other scopes, should use a `g_assert_if_reached()` or similar perhaps?

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include <config.h>
+#include <json-glib/json-glib.h>
+#include <flatpak.h>
+#include <glib.h>
+#include <gnome-software.h>
+#include <string.h>
```
cosimoc: Nitpick: alphabetical order

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+gs_plugin_initialize (GsPlugin *plugin)
+{
+	GsPluginData *priv = gs_plugin_alloc_data (plugin, sizeof(GsPluginData));
+
+	priv->flatpak = gs_flatpak_new (plugin, GS_FLATPAK_SCOPE_USER);
+
+	/* Run plugin after the flatpak ones because we need them to install
+	 * the app's headless part first */
+	gs_plugin_add_rule (plugin, GS_PLUGIN_RULE_RUN_BEFORE, "flatpak-system");
+	gs_plugin_add_rule (plugin, GS_PLUGIN_RULE_RUN_BEFORE, "flatpak-user");
```
cosimoc: The comment says "run after flatpak" but the rule here would make it run before, no?

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+	const gchar *id = gs_app_get_id (app);
+	return id != NULL &&
+		(g_str_has_prefix (id, GS_FLATPAK_USER_PREFIX ":") ||
+		 g_str_has_prefix (id, GS_FLATPAK_SYSTEM_PREFIX ":"));
+}
+
+void
+gs_plugin_adopt_app (GsPlugin *plugin, GsApp *app)
+{
+	const gchar *id = gs_app_get_id (app);
```
cosimoc: You could use `gs_app_get_id()` directly in the `g_debug()` since it's the only place it's used and the string is const

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+	if (gs_app_is_flatpak (app)) {
```
cosimoc: Nitpick: prefer early return

src/plugins/Makefile.am
```diff
@@ -192,6 +193,20 @@ libgs_plugin_flatpak_user_la_SOURCES =			\
 libgs_plugin_flatpak_user_la_LIBADD = $(GS_PLUGIN_LIBS) $(FLATPAK_LIBS)
 libgs_plugin_flatpak_user_la_LDFLAGS = -module -avoid-version
 libgs_plugin_flatpak_user_la_CFLAGS = $(GS_PLUGIN_CFLAGS) $(WARN_CFLAGS)
+
+libgs_plugin_external_apps_la_SOURCES =		\
```
cosimoc: I wonder if the plugin shouldn't instead be called `eos-external` or something to that extent

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+		if (!metadata)
+			return;
+
+		g_debug ("Adopt '%s' as an external app", id);
+		gs_app_set_management_plugin (app, gs_plugin_get_name (plugin));
+	}
+}
+
+gboolean
+gs_plugin_setup (GsPlugin *plugin, GCancellable *cancellable, GError **error)
```
cosimoc: Nitpick: one argument per line for consistency with other functions

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+	/* download icons from the cdn */
+	cache_basename = g_path_get_basename (asset);
+	cache_fn = gs_utils_get_cache_filename (gs_plugin_get_name (plugin),
+						cache_basename,
+						GS_UTILS_CACHE_FLAG_NONE,
+						error);
+	if (cache_fn == NULL)
+		return FALSE;
+	if (g_file_test (cache_fn, G_FILE_TEST_EXISTS)) {
+		if (!g_file_get_contents (cache_fn, &data, &data_len, error))
```
cosimoc: You could just check for `G_FILE_ERROR_NOENT` error coming from `g_file_get_contents()` instead of checking if the file exists before hand

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+		if (!gs_plugin_download_file (plugin,
+					      app,
+					      asset,
+					      cache_fn,
+					      NULL,
+					      error))
+			return NULL;
+	}
+
+	return g_strdup (cache_fn);
```
cosimoc: Nitpick: use `g_steal_pointer()`

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+			  const char *asset,
+			  GError **error)
+{
+	gsize data_len;
+	g_autofree gchar *cache_basename = NULL;
+	g_autofree gchar *cache_fn = NULL;
+	g_autofree gchar *cache_png = NULL;
+	g_autofree gchar *data = NULL;
+	g_autoptr(AsIcon) icon = NULL;
+	g_autoptr(GdkPixbuf) pb = NULL;
```
cosimoc: `pb` and `icon` look unused

src/plugins/gs-plugin-external-apps.c
cosimoc: Also, why do you need to load the contents of the file? They don't seem to get used

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+	priv->flatpak = gs_flatpak_new (plugin, GS_FLATPAK_SCOPE_USER);
+
+	/* Run plugin after the flatpak ones because we need them to install
+	 * the app's headless part first */
+	gs_plugin_add_rule (plugin, GS_PLUGIN_RULE_RUN_BEFORE, "flatpak-system");
+	gs_plugin_add_rule (plugin, GS_PLUGIN_RULE_RUN_BEFORE, "flatpak-user");
+}
+
+static gboolean
+gs_app_is_flatpak (GsApp *app)
```
cosimoc: Should something like this not belong to the `GsApp` class?

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+
+gboolean
+gs_plugin_setup (GsPlugin *plugin, GCancellable *cancellable, GError **error)
+{
+	GsPluginData *priv = gs_plugin_get_data (plugin);
+	return gs_flatpak_setup (priv->flatpak, cancellable, error);
+}
+
+static char *
+gs_plugin_download_asset (GsPlugin *plugin,
```
cosimoc: Should probably use a less generic name, since this does not apply to all plugins

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+					      NULL,
+					      error))
+			return NULL;
+	}
+
+	return g_strdup (cache_fn);
+}
+
+static gboolean
+gs_plugin_run_command (const char *working_dir,
```
cosimoc: Should also probably use a less generic name

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+		       const char **argv,
+		       GError **error)
+{
+	g_autofree char *stderr_buf = NULL;
+	g_autofree char *stdout_buf = NULL;
+	g_autofree char *cmd = NULL;
+	int exit_val = -1;
+
+	if (!g_spawn_sync (working_dir, (char **) argv, NULL,
```
cosimoc: Consider using `GSubprocess` here

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+			   &stderr_buf, &exit_val, error)) {
+		return FALSE;
+	}
+
+	cmd = g_strjoinv (" ", (char **) argv);
+	g_debug ("Result of running '%s': retcode=%d stdout='%s' stderr='%s'",
+		 cmd, exit_val,	(stdout_buf ? stdout_buf : "NULL"),
+		 (stderr_buf ? stderr_buf : "NULL"));
+
+	return (exit_val == 0) ? TRUE : FALSE;
```
cosimoc: This function should always set the error when returning `FALSE`

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+				 gs_app_get_id (ext_runtime));
+			return FALSE;
+		}
+		break;
+
+	case AS_APP_STATE_AVAILABLE:
+		g_debug ("Installing '%s'", gs_app_get_id (ext_runtime));
+		gs_flatpak_app_install (priv->flatpak, ext_runtime, cancellable,
+					error);
+		if (gs_app_get_state (ext_runtime) != AS_APP_STATE_INSTALLED) {
```
cosimoc: Why not just check the error or the return value of `gs_flatpak_app_install()`?

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+		      GError **error)
+{
+	GsApp *ext_runtime;
+	const char *metadata = NULL;
+	GsPluginData *priv;
+
+	/* We cache all runtimes because an external runtime may have been
+	 * adopted by the flatpak plugins */
+	if (gs_app_is_flatpak (app) &&
+	    g_strcmp0 (gs_app_get_flatpak_kind_as_str (app), "runtime") == 0) {
```
cosimoc: Any reason not to use the non-string kind?

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+					     METADATA_EXTERNAL_ASSETS);
+
+	if (!metadata)
+		return NULL;
+
+	json_data = g_uri_unescape_string(metadata, NULL);
+	id = extract_runtime_info_from_json_data (json_data, &url, &error);
+
+	if (!id) {
+		if (error) {
```
cosimoc: Should not be possible for the function to return `NULL` but not set the error

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+						      gs_plugin_get_name (plugin));
+		}
+
+		return runtime;
+	}
+
+	runtime = gs_app_new (full_id);
+	gs_app_set_metadata (runtime, METADATA_HEADLESS_APP,
+			     gs_app_get_id (headless_app));
+	gs_app_set_metadata (runtime, METADATA_URL, g_strdup (url));
```
cosimoc: No need to call `g_strdup()` AFAICS

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+	g_autoptr(JsonParser) parser = NULL;
+	JsonObject *root, *runtime;
+	g_autoptr(GList) members = NULL;
+	const char *runtime_name = NULL;
+	const char *json_url = NULL;
+	guint spec = 0;
+
+	parser = json_parser_new ();
+
+	ret = json_parser_load_from_data (parser, data, strlen (data), error);
```
cosimoc: Can pass `-1` instead of calling `strlen(data)`

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+	g_autofree char *tmp_build_dir = NULL;
+	g_autofree char *repo_name = NULL;
+	g_autofree char *dir_basename = NULL;
+	const char *runtime_name = gs_app_get_id_no_prefix (runtime);
+	const char *runtime_url = gs_app_get_metadata_item (runtime,
+							    METADATA_URL);
+
+	tmp_dir = g_dir_make_tmp (template, error);
+
+	gs_app_set_progress (app, 5);
```
cosimoc: You could move this after the error check

src/plugins/gs-plugin-external-apps.c
joaquimrocha: It used to be after :) This is why I don't like comments in the code.

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+	const char *build_dir = "3rd_party_repo";
+	g_autofree char *repo_dir = NULL;
+	g_autofree char *tmp_build_dir = NULL;
+	g_autofree char *repo_name = NULL;
+	g_autofree char *dir_basename = NULL;
+	const char *runtime_name = gs_app_get_id_no_prefix (runtime);
+	const char *runtime_url = gs_app_get_metadata_item (runtime,
+							    METADATA_URL);
+
+	tmp_dir = g_dir_make_tmp (template, error);
```
cosimoc: Unfortunately on our system, `/tmp` is in RAM, so this will take a lot of precious memory.
You should use a location in `/var` such as `/var/tmp` or perhaps even better somewhere in the user's home

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+		return FALSE;
+	}
+
+	g_debug ("Building runtime extension '%s' in dir '%s'", runtime_name,
+		 tmp_dir);
+
+	tmp_build_dir = g_strconcat (tmp_dir, "/", build_dir, NULL);
+	if (g_mkdir_with_parents (tmp_build_dir, 0777) != 0) {
+		g_debug ("Failed to make dir %s", tmp_build_dir);
+		return FALSE;
```
cosimoc: The error should be set here. Consider using `g_file_make_directory_with_parents()` if you don't want to deal manually with the `errno`

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+	}
+
+	gs_app_set_progress (app, 10);
+
+	flatpak_repo_build_init (tmp_build_dir, runtime_name, error);
```
cosimoc: Error check?

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+
+	gs_app_set_progress (app, 15);
+
+	gs_plugin_add_runtime_asset (plugin, app, tmp_dir, tmp_build_dir,
+				     runtime_name, runtime_url, error);
```
cosimoc: Error check?

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+
+	repo_dir = g_strconcat (tmp_dir, "/", runtime_name, NULL);
```
cosimoc: Nitpick: move this before calling build_export

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+	}
+
+	g_debug ("Repo '%s' exported!", repo_dir);
+
+	/* Use the temp dir random name as a prefix for the runtime. This avoids
+	 * any clashes with existing repos and makes cleaning up easier. */
+	dir_basename = g_path_get_basename (tmp_dir);
+	repo_name = g_strconcat (dir_basename, runtime_name, NULL);
+
+	gs_app_set_progress (app, 50);
```
cosimoc: This will basically happen immediately after the previous call that sets the progress to `35`, since there's no real operation that happens in between...

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,712 @@
...
+			     const gchar *build_dir,
+			     const gchar *repo_dir,
+			     const char *runtime,
+			     const char *asset,
+			     GError **error)
+{
+	g_autofree gchar *content_type = NULL;
+	g_autoptr(GFile) download_file = NULL;
+
+	char *download_name = gs_plugin_download_asset (plugin, app, runtime,
```
cosimoc: Should be `g_autofree` or it will leak

src/plugins/gs-plugin-external-apps.c
joaquimrocha: I really need the app to be in installed state and because of the changes I have seen in the flatpak code, I think this is safer :)

src/plugins/gs-plugin-external-apps.c
```diff
@@ -710,3 +710,43 @@ gs_plugin_launch (GsPlugin *plugin,
...
+		return TRUE;
+
+	gs_app_set_state (app, AS_APP_STATE_REMOVING);
+
+	priv = gs_plugin_get_data (plugin);
+
+	ext_runtime = gs_plugin_get_app_external_runtime (plugin, app);
+
+	if (!ext_runtime) {
+		g_debug ("External app '%s' have any external runtime to be"
```
cosimoc: `have any` -> `does not have any` or `has no`

src/plugins/gs-plugin-external-apps.c
```diff
@@ -710,3 +710,43 @@ gs_plugin_launch (GsPlugin *plugin,
...
+			 "removed", gs_app_get_id (app));
+	} else if (gs_app_get_state (ext_runtime) == AS_APP_STATE_INSTALLED ||
+		   gs_app_get_state (ext_runtime) == AS_APP_STATE_UPDATABLE) {
+		gboolean ret;
+		g_autoptr(GError) local_error = NULL;
+
+		ret = gs_flatpak_app_remove (priv->flatpak, ext_runtime,
+					     cancellable, &local_error);
+		if (!ret && local_error)
```
cosimoc: The function above should not return `FALSE` and not set the error (hopefully!)

src/gs-utils.h
```diff
@@ -63,7 +63,8 @@ GPermission	*gs_utils_get_permission	(const gchar	*id);
 gboolean	 gs_utils_strv_fnmatch		(gchar		**strv,
 						 const gchar	*str);
 GDesktopAppInfo *gs_utils_get_desktop_app_info	(const gchar	*id);
-
+gboolean	 gs_utils_rmtree		(const gchar	*directory,
```
cosimoc: Weird indentation of the opening paren

src/gs-utils.c
```diff
@@ -357,4 +357,60 @@ gs_utils_unlink (const gchar *filename, GError **error)
 	return TRUE;
 }
 
+static gboolean
+gs_utils_rmtree_real (const gchar *directory, GError **error)
```
cosimoc: Why is the `_real` indirection needed here?

src/plugins/gs-flatpak.c
```diff
@@ -1617,3 +1617,16 @@ gs_flatpak_new (GsPlugin *plugin, GsFlatpakScope scope)
...
+gs_flatpak_get_prefix (GsFlatpak *self)
+{
+	if (self->scope == GS_FLATPAK_SCOPE_USER)
+		return GS_FLATPAK_USER_PREFIX;
+	else if (self->scope == GS_FLATPAK_SCOPE_SYSTEM)
+		return GS_FLATPAK_SYSTEM_PREFIX;
+
+	g_assert_not_reached ();
+
+	return "";
```
cosimoc: Nitpick: `return NULL` (but not that it matters)

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,804 @@
...
+#define EXTERNAL_ASSETS_SPEC_VERSION 1
+#define JSON_SPEC_KEY "spec"
+#define JSON_RUNTIME_KEY "runtime"
+#define JSON_RUNTIME_NAME_KEY "name"
+#define JSON_RUNTIME_URL_KEY "url"
+
+#define METADATA_URL "GnomeSoftware::external-app::url"
+#define METADATA_HEADLESS_APP "GnomeSoftware::external-app::headless-app"
+#define METADATA_BUILD_DIR "GnomeSoftware::external-app::build-dir"
+#define METADATA_EXTERNAL_ASSETS "flatpack-3rdparty::external-assets"
```
cosimoc: `flatpack` -> `flatpak`?

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,804 @@
...
+		if (!gs_flatpak_refine_app (priv->flatpak, ext_runtime,
+					    GS_PLUGIN_REFINE_FLAGS_DEFAULT,
+					    cancellable, error)) {
+			g_debug ("Failed to refine '%s'",
+				 gs_app_get_id (ext_runtime));
+			return FALSE;
+		}
+	}
+
+	if (gs_app_get_state (ext_runtime) == AS_APP_STATE_UNKNOWN) {
```
cosimoc: This code path needs to set error when returning `FALSE` -- but I wonder if it's impossible to arrive here with this state here when the functions above succeed?

src/plugins/gs-plugin-external-apps.c
cosimoc: I can believe that, but then you need to make sure that `error` is set in any case when returning `FALSE`. I thoight checking the error from the function above could have been a way to kill two birds with one stone...

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,804 @@
...
+			g_debug ("Failed to install '%s'",
+				 gs_app_get_id (ext_runtime));
+			return FALSE;
+		}
+
+		/* Clean-up remote (we only needed it for installing the app) */
+		remote_name = gs_app_get_origin (ext_runtime);
+		if (!flatpak_remote_delete (remote_name,
+					    &internal_error)) {
+			if (internal_error) {
```
cosimoc: Can `flatpak_remote_delete()` return `FALSE` without setting the error?

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,804 @@
...
+		 * be dealt with in the check for the 'installed' state below */
+		break;
+	}
+
+	if (gs_app_get_state (ext_runtime) != AS_APP_STATE_INSTALLED) {
+		gs_app_set_state (app, gs_app_get_state (ext_runtime));
+		g_debug ("Cannot install '%s' because runtime '%s' is not "
+			 "installed", gs_app_get_id (app),
+			 gs_app_get_id (ext_runtime));
+		return FALSE;
```
cosimoc: Set error

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,804 @@
...
+	}
+
+	gs_flatpak_app_install (priv->flatpak, app, cancellable, error);
+
+	if (gs_app_get_state (app) != AS_APP_STATE_INSTALLED) {
+		g_debug ("Failed to install '%s'", gs_app_get_id (app));
+		return FALSE;
```
cosimoc: Set error or check the return value of `gs_flatpak_app_install()`

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,804 @@
...
+		      GError **error)
+{
+	GsApp *ext_runtime;
+	const char *metadata = NULL;
+	GsPluginData *priv;
+
+	/* We cache all runtimes because an external runtime may have been
+	 * adopted by the flatpak plugins */
+	if (app_is_flatpak (app) &&
+	    g_strcmp0 (gs_app_get_flatpak_kind_as_str (app), "runtime") == 0) {
```
cosimoc: Make `gs_app_get_flatpak_kind()` public and use that? (optional)

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,804 @@
...
+		gs_plugin_cache_add (plugin, gs_app_get_id (app), app);
+		g_debug ("Caching remote '%s'", gs_app_get_id (app));
+	}
+
+	/* only process this app if was created by this plugin */
+	if (g_strcmp0 (gs_app_get_management_plugin (app),
```
cosimoc: Should this move at the beginning of the function?

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,804 @@
...
+	if (!gs_flatpak_refine_app (priv->flatpak, app, flags, cancellable, 
+				    error)) {
+		g_debug ("Refining app %s failed!", gs_app_get_id (app));
+		return FALSE;
+	}
+
+	/* We set the state to available because we assume that we can build
+	 * the runtime */
+	if (gs_app_get_state (ext_runtime) != AS_APP_STATE_INSTALLED) {
+		gs_app_set_state (app, AS_APP_STATE_UNKNOWN);
```
cosimoc: Any reason to set the state twice here?

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,804 @@
...
+	full_id = g_strdup_printf ("%s:%s",
+				   gs_flatpak_get_prefix (priv->flatpak),
+				   id);
+
+	runtime = gs_plugin_cache_lookup (plugin, full_id);
+	if (runtime) {
+		g_debug ("Found cached '%s'", full_id);
+		/* Adopt runtime if needed */
+		if (g_strcmp0 (gs_app_get_management_plugin (runtime),
+			       gs_plugin_get_name (plugin)) != 0) {
```
cosimoc: Anything wrong happening when you always set this?

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,804 @@
...
+{
+	GsPluginData *priv = gs_plugin_get_data (plugin);
+	GError *local_error = NULL;
+	const char *runtime_name = gs_app_get_id_no_prefix (runtime);
+	g_autofree char *build_dir = g_build_filename (priv->runtimes_build_dir,
+						       runtime_name,
+						       NULL);
+
+	if (!gs_utils_rmtree (build_dir, &local_error)) {
+		if (local_error->code != G_FILE_ERROR_NOENT) {
```
cosimoc: Use `g_error_matches()`

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,804 @@
...
+			g_debug ("Cannot runtime build dir '%s': %s", build_dir,
```
cosimoc: `Cannot runtime` -> `Cannot remove`?

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,804 @@
...
+		   const char *runtime,
+		   const char *asset,
+		   GError **error)
+{
+	g_autofree char *content_type = NULL;
+	g_autoptr(GFile) download_file = NULL;
+	g_autofree char *download_name = download_asset (plugin, app, runtime,
+							 asset, error);
+
+	if (!download_name) {
```
cosimoc: No curly braces (multiple occurrences)

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,804 @@
...
+	if (!content_type) {
+		return FALSE;
+	}
+
+	g_debug ("Adding runtime asset with type '%s'", content_type);
+
+	if (g_strcmp0 (content_type, "application/x-deb") == 0) {
+		return add_runtime_deb_asset (build_dir, repo_dir,
+					      download_name, error);
+	} else {
```
cosimoc: No need for `else` here since you return before

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,804 @@
...
+			   &stderr_buf, &exit_val, error)) {
+		return FALSE;
+	}
+
+	cmd = g_strjoinv (" ", (char **) argv);
+	g_debug ("Result of running '%s': retcode=%d stdout='%s' stderr='%s'",
+		 cmd, exit_val,	(stdout_buf ? stdout_buf : "NULL"),
+		 (stderr_buf ? stderr_buf : "NULL"));
+
+	if (exit_val != 0) {
```
cosimoc: Use `g_spawn_check_exit_status()` instead

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,804 @@
...
+	g_autofree gchar *cache_png = NULL;
+	g_autofree gchar *data = NULL;
+
+	cache_basename = g_path_get_basename (asset);
+	cache_fn = gs_utils_get_cache_filename (gs_plugin_get_name (plugin),
+						cache_basename,
+						GS_UTILS_CACHE_FLAG_NONE,
+						error);
+	if (cache_fn == NULL)
+		return FALSE;
```
cosimoc: `return NULL`

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,804 @@
...
+static void
+remove_runtimes_build_dir (GsPlugin *plugin)
+{
+	g_autoptr(GError) error = NULL;
+	GsPluginData *priv = gs_plugin_get_data (plugin);
+
+	if (gs_utils_rmtree (priv->runtimes_build_dir, &error))
+		return;
+
+	if (error->code != G_FILE_ERROR_NOENT) {
```
cosimoc: Use `g_error_matches()`

src/plugins/gs-plugin-eos.c
```diff
@@ -203,10 +205,10 @@ gs_plugin_eos_update_app_shortcuts_info (GsPlugin *plugin,
 static gboolean
 gs_plugin_eos_app_is_flatpak (GsApp *app)
 {
-	const char *app_manager;
-	app_manager = gs_app_get_management_plugin (app);
+	const char *id = gs_app_get_id (app);
 
-	return g_strcmp0 (app_manager, "flatpak") == 0;
+	return g_str_has_prefix (id, GS_FLATPAK_SYSTEM_PREFIX) ||
```
cosimoc: Another version of this method (in the previous commit) checks also for `id != NULL`. Should make those consistent.

src/plugins/gs-plugin-external-apps.c
cosimoc: This still needs fixing

src/plugins/gs-plugin-external-apps.c
cosimoc: This still needs addressing

src/gs-utils.h
joaquimrocha: Why? It's aligned with the brackets above AFAICS. (and Richard uses tabs, much to my dislike but I try to be consistent with the existing code style)

src/gs-utils.c
joaquimrocha: This is just so it prints only one message about removing the directory (instead of several if we put the debug in the recursively called function).

src/plugins/gs-flatpak.c
joaquimrocha: I agree. Thanks.

src/plugins/gs-plugin-external-apps.c
joaquimrocha: Good catch!

src/plugins/gs-plugin-external-apps.c
joaquimrocha: I had this as a defensive block before (because if there's some unaccounted issue that let *build_runtime* pass, then *gs_flatpak_refine_app* can actually succeed while the state is still left as *UNKNOWN*).
However, I realize now that this is not necessary because we also deal with the case after the switch under this.
Will remove. Thanks!

src/plugins/gs-plugin-external-apps.c
joaquimrocha: In a lot of these places I don't set errors because the upstream code also doesn't do it but I agree it may seem inconsistent so I can set them.

src/plugins/gs-plugin-external-apps.c
joaquimrocha: The gs-flatpak header doesn't include the flatpak headers for returning the FLATPAK_REF_KIND and maybe that's intentional. So I prefer to add a *gs_flatpak_app_is_runtime* function returning a boolean instead (even if I use the flatpak.h header myself in gs-plugin-external-apps.c).

src/plugins/gs-plugin-external-apps.c
joaquimrocha: We are only adopting apps (setting the management plugin) that have the *external-assets* metadata because we don't want to steal apps from the normal flatpak plugins but I haven't set any appstream data in the runtimes extensions so we need the blocks in this order to not miss any runtimes that might be the external apps' needed ones.

Actually since the normal flatpak plugins run after this one, they should adopt apps and runtimes anyway but I feel like it's safer to leave to do it this way in case the other plugins change how they adopt the apps (e.g. if they start checking if the apps have already a management plugin, then without the blocks above as they are, we'd need to adopt every runtime and that'd steal them from the other plugins). If this is too confusing we can chat on Slack.

src/plugins/gs-plugin-external-apps.c
joaquimrocha: That's because state transitions are "regulated". An app cannot go from installed to available directly but can go from installed to unknown to available.
Just in case it's not clear, the reason we fake that the app is uninstalled is because it needs the ext. runtime which is not installed.

src/plugins/gs-plugin-external-apps.c
joaquimrocha: This is because *gs_app_set_management_plugin* always frees+reassigns the string. I will instead move the logic to GsApp and ask Richard if not having it was intentional somehow.

src/plugins/gs-plugin-external-apps.c
joaquimrocha: Cool. Didn't know that one.

src/gs-utils.h
cosimoc: It had a lot of stray spaces before, looks OK now :)

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,818 @@
...
+	g_debug ("Adding runtime asset with type '%s'", content_type);
+
+	if (g_strcmp0 (content_type, "application/x-deb") == 0) {
+		return add_runtime_deb_asset (build_dir, repo_dir,
+					      download_name, error);
+	}
+
+	g_debug ("Cannot deal with asset type '%s'", content_type);
+
+	return FALSE;
```
cosimoc: Still needs to set the error here

src/plugins/gs-plugin-external-apps.c
```diff
@@ -0,0 +1,818 @@
...
+	download_file = g_file_new_for_path (download_name);
+	content_type = gs_utils_get_content_type (download_file, NULL,
+						  error);
+
+	if (!content_type)
+		return FALSE;
+
+	g_debug ("Adding runtime asset with type '%s'", content_type);
+
+	if (g_strcmp0 (content_type, "application/x-deb") == 0) {
```
cosimoc: Consider using g_content_type_equals or g_content_type_is_a here instead

PR #20, opened 2016-06-20
T12038

joaquimrocha:
See https://phabricator.endlessm.com/T12038

### DIFF
--- a/src/gs-shell-overview.c
+++ b/src/gs-shell-overview.c
@@ -284,6 +284,58 @@ out:
 	}
 }
 
+/**
+ * get_categories_cb:
+ * This function is used instead of the gs_shell_get_categories_cb to set the
+ * categories in the side filter. The original function is not removed so it is
+ * easier to maintain the source code.
+ **/
+static void
+get_categories_cb (GObject *source_object,
+		   GAsyncResult *res,
+		   gpointer user_data)
+{
+	GsShellOverview *self = GS_SHELL_OVERVIEW (user_data);
+	GsShellOverviewPrivate *priv = gs_shell_overview_get_instance_private (self);
+	GsPluginLoader *plugin_loader = GS_PLUGIN_LOADER (source_object);
+	guint i;
+	GsCategory *cat;
+	gboolean has_category = FALSE;
+	g_autoptr(GError) error = NULL;
+	g_autoptr(GPtrArray) list = NULL;
+
+	list = gs_plugin_loader_get_categories_finish (plugin_loader, res, &error);
+	if (list == NULL) {
+		if (!g_error_matches (error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
+			g_warning ("failed to get categories: %s", error->message);
+		goto out;
+	}
+
+	gs_shell_side_filter_clear_categories (priv->shell);
+
+	for (i = 0; i < list->len; i++) {
+		cat = GS_CATEGORY (g_ptr_array_index (list, i));
+		if (gs_category_get_size (cat) == 0)
+			continue;
+		has_category = TRUE;
+		gs_shell_side_filter_add_category (priv->shell, cat);
+	}
+
+out:
+	priv->empty = !has_category;
+
+	/* We always show the side filter in the overview page because it
+	 * has the "Featured" row */
+	gs_shell_side_filter_set_visible (priv->shell, TRUE);
+
+	priv->loading_categories = FALSE;
+	priv->refresh_count--;
+	if (priv->refresh_count == 0) {
+		priv->cache_valid = TRUE;
+		g_signal_emit (self, signals[SIGNAL_REFRESHED], 0);
+	}
+}
+
 static void
 category_tile_clicked (GsCategoryTile *tile, gpointer data)
 {
@@ -479,7 +531,7 @@ gs_shell_overview_load (GsShellOverview *self)
 		gs_plugin_loader_get_categories_async (priv->plugin_loader,
 						       GS_PLUGIN_REFINE_FLAGS_DEFAULT,
 						       priv->cancellable,
-						       gs_shell_overview_get_categories_cb,
+						       get_categories_cb,
 						       self);
 		priv->refresh_count++;
 	}
@@ -528,6 +580,11 @@ gs_shell_overview_switch_to (GsPage *page, gboolean scroll_up)
 	}
 
 	gs_grab_focus_when_mapped (priv->scrolledwindow_overview);
+
+	/* hide the category related UI because it is handled in the
+	 * side filter */
+	gtk_widget_set_visible (priv->categories_expander, FALSE);
+	gtk_widget_set_visible (priv->category_heading, FALSE);
 
 	if (priv->cache_valid || priv->refresh_count > 0)
 		return;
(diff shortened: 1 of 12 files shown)


src/gs-side-filter-row.c
```diff
@@ -0,0 +1,146 @@
+/* -*- Mode: C; tab-width: 8; indent-tabs-mode: t; c-basic-offset: 8 -*-
+ *
+ * Copyright (C) 2016 Endless Mobile, Inc.
+ * Copyright (C) 2013 Matthias Clasen <mclasen@redhat.com>
```
cosimoc: Nitpick: swap these two

src/gs-side-filter-row.c
```diff
@@ -0,0 +1,146 @@
...
+void
+gs_side_filter_row_set_category (GsSideFilterRow *row, GsCategory *cat)
+{
+	GPtrArray *key_colors;
+
+	g_return_if_fail (GS_IS_SIDE_FILTER_ROW (row));
+	g_return_if_fail (GS_IS_CATEGORY (cat));
+
+	g_clear_object (&row->cat);
+	row->cat = g_object_ref (cat);
```
cosimoc: Use `g_set_object()`?

src/gs-side-filter-row.c
```diff
@@ -0,0 +1,146 @@
...
+					      leftborder);
+}
+
+GtkWidget *
+gs_side_filter_row_new (GsCategory *cat)
+{
+	GsSideFilterRow *row;
+
+	row = g_object_new (GS_TYPE_SIDE_FILTER_ROW, NULL);
+	gs_side_filter_row_set_category (row, cat);
```
cosimoc: I usually try to keep the _new() function as dumb as possible (i.e. only call `g_object_new()`), because that's what would be expected when using a class from bindings (not an issue in this case).
Still you could either have the caller call the setter, or make the category a property you also set in the call to `g_object_new()`.

src/gs-side-filter-row.h
```diff
@@ -0,0 +1,49 @@
...
+
+G_BEGIN_DECLS
+
+#define GS_TYPE_SIDE_FILTER_ROW (gs_side_filter_row_get_type ())
+
+G_DECLARE_FINAL_TYPE (GsSideFilterRow, gs_side_filter_row, GS, SIDE_FILTER_ROW, GtkListBoxRow)
+
+GtkWidget	*gs_side_filter_row_new			(GsCategory		*cat);
+GsCategory      *gs_side_filter_row_get_category	(GsSideFilterRow	*row);
+void		 gs_side_filter_row_set_category	(GsSideFilterRow	*row,
```
cosimoc: Indentation is a bit off here

src/gtk-style.css
```diff
@@ -303,3 +303,36 @@ button.star, .button.star {
 .upgrade-progressbar {
 	box-shadow: none
 }
+
+.sidefilter_scrolledwindow {
```
cosimoc: I don't see this being created anywhere

cosimoc: Oh it's in the next commit. Perhaps move the CSS changes there too then

src/gs-shell.c
```diff
@@ -853,6 +877,125 @@ gs_shell_show_search_result (GsShell *shell, const gchar *id, const gchar *searc
...
+
+	gdk_rgba_parse (&color, color_str);
+	cat = gs_category_new (id);
+	gs_category_set_name (cat, name);
+	gs_category_set_icon (cat, icon);
+	gs_category_add_key_color (cat, &color);
+
+	row = GS_SIDE_FILTER_ROW (gs_shell_side_filter_add_category (shell,
+								     cat));
+	gs_side_filter_row_set_mode (row, mode);
```
cosimoc: Should unref the category at the end of this method, since the side filter will take a reference.

src/gs-shell.c
```diff
@@ -853,6 +877,125 @@ gs_shell_show_search_result (GsShell *shell, const gchar *id, const gchar *searc
...
+	GsShellPrivate *priv = gs_shell_get_instance_private (shell);
+
+	row = gs_side_filter_row_new (cat);
+	gtk_list_box_insert (GTK_LIST_BOX (priv->side_filter), row, -1);
+
+	return row;
+}
+
+static void
+gs_shell_side_filter_add_mode (GsShell *shell,
```
cosimoc: I think this should be called `add_overview()`, since it only works with that.

src/gs-shell.c
```diff
@@ -884,6 +884,24 @@ gs_shell_side_filter_set_visible (GsShell *shell, gboolean visible)
...
+	g_autoptr(GList) rows = NULL;
+	GList *l;
+	GsShellPrivate *priv = gs_shell_get_instance_private (shell);
+
+	rows = gtk_container_get_children (GTK_CONTAINER (priv->side_filter));
+	for (l = rows; l; l = l->next) {
+		GsSideFilterRow *current = GS_SIDE_FILTER_ROW (l->data);
+		GsShellMode mode = gs_side_filter_row_get_mode (current);
+		if (mode == GS_SHELL_MODE_CATEGORY) {
+			gtk_container_remove (GTK_CONTAINER (priv->side_filter),
```
cosimoc: I usually prefer to call `gtk_widget_destroy()` rather than `gtk_container_remove()`.

src/gs-shell-overview.c
```diff
@@ -284,6 +284,60 @@ gs_shell_overview_get_featured_cb (GObject *source_object,
...
+		cat = GS_CATEGORY (g_ptr_array_index (list, i));
+		if (gs_category_get_size (cat) == 0)
+			continue;
+		has_category = TRUE;
+		gs_shell_side_filter_add_category (priv->shell, cat);
+	}
+
+out:
+	if (has_category) {
+		priv->empty = FALSE;
```
cosimoc: Should this not be set to `!has_category` instead?

src/gs-side-filter-row.c
joaquimrocha: The category is not a property of the gs_side_filter_row. Maybe it should be but this was highly based on the GsCategoryTile (although we can/should of course improve things).

src/gs-shell.c
joaquimrocha: This is future proof code as I hope we end up moving the top tabs into the side filter list as well :)

src/gs-shell-overview.c
joaquimrocha: No. If it **has** categories then it is not empty.

cosimoc: Yeah, would `priv->empty = !has_category` not achieve that?

src/gs-shell.c
cosimoc: Then I would just return instead of asserting the passed-in mode is the overview one.

src/gs-side-filter-row.c
cosimoc: `g_set_object()` can be used without the category being a property, and has the advantage of also working in case `cat` and `row->cat` are exactly the same instance (the current code may release the last reference of the instance when calling `g_clear_object()`)

PR #21, opened 2016-06-26
T12036

joaquimrocha:
https://phabricator.endlessm.com/T12036

### DIFF
--- a/src/gs-image-tile.c
+++ b/src/gs-image-tile.c
@@ -1,0 +1,230 @@
+/* -*- Mode: C; tab-width: 8; indent-tabs-mode: t; c-basic-offset: 8 -*-
+ *
+ * Copyright (C) 2013 Matthias Clasen <mclasen@redhat.com>
+ * Copyright (C) 2016 Endless Mobile, Inc.
+ *
+ * Licensed under the GNU General Public License Version 2
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include "config.h"
+
+#include <glib/gi18n.h>
+#include <gtk/gtk.h>
+
+#include "gs-image-tile.h"
+#include "gs-star-widget.h"
+#include "gs-common.h"
+
+struct _GsImageTile
+{
+	GsAppTile	 parent_instance;
+
+	GsApp		*app;
+	GtkWidget	*app_name;
+	GtkWidget	*app_summary;
+	GtkWidget	*hover_app_name;
+	GtkWidget	*icon;
+	GtkWidget	*eventbox;
+	GtkWidget	*stack;
+	GtkWidget	*stars;
+	GtkWidget	*details_revealer;
+};
+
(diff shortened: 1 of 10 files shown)


src/gs-image-tile.c
```diff
@@ -0,0 +1,230 @@
+/* -*- Mode: C; tab-width: 8; indent-tabs-mode: t; c-basic-offset: 8 -*-
+ *
+ * Copyright (C) 2016 Endless Mobile, Inc.
+ * Copyright (C) 2013 Matthias Clasen <mclasen@redhat.com>
```
cosimoc: Swap these two around

src/gs-image-tile.c
```diff
@@ -0,0 +1,230 @@
...
+					      details_revealer);
+}
+
+GtkWidget *
+gs_image_tile_new (GsApp *app)
+{
+	GsImageTile *tile;
+
+	tile = g_object_new (GS_TYPE_IMAGE_TILE, NULL);
+	gs_app_tile_set_app (GS_APP_TILE (tile), app);
```
cosimoc: See the comment I made in the other PR about keeping _new() functions as dumb as possible.

src/gs-image-tile.c
```diff
@@ -0,0 +1,230 @@
...
+	gs_star_widget_set_icon_size (GS_STAR_WIDGET (tile->stars), 12);
+}
+
+static gboolean
+gs_image_tile_crossing_notify_event (GtkWidget *widget,
+				     GdkEventCrossing *event)
+{
+	GtkWidgetClass *klass = GTK_WIDGET_CLASS (gs_image_tile_parent_class);
+	GsImageTile *tile = GS_IMAGE_TILE (widget);
+	gboolean type_enter = event->type == GDK_ENTER_NOTIFY;
```
cosimoc: Could use parenthesis around the `==` condition

src/gs-image-tile.c
```diff
@@ -0,0 +1,230 @@
...
+	gtk_label_set_label (GTK_LABEL (tile->app_summary),
+			     gs_app_get_summary (app));
+}
+
+static void
+gs_image_tile_destroy (GtkWidget *widget)
+{
+	GsImageTile *tile = GS_IMAGE_TILE (widget);
+
+	if (tile->app)
```
cosimoc: You could call `gs_image_tile_set_app(NULL)`

src/gs-image-tile.c
```diff
@@ -0,0 +1,230 @@
...
+		gtk_widget_set_visible (tile->stars, TRUE);
+		gs_star_widget_set_rating (GS_STAR_WIDGET (tile->stars),
+					   gs_app_get_rating (tile->app));
+	} else {
+		gtk_widget_set_visible (tile->stars, FALSE);
+	}
+	gtk_stack_set_visible_child_name (GTK_STACK (tile->stack), "content");
+
+	g_signal_connect (tile->app, "notify::state",
+		 	  G_CALLBACK (app_state_changed), tile);
```
cosimoc: You can make the callback signature just have a `GsAppTile` parameter and use `g_signal_connect_swapped()` above (we don't need the pspec)

src/gs-image-tile.c
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	g_object_unref (tile);
+	return G_SOURCE_REMOVE;
+}
+
+static void
+app_state_changed (GsApp *app, GParamSpec *pspec, GsImageTile *tile)
+{
+	g_idle_add (app_state_changed_idle, g_object_ref (tile));
```
cosimoc: Why do we need to do this in an idle instead of right here?

src/gs-image-tile.c
```diff
@@ -0,0 +1,230 @@
...
+	switch (gs_app_get_state (tile->app)) {
+	case AS_APP_STATE_INSTALLED:
+	case AS_APP_STATE_INSTALLING:
+	case AS_APP_STATE_REMOVING:
+	case AS_APP_STATE_UPDATABLE:
+	case AS_APP_STATE_UPDATABLE_LIVE:
+		installed = TRUE;
+		name = g_strdup_printf ("%s (%s)",
+					gs_app_get_name (tile->app),
+					_("Installed"));
```
cosimoc: This does not seem particularly friendly to translators. What if the language does not use parenthesis like that? Perhaps the whole string `%s (Installed)` should be translatable, with a comment on what `%s` is.

PR #23, opened 2016-06-26
Show the app's icon its image tile when there's no main image

joaquimrocha:
This is accomplished by including the icon in the image tile's normal
view and changing how the main image is assigned to it: instead of
applying it as the background of the entire widget, it is applied as
the background of a widget that is layed over the mentioned icon. This
way, when there is a main image the icon is hidden behind it, but when
there is none, the icon is visible.

This allows the GsImageTile to display a graphical asset related to its
app even with apps that do not have that main image, which keeps a good
UX.

See https://phabricator.endlessm.com/T12144

### DIFF
--- a/src/gs-image-tile.ui
+++ b/src/gs-image-tile.ui
@@ -33,6 +33,83 @@
             <property name="visible">True</property>
             <property name="halign">fill</property>
             <property name="valign">fill</property>
+	    <child type="overlay">
+	      <object class="GtkEventBox" id="image_box">
+                <property name="visible">True</property>
+		<property name="halign">fill</property>
+		<property name="valign">fill</property>
+		<style>
+		  <class name="main-image"/>
+		</style>
+	      </object>
+	    </child>
+	    <child type="overlay">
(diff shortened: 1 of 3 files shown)


src/gs-image-tile.ui
```diff
@@ -140,69 +217,20 @@
...
-                        <property name="expand">False</property>
-                        <property name="fill">True</property>
-		      </packing>
-		    </child>
-                  </object>
-                </child>
+	      <!-- The fallback icon should be covered by the main image but
+	           is here for the cases where that image doesn't exist, as
+	           without a graphical reference it is very difficult to spot
+	           the applications -->
```
cosimoc: Does e.g. Glade preserve comments in UI files like this? Otherwise it would be better to have this comment in the code somewhere.

joaquimrocha: Richard told me that some of the UI files have been so modified manually that I shouldn't use glade so I have edited all this manually. Still I will move this to the code instead.

PR #26, opened 2016-06-30
plugin/eos: Add app to desktop after it is installed

joaquimrocha:
https://phabricator.endlessm.com/T11382

### DIFF
--- a/src/plugins/gs-plugin-eos.c
+++ b/src/plugins/gs-plugin-eos.c
@@ -362,3 +362,22 @@ gs_plugin_remove_shortcut (GsPlugin	*plugin,
 	gs_app_remove_quirk (app, AS_APP_QUIRK_HAS_SHORTCUT);
 	return remove_app_from_shell (plugin, app, cancellable, error);
 }
+
+gboolean
+gs_plugin_app_install (GsPlugin *plugin,
+		       GsApp *app,
+		       GCancellable *cancellable,
+		       GError **error)
+{
+	if (!gs_plugin_eos_app_is_flatpak (app))
+		return TRUE;
+
+	/* We're only interested in already installed flatpak apps so we can
+	 * add them to the desktop */
+	if (gs_app_get_state (app) != AS_APP_STATE_INSTALLED)
+		return TRUE;
+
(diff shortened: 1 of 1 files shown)


src/plugins/gs-plugin-eos.c
```diff
@@ -362,3 +362,30 @@ gs_plugin_remove_shortcut (GsPlugin	*plugin,
 	gs_app_remove_quirk (app, AS_APP_QUIRK_HAS_SHORTCUT);
 	return remove_app_from_shell (plugin, app, cancellable, error);
 }
+
+static gboolean
+app_is_flatpak (GsApp *app)
```
cosimoc: You already have `gs_plugin_eos_app_is_flatpak()` here no?

src/plugins/gs-plugin-eos.c
```diff
@@ -362,3 +362,30 @@ gs_plugin_remove_shortcut (GsPlugin	*plugin,
...
+gs_plugin_app_install (GsPlugin *plugin,
+		       GsApp *app,
+		       GCancellable *cancellable,
+		       GError **error)
+{
+	if (!app_is_flatpak (app))
+		return TRUE;
+
+	/* We're only interested in already installed flatpak apps so we can
+	 * add them to the desktop */
```
cosimoc: Can this be called also with apps that are not installed? It's a bit unclear to me when this function is called.

src/plugins/gs-plugin-eos.c
joaquimrocha: Good catch.. I am surprised I missed that. Too many plugins I guess :)

src/plugins/gs-plugin-eos.c
joaquimrocha: Yes. *gs_plugin_app_install* is called for every app that the user presses "Install". It's up for the plugins to do what they need in this method (usually plugins check if the app is "theirs" using the management plugin property before proceeding). I took advantage of this method in order to complete the installation on EOS (i.e. adding the icon to the desktop). This works because the plugin runs **after** the flatpak ones and I check if the app is **a flatpak** + **installed**.

PR #28, opened 2016-07-04
Add client side of thumbnails downloading

smspillaz:
Currently we pass the URL of the thumbnail directly to GtkCSS - this will create a lot of unnecessary load and network traffic every time gnome-software is booted up. It also doesn't work offline.

We want to cache these thumbnails. To do that, we need to download and save them to the cache directory using libsoup and then pass the filesystem location to GtkCSS.

In some cases, we might end up downloading files after the `_refine` functions on plugins have been run. A signal was added to GsApp to indicate that new metadata has been added after the fact so that interested views can refresh themselves appropriately.

https://phabricator.endlessm.com/T12142

### DIFF
--- a/src/gs-app.c
+++ b/src/gs-app.c
@@ -131,6 +131,13 @@ enum {
 	PROP_LAST
 };
 
+enum {
+	SIGNAL_METADATA_CHANGED,
+	SIGNAL_LAST
+};
+
+static guint signals [SIGNAL_LAST] = { 0 };
+
 G_DEFINE_TYPE (GsApp, gs_app, G_TYPE_OBJECT)
 
 static void
@@ -2075,6 +2082,10 @@ gs_app_set_metadata (GsApp *app, const gchar *key, const gchar *value)
 
 	/* if no value, then remove the key */
 	if (value == NULL) {
+		g_signal_emit (app,
+	        	       signals[SIGNAL_METADATA_CHANGED],
+			       g_quark_from_string (key),
+			       key);
 		g_hash_table_remove (app->metadata, key);
 		return;
 	}
@@ -2093,6 +2104,11 @@ gs_app_set_metadata (GsApp *app, const gchar *key, const gchar *value)
 	g_hash_table_insert (app->metadata,
 			     g_strdup (key),
 			     g_string_free (str, FALSE));
+
+	g_signal_emit (app,
+	               signals[SIGNAL_METADATA_CHANGED],
+	               g_quark_from_string (key),
+	               key);
 }
 
 /**
@@ -2905,6 +2921,24 @@ gs_app_class_init (GsAppClass *klass)
 				     0, G_MAXUINT64, 0,
 				     G_PARAM_READWRITE | G_PARAM_CONSTRUCT);
 	g_object_class_install_property (object_class, PROP_QUIRK, pspec);
+
+        /**
+         * GsApp:metadata-changed:
+         * Fired when a plugin either adds or removes metadata for this
+         * GsApp. Mutating existing metadata is not a valid operation
+         * and the signal will not be fired in that case.
+         *
+         * The detail of this signal will be the key that was either
+         * added or removed from this GsApp's metadata.
+         */
+	signals [SIGNAL_METADATA_CHANGED] =
+	        g_signal_new ("metadata-changed",
+	                      G_TYPE_FROM_CLASS (object_class),
+	                      G_SIGNAL_RUN_LAST | G_SIGNAL_DETAILED,
+	                      0,
+	                      NULL, NULL, g_cclosure_marshal_VOID__STRING,
+	                      G_TYPE_NONE, 1,
+	                      G_TYPE_STRING);
 }
 
 static void
--- a/src/gs-image-tile.c
+++ b/src/gs-image-tile.c
@@ -106,15 +106,29 @@ app_state_changed (GsApp *app, GParamSpec *pspec, GsImageTile *tile)
 }
 
 static void
+app_image_tile_css_added (GsApp *app, const char *metadata, GsAppTile *tile)
+{
+	if (g_strcmp0 (metadata, "GnomeSoftware::ImageTile-css") == 0) {
+		gs_utils_widget_set_css_app (app, GTK_WIDGET (tile),
+		                             "GnomeSoftware::ImageTile-css");
+	} else {
+		g_assert_not_reached ();
+	}
+}
+
+static void
 gs_image_tile_set_app (GsAppTile *app_tile, GsApp *app)
 {
 	GsImageTile *tile = GS_IMAGE_TILE (app_tile);
 
 	g_return_if_fail (GS_IS_APP (app) || app == NULL);
 
-	if (tile->app)
+	if (tile->app) {
 		g_signal_handlers_disconnect_by_func (tile->app,
 						      app_state_changed, tile);
+		g_signal_handlers_disconnect_by_func (tile->app,
+		                                      app_image_tile_css_added, tile);
+	}
 
 	g_set_object (&tile->app, app);
 	if (!app)
@@ -131,6 +145,8 @@ gs_image_tile_set_app (GsAppTile *app_tile, GsApp *app)
 
 	g_signal_connect (tile->app, "notify::state",
 		 	  G_CALLBACK (app_state_changed), tile);
+	g_signal_connect (tile->app, "metadata-changed::GnomeSoftware::ImageTile-css",
+	                  G_CALLBACK (app_image_tile_css_added), tile);
 	app_state_changed (tile->app, NULL, tile);
 
 	/* perhaps set custom css */
--- a/src/plugins/gs-plugin-eos.c
+++ b/src/plugins/gs-plugin-eos.c
@@ -22,9 +22,12 @@
 #include <config.h>
 
 #include <gio/gdesktopappinfo.h>
+#include <libsoup/soup.h>
 #include <gnome-software.h>
 #include <glib/gi18n.h>
+#include <gs-common.h>
 #include <gs-plugin.h>
+#include <gs-utils.h>
 
 #include "gs-flatpak.h"
 
@@ -38,6 +41,7 @@ struct GsPluginData
 	GDBusConnection *session_bus;
 	GHashTable *desktop_apps;
 	int applications_changed_id;
+	SoupSession *soup_session;
 };
 
 static GHashTable *
@@ -106,6 +110,9 @@ gs_plugin_initialize (GsPlugin *plugin)
 						    G_DBUS_SIGNAL_FLAGS_NONE,
 						    (GDBusSignalCallback) on_desktop_apps_changed,
 						    plugin, NULL);
+	priv->soup_session = soup_session_new_with_options (SOUP_SESSION_USER_AGENT,
+	                                                    gs_user_agent (),
+	                                                    NULL);
 }
 
 /**
@@ -123,6 +130,7 @@ gs_plugin_destroy (GsPlugin *plugin)
 	}
 
 	g_clear_object (&priv->session_bus);
+	g_clear_object (&priv->soup_session);
 	g_hash_table_destroy (priv->desktop_apps);
 }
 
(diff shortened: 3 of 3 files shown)


src/gs-app.c
```diff
@@ -2093,6 +2100,12 @@ gs_app_set_metadata (GsApp *app, const gchar *key, const gchar *value)
 	g_hash_table_insert (app->metadata,
 			     g_strdup (key),
 			     g_string_free (str, FALSE));
+
+	g_signal_emit (app,
+	               signals[SIGNAL_METADATA_ADDED],
```
cosimoc: I'd call this `metadata-changed` and also call it when a key is removed (since replacing it does not seem a valid operation).

src/gs-app.c
```diff
@@ -2093,6 +2100,12 @@ gs_app_set_metadata (GsApp *app, const gchar *key, const gchar *value)
...
+	               g_quark_from_string (key),
+	               key,
+	               NULL);
```
cosimoc: I don't think this needs to be `NULL`-terminated.

src/gs-app.c
```diff
@@ -2905,6 +2918,15 @@ gs_app_class_init (GsAppClass *klass)
 				     0, G_MAXUINT64, 0,
 				     G_PARAM_READWRITE | G_PARAM_CONSTRUCT);
 	g_object_class_install_property (object_class, PROP_QUIRK, pspec);
+
+	signals [SIGNAL_METADATA_ADDED] =
+	        g_signal_new ("metadata-added",
+	                      G_TYPE_FROM_CLASS (object_class),
+	                      G_SIGNAL_RUN_LAST | G_SIGNAL_DETAILED,
+	                      0,
+	                      NULL, NULL, g_cclosure_marshal_VOID__CHAR,
```
cosimoc: The closure should be `VOID__STRING`

src/gs-app.c
```diff
@@ -131,6 +131,13 @@ enum {
 	PROP_LAST
 };
 
+enum {
+	SIGNAL_METADATA_ADDED = 1,
```
cosimoc: There's no need to start these from `1` IIRC

src/gs-image-tile.c
```diff
@@ -105,16 +105,67 @@ app_state_changed (GsApp *app, GParamSpec *pspec, GsImageTile *tile)
...
+	data->tile = g_object_ref (tile);
+	return data;
+}
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC(AppTileCSSChangedUserData, app_tile_css_changed_user_data_destroy);
+
+static gboolean
+app_tile_css_changed_idle (gpointer user_data)
+{
+	g_autoptr(AppTileCSSChangedUserData) data = user_data;
```
cosimoc: Just call the free function after setting the CSS -- I think it's clearer for a simple struct like this.

src/gs-image-tile.c
```diff
@@ -105,16 +105,67 @@ app_state_changed (GsApp *app, GParamSpec *pspec, GsImageTile *tile)
...
+	                             "GnomeSoftware::ImageTile-css");
+
+	return G_SOURCE_REMOVE;
+}
+
+static void
+app_image_tile_css_added (GsApp *app, const char *metadata, GsAppTile *tile)
+{
+	if (g_strcmp0 (metadata, "GnomeSoftware::ImageTile-css") == 0) {
+		g_idle_add (app_tile_css_changed_idle,
```
cosimoc: Why do you need an idle at all?

src/gs-image-tile.c
```diff
@@ -105,16 +105,67 @@ app_state_changed (GsApp *app, GParamSpec *pspec, GsImageTile *tile)
...
+		            app_tile_css_changed_user_data_new (app, tile));
```
cosimoc: Even with the idle, why do you need the additional struct? Isn't `app` just `tile->app`?

src/plugins/gs-plugin-eos.c
```diff
@@ -22,9 +22,12 @@
 #include <config.h>
 
 #include <gio/gdesktopappinfo.h>
+#include <libsoup/soup.h>
 #include <gnome-software.h>
 #include <glib/gi18n.h>
 #include <gs-plugin.h>
+#include <gs-utils.h>
+#include <gs-common.h>
```
cosimoc: Alphabetical order

src/plugins/gs-plugin-eos.c
```diff
@@ -234,8 +303,54 @@ gs_plugin_eos_refine_popular_app (GsPlugin *plugin,
 	    gs_app_get_metadata_item (app, "GnomeSoftware::ImageTile-css"))
 		return;
 
-	css = g_strdup_printf ("background-image: url('%s');", popular_bg);
-	gs_app_set_metadata (app, "GnomeSoftware::ImageTile-css", css);
+	url_basename = g_path_get_basename (popular_bg);
+
+	/* First take a hash of this URL and see if it is in our cache */
+	tile_cache_hash = g_compute_checksum_for_string(G_CHECKSUM_SHA256,
```
cosimoc: Space before paren

src/plugins/gs-plugin-eos.c
```diff
@@ -234,8 +303,54 @@ gs_plugin_eos_refine_popular_app (GsPlugin *plugin,
...
+	                                                popular_bg,
+	                                                -1);
+	cache_identifier = g_strdup_printf("%s-%s", tile_cache_hash, url_basename);
```
cosimoc: Space before paren

src/plugins/gs-plugin-eos.c
```diff
@@ -234,8 +303,54 @@ gs_plugin_eos_refine_popular_app (GsPlugin *plugin,
...
+	                                              cache_identifier,
+	                                              GS_UTILS_CACHE_FLAG_NONE,
+	                                              NULL);
+
+	/* Check to see if the file exists in the cache at the time we called this
+	 * function. If it does, then change the css so that the tile loads. Otherwise,
+	 * we'll need to asynchronously fetch the image from the server and write it
+	 * to the cache */
+	if (g_file_test (cache_filename, G_FILE_TEST_EXISTS)) {
+		gs_plugin_eos_update_tile_image_from_filename (app, cache_filename);
```
cosimoc: Return here and remove the `else`

src/plugins/gs-plugin-eos.c
```diff
@@ -234,8 +303,54 @@ gs_plugin_eos_refine_popular_app (GsPlugin *plugin,
...
+	} else {
+		GsPluginData *priv = gs_plugin_get_data (plugin);
+		PopularBackgroundImageTileRequestData *request_data = NULL;
+		g_autoptr(SoupURI) soup_uri = soup_uri_new (popular_bg);
+		g_autoptr(SoupMessage) message = NULL;
+
+		g_debug("Downloading thumbnail %s to %s\n", popular_bg, cache_filename);
```
cosimoc: `g_debug` should not need `\n`

cosimoc: Also space before paren

src/plugins/gs-plugin-eos.c
```diff
@@ -234,8 +303,54 @@ gs_plugin_eos_refine_popular_app (GsPlugin *plugin,
...
+
+		/* XXX: Note that we might have multiple downloads in progress here. We
+		 * don't make any attempt to keep track of this. */
+		message = soup_message_new_from_uri (SOUP_METHOD_GET, soup_uri);
+		if (!message) {
+			g_debug ("Couldn't download %s, network not available", popular_bg);
+			return;
+		}
+
+		request_data = g_new0(PopularBackgroundImageTileRequestData, 1);
```
cosimoc: Space before paren

src/plugins/gs-plugin-eos.c
```diff
@@ -234,8 +303,54 @@ gs_plugin_eos_refine_popular_app (GsPlugin *plugin,
...
+		request_data->app = app;
+		request_data->plugin = plugin;
+		request_data->cache_filename = g_steal_pointer(&cache_filename);
```
cosimoc: Space before paren

src/plugins/gs-plugin-eos.c
```diff
@@ -234,8 +303,54 @@ gs_plugin_eos_refine_popular_app (GsPlugin *plugin,
...
+
+		soup_session_queue_message (priv->soup_session,
+		                            g_object_ref(message),
```
cosimoc: Pretty sure `soup_session_queue_message()` already takes a ref to the message

src/plugins/gs-plugin-eos.c
```diff
@@ -220,12 +228,73 @@ gs_plugin_eos_refine_core_app (GsApp *app)
...
+                                        gpointer user_data)
+{
+	g_autoptr(PopularBackgroundImageTileRequestData) data = user_data;
+	g_autoptr(GError) error = NULL;
+
+	if (msg->status_code == SOUP_STATUS_CANCELLED)
+		return;
+
+	if (msg->status_code != SOUP_STATUS_OK) {
+		g_debug("Failed to download tile image corresponding to cache entry %s",
```
cosimoc: Space before paren

src/plugins/gs-plugin-eos.c
```diff
@@ -220,12 +228,73 @@ gs_plugin_eos_refine_core_app (GsApp *app)
...
+		        data->cache_filename);
```
cosimoc: Would be nice to get the reason why it failed (e.g. the status code)

src/plugins/gs-plugin-eos.c
```diff
@@ -220,12 +228,73 @@ gs_plugin_eos_refine_core_app (GsApp *app)
...
+		return;
+	}
+
+	/* Write out the cache image to disk */
+	if (!g_file_set_contents (data->cache_filename,
+	                          msg->response_body->data,
+	                          msg->response_body->length,
+	                          &error)) {
+		g_debug("Failed to write cache image %s, %s",
```
cosimoc: Space before paren

src/plugins/gs-plugin-eos.c
```diff
@@ -220,12 +228,73 @@ gs_plugin_eos_refine_core_app (GsApp *app)
...
+		        data->cache_filename,
+		        error->message);
+		return;
+	}
+
+	gs_plugin_eos_refine_popular_app (data->plugin, data->app);
```
cosimoc: I'd also like to explicit free the closure struct here, instead of using autoptr

src/plugins/gs-plugin-eos.c
```diff
@@ -220,12 +228,73 @@ gs_plugin_eos_refine_core_app (GsApp *app)
...
+G_DEFINE_AUTOPTR_CLEANUP_FUNC(PopularBackgroundImageTileRequestData,
+                              popular_background_image_tile_request_data_destroy);
+
+static void gs_plugin_eos_refine_popular_app(GsPlugin *plugin, GsApp *app);
+
+static void
+gs_plugin_eos_update_tile_image_from_filename (GsApp      *app,
+                                               const char *filename)
+{
+	g_autofree char *css = g_strdup_printf("background-image: url('%s')",
```
cosimoc: Space before paren

src/plugins/gs-plugin-eos.c
cosimoc: Also why do you call `refine` here, instead of setting the CSS like in the other case?

src/gs-image-tile.c
smspillaz: I can test this branch without it. I modelled this off of the screenshots code, but if we don't need an idle then I'd be more than happy to drop it.

src/plugins/gs-plugin-eos.c
smspillaz: Hmm, I'm not sure about using explicit frees here. The fact that this function has multiple return paths lends itself to using `g_autoptr`. Was there a good reason not to use it in this case?

smspillaz: > Also why do you call refine here, instead of setting the CSS like in the other case?

Thanks for catching that! I had meant to add that when I split out `gs_plugin_eos_update_image_from_filename` but obviously never got around to calling it here.

src/plugins/gs-plugin-eos.c
smspillaz: Done

src/plugins/gs-plugin-eos.c
```diff
@@ -234,8 +306,51 @@ gs_plugin_eos_refine_popular_app (GsPlugin *plugin,
...
+		return;
+	}
+
+	request_data = g_new0 (PopularBackgroundImageTileRequestData, 1);
+	request_data->app = app;
+	request_data->plugin = plugin;
+	request_data->cache_filename = g_steal_pointer (&cache_filename);
+
+	soup_session_queue_message (priv->soup_session,
+	                            g_steal_pointer(&message),
```
cosimoc: Dat paren

cosimoc: Also my comment still stands... Does `soup_session_queue_message()` actually consume the passed-in reference? Otherwise it's just fine to pass the pointer as-is.

src/plugins/gs-plugin-eos.c
```diff
@@ -220,12 +228,76 @@ gs_plugin_eos_refine_core_app (GsApp *app)
...
+{
+	GsApp *app;
+	GsPlugin *plugin;
+	char *cache_filename;
+} PopularBackgroundImageTileRequestData;
+
+static void
+popular_background_image_tile_request_data_destroy (PopularBackgroundImageTileRequestData *data)
+{
+	g_free (data->cache_filename);
```
cosimoc: You also need to free `data`, no?

src/plugins/gs-plugin-eos.c
smspillaz: Heh, I'll need someone to beat my old coding style out of me :p 

`soup_session_queue_reference` is transfer full: https://developer.gnome.org/libsoup/stable/SoupSession.html#soup-session-queue-message

src/plugins/gs-plugin-eos.c
smspillaz: doh, of course.

smspillaz: Done

PR #32, opened 2016-07-08
T12307

joaquimrocha:
https://phabricator.endlessm.com/T12307

### DIFF
--- a/src/gs-shell-overview.c
+++ b/src/gs-shell-overview.c
@@ -33,7 +33,7 @@
 #include "gs-category-tile.h"
 #include "gs-common.h"
 
-#define N_TILES 12
+#define N_TILES 35
 
 typedef struct
 {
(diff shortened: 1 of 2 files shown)


src/gs-shell-overview.c
```diff
@@ -159,7 +159,6 @@ gs_shell_overview_get_popular_cb (GObject *source_object,
 		goto out;
 	}
 	/* Don't show apps from the category that's currently featured as the category of the day */
```
rshuler: Remove the comment that goes with the code being removed

rshuler: If you simply add the "FALSE" below making it really easy to bring back that line, perhaps this should be in a similar state.  I would suggest either completely removing in both places, or do something like add a constant that is checked in both places.

joaquimrocha: Well, one line generated less possible conflicts then 20 but you got a point. Thanks (and thanks for the warning about the forgotten comment...).

PR #39, opened 2016-07-13
eos: Blacklist renamed apps

joaquimrocha:
This is in order to avoid repeated entries in the installed page (the
original name + renamed one).

https://phabricator.endlessm.com/T11786


cosimoc:
Looks good, but please also:
- bump the required version of appstream-glib in configure.ac
- bump the required version of appstream-glib in debian/control

PR #47, opened 2016-07-26
T12251 -- Add suport for Dropbox

xaiki:
in this PR we first reorganize the external-apps plugin a bit so that we can easily add more fields to the json, then implement a tar extractor (with a tar json type).

that's all folks.

https://phabricator.endlessm.com/T12251

### DIFF
--- a/src/plugins/gs-plugin-external-apps.c
+++ b/src/plugins/gs-plugin-external-apps.c
@@ -35,13 +35,21 @@
 #define JSON_RUNTIME_KEY "runtime"
 #define JSON_RUNTIME_NAME_KEY "name"
 #define JSON_RUNTIME_URL_KEY "url"
+#define JSON_RUNTIME_TYPE_KEY "type"
 
 #define METADATA_URL "GnomeSoftware::external-app::url"
+#define METADATA_TYPE "GnomeSoftware::external-app::type"
 #define METADATA_HEADLESS_APP "GnomeSoftware::external-app::headless-app"
 #define METADATA_BUILD_DIR "GnomeSoftware::external-app::build-dir"
 #define METADATA_EXTERNAL_ASSETS "flatpak-3rdparty::external-assets"
 
 #define TMP_ASSSETS_PREFIX "gs-external-apps"
+
+typedef enum {
+	GS_PLUGIN_EXTERNAL_TYPE_UNKNOWN = 0,
+	GS_PLUGIN_EXTERNAL_TYPE_DEB,
+	GS_PLUGIN_EXTERNAL_TYPE_TAR
+} GsPluginExternalType;
 
 struct GsPluginData {
 	GsFlatpak	*flatpak;
@@ -283,12 +291,24 @@ add_runtime_deb_asset (const gchar *build_dir,
 	return TRUE;
 }
 
+static inline GsPluginExternalType
+get_type_from_string (const char *type)
+{
+	if (g_strcmp0 (type, "deb") == 0)
+		return GS_PLUGIN_EXTERNAL_TYPE_DEB;
+	else if (g_strcmp0 (type, "tar") == 0)
+		return  GS_PLUGIN_EXTERNAL_TYPE_TAR;
+
+	return GS_PLUGIN_EXTERNAL_TYPE_UNKNOWN;
+}
+
 static gboolean
 add_runtime_asset (GsPlugin *plugin,
 		   GsApp *app,
 		   const gchar *build_dir,
 		   const gchar *repo_dir,
 		   const char *runtime,
+		   const char *archive_type,
 		   const char *asset,
 		   GError **error)
 {
@@ -296,6 +316,7 @@ add_runtime_asset (GsPlugin *plugin,
 	g_autoptr(GFile) download_file = NULL;
 	g_autofree char *download_name = download_asset (plugin, app, runtime,
 							 asset, error);
+	GsPluginExternalType type = get_type_from_string (archive_type);
 
 	if (!download_name)
 		return FALSE;
@@ -309,7 +330,8 @@ add_runtime_asset (GsPlugin *plugin,
 
 	g_debug ("Adding runtime asset with type '%s'", content_type);
 
-	if (g_content_type_is_a (content_type, "application/x-deb")) {
+	if ((type == GS_PLUGIN_EXTERNAL_TYPE_DEB) ||
+	    g_content_type_is_a (content_type, "application/x-deb")) {
 		return add_runtime_deb_asset (build_dir, repo_dir,
 					      download_name, error);
 	}
@@ -368,6 +390,8 @@ build_runtime (GsPlugin *plugin,
 	const char *runtime_name = gs_app_get_id_no_prefix (runtime);
 	const char *runtime_url = gs_app_get_metadata_item (runtime,
 							    METADATA_URL);
+	const char *runtime_type = gs_app_get_metadata_item (runtime,
+							     METADATA_TYPE);
 
 	tmp_dir = g_build_filename (priv->runtimes_build_dir, runtime_name,
 				    NULL);
@@ -398,7 +422,7 @@ build_runtime (GsPlugin *plugin,
 	gs_app_set_progress (app, 15);
 
 	if (!add_runtime_asset (plugin, app, tmp_dir, build_dir, runtime_name,
-				runtime_url, error)) {
+				runtime_type, runtime_url, error)) {
 		g_debug ("Failed to add the asset '%s'", runtime_name);
 		return FALSE;
 	}
@@ -442,6 +466,7 @@ build_runtime (GsPlugin *plugin,
 static char *
 extract_runtime_info_from_json_data (const char *data,
 				     char **url,
+				     char **type,
 				     GError **error)
 {
 	gboolean ret;
@@ -450,6 +475,7 @@ extract_runtime_info_from_json_data (const char *data,
 	g_autoptr(GList) members = NULL;
 	const char *runtime_name = NULL;
 	const char *json_url = NULL;
+	const char *type_str = NULL;
 	guint spec = 0;
 
 	parser = json_parser_new ();
@@ -509,6 +535,10 @@ extract_runtime_info_from_json_data (const char *data,
 		return NULL;
 	}
 
+	/* optional elements */
+	type_str = json_object_get_string_member (runtime,
+						  JSON_RUNTIME_TYPE_KEY);
+	*type = g_strdup (type_str);
 	*url = g_strdup (json_url);
 
 	return g_strdup (runtime_name);
@@ -523,6 +553,7 @@ gs_plugin_get_app_external_runtime (GsPlugin *plugin,
 	g_autofree char *id = NULL;
 	g_autofree char *full_id = NULL;
 	g_autofree char *url = NULL;
+	g_autofree char *type = NULL;
 	g_autofree char *json_data = NULL;
 	g_autoptr (GError) error = NULL;
 	const char *metadata;
@@ -534,7 +565,7 @@ gs_plugin_get_app_external_runtime (GsPlugin *plugin,
 		return NULL;
 
 	json_data = g_uri_unescape_string (metadata, NULL);
-	id = extract_runtime_info_from_json_data (json_data, &url, &error);
+	id = extract_runtime_info_from_json_data (json_data, &url, &type, &error);
 
 	if (!id) {
 		g_debug ("Error getting external runtime from "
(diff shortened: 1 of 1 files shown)


xaiki:
commits squashed and properly commented.

src/plugins/gs-plugin-external-apps.c
```diff
@@ -296,6 +296,23 @@ add_runtime_deb_asset (const gchar *build_dir,
...
 static gboolean
+add_runtime_tar_asset (const gchar *build_dir,
+		       const gchar *repo_dir,
+		       const char *asset_path,
+		       GError **error)
+{
+	/* flatpak build --runtime needs files in /usr, when coming from a
+	   debian package we can asume some files in /usr, when coming from
+	   a tgz, it's harder, so force that here. */
+	g_autofree gchar *extract_path = g_strdup_printf ("%s/usr", repo_dir);
```
joaquimrocha: Use g_build_filename.

src/plugins/gs-plugin-external-apps.c
```diff
@@ -43,8 +44,15 @@
 
 #define TMP_ASSSETS_PREFIX "gs-external-apps"
 
+enum GsPluginExternalType {
+	GS_PLUGIN_EXTERNAL_TYPE_DEB = 1,
```
joaquimrocha: Why starting with 1?

src/plugins/gs-plugin-external-apps.c
```diff
@@ -43,8 +44,15 @@
...
+	GS_PLUGIN_EXTERNAL_TYPE_TAR,
+	GS_PLUGIN_EXTERNAL_TYPE_UNKNOWN,
```
joaquimrocha: Extra comma here?
Also, I wonder whether the TYPE_UNKNOWN should come as the first type, so it's value doesn't keep changing (assuming you'd keep UNKNOWN as the last type).

src/plugins/gs-plugin-external-apps.c
```diff
@@ -43,8 +44,15 @@
...
+};
+
 struct GsPluginData {
 	GsFlatpak	*flatpak;
+	enum GsPluginExternalType type;
```
joaquimrocha: I'd prefer that you typedef'ed this type in the previous enum declaration. So we'd skip using *enum* here.

src/plugins/gs-plugin-external-apps.c
```diff
@@ -514,6 +525,17 @@ extract_runtime_info_from_json_data (GsPlugin *plugin,
 		return FALSE;
 	}
 
+	/* optional elements */
+	type = json_object_get_string_member (runtime, JSON_RUNTIME_TYPE_KEY);
+	if (type) {
+		if (g_strcmp0(type, "deb") == 0)
```
joaquimrocha: Space needed between function name and parenthesis. Same below.

joaquimrocha:
@xaiki  I don't understand why the first commit is needed here (96e2153).
I understand that if we need to extract a bunch of data fields from the json we should use a structure instead of output params, but I don't like the idea of adding data that is related to one individual runtime to a plugin struct that is general.

If you want to keep things more organized, as I mentioned above, please create an asset/runtime struct to represent that object.

src/plugins/gs-plugin-external-apps.c
xaiki: so that 0 initialized is caught.

src/plugins/gs-plugin-external-apps.c
xaiki: that allows me to test for >= UNKNOWN

src/plugins/gs-plugin-external-apps.c
xaiki: done.

src/plugins/gs-plugin-external-apps.c
xaiki: done

xaiki:
why add complexity with yet another struct that we need to unpack from priv, as our priv is pretty small ?

src/plugins/gs-plugin-external-apps.c
joaquimrocha: Then that's one more reason why UNKNOWN should come first IMO.

src/plugins/gs-plugin-external-apps.c
joaquimrocha: Why would you want to test it like that? I don't think enums should be tested with < or > .
Mind the comma too.

joaquimrocha:
Because of the reasons I've stated before + getting the code less messy. I'd likely not even add this extra structure to priv but just pass it around if feasible.

xaiki:
i'll ask @cosimoc to cut on this, we clearly have oposite views of what's messy.

joaquimrocha:
Sure, let's have @cosimoc untie this.
FTR, here's why I think it's messy to have those attributes in the plugin's priv:
1) They belong to the runtime extension, not the plugin;
2) You're using a permanent struct (the plugin's priv) to store values that will be temporary and depending on each runtime used (this is error prone);
3) You're just adding them to the priv because it is a bit laborious to keep passing multiple arguments so you accomplish the same with my proposal;
4) If you need 5 more attributes you'll need to add them all to the priv, making it messier in this case alone.

cosimoc:
I have to agree with @joaquimrocha here, since I had the same comment when reviewing the code yesterday...

Basically that data is only valid in the context of the code that builds the external runtime, not in the context of the whole plugin, which is the owner of the private struct. If another caller was to call into the plugin and see/use the data that was stored in there from a previous installation, that would be wrong and could lead to trouble.

src/plugins/gs-plugin-external-apps.c
```diff
@@ -283,19 +291,54 @@ add_runtime_deb_asset (const gchar *build_dir,
 	return TRUE;
 }
 
+GsPluginExternalType
+get_type_from_string (const gchar *type)
+{
+	if (type) {
```
cosimoc: You use `g_strcmp0()` below, so there's no need to check for `NULL` here

src/plugins/gs-plugin-external-apps.c
```diff
@@ -283,19 +291,54 @@ add_runtime_deb_asset (const gchar *build_dir,
...
+		if (g_strcmp0(type, "deb") == 0)
```
cosimoc: Space before paren

src/plugins/gs-plugin-external-apps.c
```diff
@@ -283,19 +291,54 @@ add_runtime_deb_asset (const gchar *build_dir,
...
+add_runtime_tar_asset (const gchar *build_dir,
+		       const gchar *repo_dir,
+		       const char *asset_path,
+		       GError **error)
+{
+	/* flatpak build --runtime needs files in /usr, when coming from a
+	   debian package we can asume some files in /usr, when coming from
+	   a tgz, it's harder, so force that here. */
+	g_autofree gchar *extract_path =
+		g_build_filename(repo_dir, "usr", NULL);
```
cosimoc: Space before paren

src/plugins/gs-plugin-external-apps.c
```diff
@@ -283,19 +291,54 @@ add_runtime_deb_asset (const gchar *build_dir,
...
+}
+
+static gboolean
+add_runtime_tar_asset (const gchar *build_dir,
+		       const gchar *repo_dir,
+		       const char *asset_path,
+		       GError **error)
+{
+	/* flatpak build --runtime needs files in /usr, when coming from a
+	   debian package we can asume some files in /usr, when coming from
```
cosimoc: `assume`

src/plugins/gs-plugin-external-apps.c
```diff
@@ -283,19 +291,54 @@ add_runtime_deb_asset (const gchar *build_dir,
...
+		   const char *archive_type,
 		   const char *asset,
 		   GError **error)
 {
 	g_autofree char *content_type = NULL;
 	g_autoptr(GFile) download_file = NULL;
 	g_autofree char *download_name = download_asset (plugin, app, runtime,
 							 asset, error);
+	GsPluginExternalType type =
+		get_type_from_string(archive_type);
```
cosimoc: Space before paren -- also can be on the same line probably

src/plugins/gs-plugin-external-apps.c
```diff
@@ -309,11 +352,18 @@ add_runtime_asset (GsPlugin *plugin,
...
 	g_debug ("Adding runtime asset with type '%s'", content_type);
 
-	if (g_content_type_is_a (content_type, "application/x-deb")) {
+	if ((type == GS_PLUGIN_EXTERNAL_TYPE_DEB) ||
+	    g_content_type_is_a (content_type, "application/x-deb")) {
 		return add_runtime_deb_asset (build_dir, repo_dir,
 					      download_name, error);
 	}
 
+	if ((priv->type == GS_PLUGIN_EXTERNAL_TYPE_TAR) ||
```
cosimoc: Should be `type`, not `priv->type`

src/plugins/gs-plugin-external-apps.c
```diff
@@ -368,6 +418,8 @@ build_runtime (GsPlugin *plugin,
 	const char *runtime_name = gs_app_get_id_no_prefix (runtime);
 	const char *runtime_url = gs_app_get_metadata_item (runtime,
 							    METADATA_URL);
+	const char *runtime_type = gs_app_get_metadata_item(runtime,
```
cosimoc: Space before paren

src/plugins/gs-plugin-external-apps.c
```diff
@@ -509,6 +563,12 @@ extract_runtime_info_from_json_data (const char *data,
 		return NULL;
 	}
 
+	/* optional elements */
+	type_str = json_object_get_string_member (runtime, JSON_RUNTIME_TYPE_KEY);
+	if (type_str) {
```
cosimoc: `g_strdup()` works just fine with `NULL`, so I think you don't need to check for this here.

src/plugins/gs-plugin-external-apps.c
```diff
@@ -294,17 +294,34 @@ add_runtime_deb_asset (const gchar *build_dir,
 static inline GsPluginExternalType
 get_type_from_string (const char *type)
 {
-	if (type) {
-		if (g_strcmp0(type, "deb") == 0)
-			return GS_PLUGIN_EXTERNAL_TYPE_DEB;
-               else if (g_strcmp0(type, "tar") == 0)
-                       return  GS_PLUGIN_EXTERNAL_TYPE_TAR;
-	}
+
```
joaquimrocha: Unneeded newline here.

src/plugins/gs-plugin-external-apps.c
```diff
@@ -294,17 +294,34 @@ add_runtime_deb_asset (const gchar *build_dir,
...
+add_runtime_tar_asset (const gchar *build_dir,
+		       const gchar *repo_dir,
+		       const char *asset_path,
+		       GError **error)
+{
+	/* flatpak build --runtime needs files in /usr, when coming from a
+	   debian package we can assume some files in /usr, when coming from
+	   a tgz, it's harder, so force that here. */
+	g_autofree gchar *extract_path = g_build_filename (repo_dir, "usr", NULL);
+
```
joaquimrocha: Unneeded newline here.

src/plugins/gs-plugin-external-apps.c
```diff
@@ -393,7 +416,7 @@ build_runtime (GsPlugin *plugin,
 	const char *runtime_name = gs_app_get_id_no_prefix (runtime);
 	const char *runtime_url = gs_app_get_metadata_item (runtime,
 							    METADATA_URL);
-	const char *runtime_type = gs_app_get_metadata_item(runtime,
+	const char *runtime_type = gs_app_get_metadata_item (runtime,
 							    METADATA_TYPE);
```
joaquimrocha: This fix doesn't belong in this commit but rather in the one where Cosimo commented on.
Also, please verify that METADATA_TYPE is well-aligned after the new space introduction (I am not sure if it's right or wrong here due to the mix of tabs and spaces that g-s uses..).

src/plugins/gs-plugin-external-apps.c
```diff
@@ -540,9 +563,7 @@ extract_runtime_info_from_json_data (const char *data,
 
 	/* optional elements */
 	type_str = json_object_get_string_member (runtime, JSON_RUNTIME_TYPE_KEY);
-	if (type_str) {
-		*type = g_strdup (type_str);
-	}
+	*type = g_strdup (type_str);
```
joaquimrocha: This fix also doesn't belong to this commit.
If you need help with git sorcery let me know on Slack. I'll be happy to help.

PR #50, opened 2016-07-29
T12561

joaquimrocha:
https://phabricator.endlessm.com/T12561

### DIFF
--- a/src/plugins/gs-plugin-eos.c
+++ b/src/plugins/gs-plugin-eos.c
@@ -31,7 +31,6 @@
 
 #include "gs-flatpak.h"
 
-#define FLATPAK_EXTRA_CONF_FILE "/etc/gnome-software/flatpak-extra.conf"
 #define FLATPAK_EXTRA_CONF_REMOTE_GROUP_PREFIX "remote:"
 #define FLATPAK_EXTRA_CONF_REMOTE_DEFAULT_BRANCH "default-branch"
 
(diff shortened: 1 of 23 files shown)


src/plugins/gs-plugin-eos.c
```diff
@@ -401,6 +460,38 @@ gs_plugin_adopt_app (GsPlugin *plugin, GsApp *app)
...
+
+	/* if an app has no branch set, maybe it will be set later so we let
+	 * it pass */
+	branch = gs_app_get_flatpak_branch (app);
+	if (!branch)
+		return;
+
+	/* do not show an app if it doesn't belong to the default branch that
+	 * is configured for its remote */
+	if (branch && g_strcmp0 (branch , default_branch) != 0)
```
cosimoc: `default_branch` can't be `NULL` here (you check above), and `g_strcmp0()` will protect you from `branch` being `NULL`, so you can remove the `branch &&` part of this

src/plugins/gs-plugin-eos.c
```diff
@@ -401,6 +460,38 @@ gs_plugin_adopt_app (GsPlugin *plugin, GsApp *app)
...
+	const char *default_branch;
+	const char *branch = NULL;
+	const char *origin = gs_app_get_origin (app);
+	GsPluginData *priv = NULL;
+
+	if (!gs_plugin_eos_app_is_flatpak (app))
+		return;
+
+	priv = gs_plugin_get_data (plugin);
+	default_branch = (char *) g_hash_table_lookup (priv->default_branches,
```
cosimoc: There should be no need for the cast here, also because `default_branch` is `const char *`

src/plugins/gs-plugin-eos.c
```diff
@@ -83,6 +88,57 @@ on_desktop_apps_changed (GDBusConnection *connection,
...
 }
 
+static void
+reload_default_branches (GsPlugin *plugin)
+{
+	GsPluginData *priv = gs_plugin_get_data (plugin);
+	g_autoptr(GKeyFile) config_file = g_key_file_new ();
+	g_autoptr(GError) error = NULL;
+	g_auto(GStrv) groups = NULL;
+	GStrv group;
```
cosimoc: Can you just do the iteration with an index instead? And check for `groups[index] == NULL` as the termination condition

src/plugins/gs-plugin-eos.c
```diff
@@ -83,6 +88,57 @@ on_desktop_apps_changed (GDBusConnection *connection,
...
+	groups = g_key_file_get_groups (config_file, NULL);
+	for (group = groups; group && *group; ++group) {
+		char *remote = NULL;
+		char *default_branch = NULL;
+
+		if (!g_str_has_prefix (*group,
+				       FLATPAK_EXTRA_CONF_REMOTE_GROUP_PREFIX))
+			continue;
+
+		default_branch = g_key_file_get_value (config_file,
```
cosimoc: Should use `g_key_file_get_string()`

PR #54, opened 2016-08-02
T12643

cosimoc:
https://phabricator.endlessm.com/T12643

### DIFF
--- a/src/plugins/gs-flatpak.c
+++ b/src/plugins/gs-flatpak.c
@@ -1867,5 +1867,5 @@ gs_flatpak_get_remotes_names (GsFlatpak *self,
 gboolean
 gs_flatpak_app_is_runtime (GsApp *app)
 {
-	return gs_app_get_flatpak_kind (app) == FLATPAK_REF_KIND_RUNTIME;
+	return g_strcmp0 (gs_app_get_flatpak_kind_as_str (app), "runtime") == 0;
 }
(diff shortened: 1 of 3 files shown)


src/plugins/gs-flatpak.c
```diff
@@ -1867,5 +1867,5 @@ gs_flatpak_get_remotes_names (GsFlatpak *self,
 gboolean
 gs_flatpak_app_is_runtime (GsApp *app)
 {
-	return gs_app_get_flatpak_kind (app) == FLATPAK_REF_KIND_RUNTIME;
+	return g_strcmp0 (gs_app_get_metadata_item (app, "flatpak::kind"), "runtime") == 0;
```
joaquimrocha: There's the helper function gs_app_get_flatpak_kind_as_str(app) which we should here here instead of checking the metadata directly.

src/plugins/gs-plugin-eos.c
```diff
@@ -173,7 +173,7 @@ gs_plugin_initialize (GsPlugin *plugin)
 
 	/* let the flatpak plugins run first so we deal with the apps
 	 * in a more complete/refined state */
-	gs_plugin_add_rule (plugin, GS_PLUGIN_RULE_RUN_AFTER, "flatpak");
+	gs_plugin_add_rule (plugin, GS_PLUGIN_RULE_RUN_AFTER, "flatpak-system");
```
joaquimrocha: Wow! I thought I had changed this when Richard renamed the system plugin...
Thanks! I have just tested (because it could have implications) but everything looks fine.

PR #56, opened 2016-08-02
T12653

joaquimrocha:
https://phabricator.endlessm.com/T12653

### DIFF
--- a/src/plugins/gs-plugin-eos.c
+++ b/src/plugins/gs-plugin-eos.c
@@ -33,6 +33,7 @@
 
 #define FLATPAK_EXTRA_CONF_REMOTE_GROUP_PREFIX "remote:"
 #define FLATPAK_EXTRA_CONF_REMOTE_DEFAULT_BRANCH "default-branch"
+#define ENDLESS_ID_PREFIX "com.endlessm."
 
 /*
  * SECTION:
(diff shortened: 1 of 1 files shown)


src/plugins/gs-plugin-eos.c
```diff
@@ -264,6 +265,155 @@ app_is_renamed (GsApp *app)
...
+		gs_app_add_category (cached_app, "Blacklisted");
+	}
+
+	gs_plugin_cache_add (plugin, locale_app_id, app);
+}
+
+static gboolean
+gs_plugin_eos_blacklist_kapp_if_needed (GsPlugin *plugin, GsApp *app)
+{
+	const guint endless_prefix_len = strlen (ENDLESS_ID_PREFIX);
```
cosimoc: No need for this to be `const`

src/plugins/gs-plugin-eos.c
```diff
@@ -264,6 +265,155 @@ app_is_renamed (GsApp *app)
...
+	g_autofree char *locale_cache_name = NULL;
+	g_auto(GStrv) tokens = NULL;
+	const char *last_token = NULL;
+	const char *locale = gs_plugin_get_locale (plugin);
+	guint num_tokens = 0;
+	/* getting the app name, besides skipping the '.desktop' part of the id
+	 * also makes sure we're dealing with a Flatpak app */
+	const char *app_name = gs_app_get_flatpak_name (app);
+
+	if (!app_name || !locale ||
```
cosimoc: I don't think `locale` can ever be `NULL` here. Other callers also assume it's not NULL

src/plugins/gs-plugin-eos.c
```diff
@@ -264,6 +265,155 @@ app_is_renamed (GsApp *app)
...
+gs_plugin_locale_is_compatible (GsPlugin *plugin,
+				const char *locale)
+{
+	g_autofree char *lang = NULL;
+	const char *plugin_locale = gs_plugin_get_locale (plugin);
+	const char *plugin_lang = gs_plugin_get_language (plugin);
+
+	/* if we don't know the current locale than all locales should be
+	 * compatible */
+	if (!plugin_locale)
```
cosimoc: I don't think this can happen

src/plugins/gs-plugin-eos.c
```diff
@@ -264,6 +265,155 @@ app_is_renamed (GsApp *app)
...
+		return TRUE;
+
```
cosimoc: I think what I would do here is
- call `g_get_locale_variants(locale)`
- for each of the strings returned there, compare them with both `plugin_locale` and `plugin_lang`, returning `TRUE` if either match

This way we don't need to copy `locale_to_language()`, and the code should be a bit simpler to follow

src/plugins/gs-plugin-eos.c
```diff
@@ -264,6 +265,155 @@ app_is_renamed (GsApp *app)
...
+	 * component */
+	if ((g_strcmp0 (lang, plugin_locale) == 0) ||
+	    (g_strcmp0 (plugin_lang, locale) == 0))
+		return TRUE;
+
+	return FALSE;
+}
+
+static char *
+get_app_locale_cache_name (const char *name)
```
cosimoc: I would rename this to `get_app_locale_cache_key` or something like that, to reinforce the fact we're building a cache key

src/plugins/gs-plugin-eos.c
```diff
@@ -264,6 +265,155 @@ app_is_renamed (GsApp *app)
...
+		gs_app_add_category (app, "Blacklisted");
+		return TRUE;
+	}
+
+	tokens = g_strsplit (app_name + endless_prefix_len, ".", 0);
+	num_tokens = g_strv_length (tokens);
+
+	/* we need at least 2 tokens: app-name & locale */
+	if (num_tokens < 2)
+		return FALSE;
```
cosimoc: This block that splits the name should go right after the one where the prefix is checked.
We never want to blacklist any app that don't have a locale component.

src/plugins/gs-plugin-eos.c
```diff
@@ -354,8 +354,7 @@ gs_plugin_eos_blacklist_kapp_if_needed (GsPlugin *plugin, GsApp *app)
 	if (num_tokens < 2)
 		return FALSE;
 
-	/* since the last token should always be 'desktop' in Flatpak, we
-	 * need the one before that, which can be the locale */
+	/* last token may be the locale */
```
joaquimrocha: @cosimoc my comment was not about how the index works. It's an outdated comment from a first version of the code where I was using the 'app-id' which has ends up always in '.desktop', so we'd take the token before the last one in that case.
I can fixup this commit if that's better.

PR #62, opened 2016-08-07
Add debian/postinst which handles some needed system changes for exte…

joaquimrocha:
…rnal apps

For using the external apps in the system we need a privileged user, a
folder for the flatpak repository and to add this repository to the
system flatpak installation.

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -1,0 +1,44 @@
+#!/bin/sh
+
+set -u
+
+g_s_helper_user=g-s-helper
+g_s_helper_home=/var/lib/${g_s_helper_user}
+
+# creating g-s-helper user if it isn't already there
+if ! getent passwd ${g_s_helper_user} >/dev/null; then
+    adduser --system --force-badname --quiet \
+            --no-create-home \
+            --home "${g_s_helper_home}" \
+            --shell /usr/sbin/nologin \
+            --group ${g_s_helper_user}
+fi
+
+# create /var/lib/g-s-helper or update ownership of existing directory
+if [ -d "${g_s_helper_home}" ]; then
+    chown -R ${g_s_helper_user}:${g_s_helper_user} "${g_s_helper_home}"
+else
+    install -d -m0755 -o ${g_s_helper_user} -g ${g_s_helper_user} "${g_s_helper_home}"
+fi
+
+# create external apps repo or update ownership of existing directory
+ext_apps_repo=/var/lib/flatpak-external-apps
+
+if [ -d "${ext_apps_repo}" ]; then
+    chown -R ${g_s_helper_user}:${g_s_helper_user} "${ext_apps_repo}"
+else
+    install -d -m0755 -o ${g_s_helper_user} -g ${g_s_helper_user} "${ext_apps_repo}"
+fi
+
+# add or modify external apps repo if needed
+ext_apps_repo_name=eos-external-apps
+
+flatpak remote-add --no-gpg-verify ${ext_apps_repo_name} "${ext_apps_repo}" 2> /dev/null
+
+if [ $? != 0 ]; then
+    flatpak remote-modify ${ext_apps_repo_name} --no-gpg-verify --url "file://${ext_apps_repo}"
+fi
+
+#DEBHELPER#
+
+exit 0


debian/postinst
```diff
@@ -0,0 +1,45 @@
...
+    adduser --system --force-badname --quiet \
+            --no-create-home \
+            --home ${g_s_helper_home} \
+            --shell /usr/sbin/nologin \
+            --group ${g_s_helper_user}
+fi
+
+# create /var/lib/g-s-helper or update ownership of existing directory
+if [ -d ${g_s_helper_home} ]; then
+    chown -R ${g_s_helper_user}:${g_s_helper_user} ${g_s_helper_home}
```
cosimoc: Quotes around `${g_s_helper_home}` (also below) for consistency

debian/postinst
```diff
@@ -0,0 +1,45 @@
...
+else
+    install -d -m0755 -o ${g_s_helper_user} -g ${g_s_helper_user} ${g_s_helper_home}
+fi
+
+# create external apps folders or update ownership of existing directory
+ext_apps_folder=/opt/flatpak-external-apps
```
cosimoc: Why did you put this in `/opt`? I was thinking it could live in `/var/lib/flatpak-external-apps` next to the other canonical location.

debian/postinst
```diff
@@ -0,0 +1,45 @@
...
+    chown -R ${g_s_helper_user}:${g_s_helper_user} "${ext_apps_folder}"
+else
+    install -d -m0755 -o ${g_s_helper_user} -g ${g_s_helper_user} "${ext_apps_folder}"
+    install -d -m0755 -o ${g_s_helper_user} -g ${g_s_helper_user} "${ext_apps_repo}"
+fi
+
+# add or modify external apps repo if needed
+ext_apps_repo_name=eos-external-apps
+
+flatpak remote-add --no-gpg-verify ${ext_apps_repo_name} ${ext_apps_repo}
```
cosimoc: I think it would be enough for this configuration to only happen in the image builder, since we don't add the other repository from the g-s package itself. No strong feelings either way though.

joaquimrocha: This is just so that us devs also get the external apps working. Otherwise I feel like we'll have bug reports, so I think we should keep it.

debian/postinst
```diff
@@ -0,0 +1,45 @@
+#!/bin/sh
+
+set -eu
+
+g_s_helper_user=g-s-helper
+g_s_helper_home=/var/lib/${g_s_helper_user}
```
dbnicholson: Why not have the home directory be the external apps directory?

debian/postinst
```diff
@@ -0,0 +1,45 @@
...
+else
+    install -d -m0755 -o ${g_s_helper_user} -g ${g_s_helper_user} "${ext_apps_folder}"
+    install -d -m0755 -o ${g_s_helper_user} -g ${g_s_helper_user} "${ext_apps_repo}"
+fi
+
+# add or modify external apps repo if needed
+ext_apps_repo_name=eos-external-apps
+
+flatpak remote-add --no-gpg-verify ${ext_apps_repo_name} ${ext_apps_repo}
+if [ $? != 0 ]; then
```
dbnicholson: This won't work since you're running the script with `-e`.

debian/postinst
```diff
@@ -0,0 +1,45 @@
+#!/bin/sh
+
+set -eu
+
+g_s_helper_user=g-s-helper
```
dbnicholson: Why is the user called `g-s-helper`? Since this is entirely flatpak specific, that seems wrong.

joaquimrocha: I explained that in the Phab task. I am thinking of using the same user for other g-s privileged tasks like downloading the appstream.xml files for categorizing apps. Besides, this is Flatpak related but I wouldn't say it's Flatpak specific as gnome-software is the one who manages the external apps (e.g. calling the build; marking the headless apps as not installed if the runtime extension is missing; etc.)

debian/postinst
joaquimrocha: I have chosen it like this because it's more future proof. I will move the download/build of the apps into the helper user's home but I think the repo should be a different dir. That way we can use the helper's home for other things not external apps related.

PR #65, opened 2016-08-09
T12538

joaquimrocha:
https://phabricator.endlessm.com/T12538

### DIFF
--- a/src/plugins/gs-plugin-eos.c
+++ b/src/plugins/gs-plugin-eos.c
@@ -22,6 +22,7 @@
 #include <config.h>
 
 #include <gio/gdesktopappinfo.h>
+#include <flatpak.h>
 #include <libsoup/soup.h>
 #include <gnome-software.h>
 #include <glib/gi18n.h>
@@ -54,6 +55,8 @@ struct GsPluginData
 	SoupSession *soup_session;
 	GHashTable *default_branches;
 	char *personality;
+	GsFlatpak	*usr_flatpak;
+	GsFlatpak	*sys_flatpak;
 };
 
 static GHashTable *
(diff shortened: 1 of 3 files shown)


src/plugins/gs-plugin-eos.c
```diff
@@ -930,3 +939,169 @@ gs_plugin_refresh (GsPlugin *plugin,
...
+		const char *management;
+		gboolean update_result = FALSE;
+
+		g_debug ("Updating '%s' from proxy '%s' ",
+			 gs_app_get_unique_id (app),
+			 gs_app_get_unique_id (proxy));
+
+		/* set the management plugin momentaneously so we can really
+		 * update it; we reset it back later */
+		management = gs_app_get_management_plugin (app);
```
cosimoc: You may need to strdup this, or it may be freed when you call `gs_app_set_management_plugin()` right after...

src/plugins/gs-plugin-eos.c
```diff
@@ -930,3 +939,169 @@ gs_plugin_refresh (GsPlugin *plugin,
...
+
+		/* in case one of the updates failed we fail too */
+		if (!update_result) {
+			gs_app_set_state_recover (proxy);
+			return FALSE;
+		}
+
+		++num_apps_updated;
+
+		if (g_cancellable_is_cancelled (cancellable))
```
cosimoc: I am not sure it's worth to check separately here the cancellable state, as the next app will fail if it was cancelled. But no big deal one way or the other.

src/plugins/gs-plugin-eos.c
joaquimrocha: Great catch! Here's a new PR for that. Thanks!
https://github.com/endlessm/gnome-software/pull/73

PR #70, opened 2016-08-11
Hide the license details in the details page

joaquimrocha:
This is in order to avoid the user seeing the wrong license that we have
in our Flatpak apps but should be properly fixed later.

https://phabricator.endlessm.com/T12854


rshuler:
@joaquimrocha Looks like you did something weird here and force pushed eos3.0 onto your branch???

joaquimrocha:
Sorry. It was a mistake. Reopened now.

PR #72, opened 2016-08-11
Do not show Updated: Never in the apps' details page

joaquimrocha:
https://phabricator.endlessm.com/T12879


cosimoc:
I think this one would be appropriate for upstream inclusion too.

PR #74, opened 2016-08-11
T11731

rshuler:


dbnicholson:
I think this should be closed in favor of #75 and #76.

PR #77, opened 2016-08-12
T12909

joaquimrocha:
https://phabricator.endlessm.com/T12909


src/plugins/gs-plugin-eos.c
```diff
@@ -387,12 +387,15 @@ get_app_locale_cache_key (const char *app_name)
 	guint name_length = strlen (locale_cache_name);
 	char *suffix = NULL;
 	/* locales can be as long as 5 chars (e.g. pt_PT) so  */
-	if (name_length <= 5)
+	const locale_max_length = 5;
```
cosimoc: Should be `const int` I think

PR #87, opened 2016-08-22
Composite mode enablement

GeorgesStavracas:
This PR adds some minor adjustments to make GNOME Software (mostly) work on composite mode.

https://phabricator.endlessm.com/T11698


src/gs-shell-category.ui
```diff
@@ -127,15 +127,14 @@
...
-                                <property name="margin_start">24</property>
-                                <property name="margin_end">24</property>
+                                <property name="margin_end">12</property>
                                 <property name="margin_top">21</property>
                                 <property name="margin_bottom">21</property>
                                 <property name="halign">center</property>
                                 <property name="visible">True</property>
                                 <property name="can_focus">False</property>
-                                <property name="row_spacing">20</property>
-                                <property name="column_spacing">20</property>
```
joaquimrocha: These values came from Robin's design specs. After you decrease the margins above (which are also from the spec), does it still not fit in composite? (I think that the row spacing is more important for the UX than the margins so we should keep it if it works for us)

src/gnome-software.ui
```diff
@@ -312,6 +312,7 @@
             <child>
               <object class="GtkStack" id="stack_main">
                 <property name="visible">True</property>
+                <property name="homogeneous">False</property>
```
joaquimrocha: @GeorgesStavracas , while this fixes the low-res prob partially for some views, it also breaks the non-composite use a bit by not having homogeneous sizes making the window sometimes increase its size, right?
I would rather get the real fix for the views that are broken than creating a new problem for partially fixing one.

joaquimrocha:
@GeorgesStavracas Your feature-tile patch is not needed since we are not using that widget at all (we are using the GsImageTile) so it only makes the differences with upstream bigger for no reason.

PR #91, opened 2016-08-24
Cleanup external apps setup on purge

dbnicholson:
When purging, the external apps setup and g-s-helper user should be
removed.

https://phabricator.endlessm.com/T12949


mcatanzaro:
Looks good to me

PR #97, opened 2016-08-29
external-apps: Really run enable service

dbnicholson:
ConditionNeedsUpdate only works if the service runs before
systemd-update-done.service. Unfortunately, this also means running with
DefaultDependencies=no in sysinit.target. Since this operation is so
lightweight (it just run flatpak remote-add and flatpak remote-modify),
just have it run on every boot.


dbnicholson:
Wrong issue.

PR #113, opened 2016-09-17
T12725

GeorgesStavracas:
This PR make GNOME Software:
 - Show compulsory (aka core) apps with 'eos-' suffixed icon
 - Show apps with low score on appstream (ie Writer, Presentation, Evolution, etc)

https://phabricator.endlessm.com/T12725

### DIFF
--- a/src/gs-plugin-loader.c
+++ b/src/gs-plugin-loader.c
@@ -858,12 +858,6 @@ gs_plugin_loader_filter_qt_for_gtk (GsApp *app, gpointer user_data)
 }
 
 static gboolean
-gs_plugin_loader_app_is_non_compulsory (GsApp *app, gpointer user_data)
-{
-	return !gs_app_has_quirk (app, AS_APP_QUIRK_COMPULSORY);
-}
-
-static gboolean
 gs_plugin_loader_app_is_non_installed (GsApp *app, gpointer user_data)
 {
 	if (gs_app_get_state (app) == AS_APP_STATE_INSTALLED)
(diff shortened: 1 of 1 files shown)


src/gs-plugin-loader.c
```diff
@@ -858,9 +858,31 @@ gs_plugin_loader_filter_qt_for_gtk (GsApp *app, gpointer user_data)
...
+	/*
+	 * If the app has an 'eos-' prefixed icon, we always include it regardless
+	 * of the COMPULSORY quirk.
+	 */
+	for (i = 0; i < icons->len; i++) {
+		AsIcon *icon = g_ptr_array_index (icons, i);
+
+		if (g_str_has_prefix (as_icon_get_name (icon), "eos-")) {
+			return TRUE;
+			break;
```
rshuler: This break is unreachable, but I suppose it doesn't hurt anything.

src/gs-plugin-loader.c
```diff
@@ -858,9 +858,31 @@ gs_plugin_loader_filter_qt_for_gtk (GsApp *app, gpointer user_data)
...
+	}
+
+	return FALSE;
+}
+
+static gboolean
 gs_plugin_loader_app_is_non_compulsory (GsApp *app, gpointer user_data)
 {
-	return !gs_app_has_quirk (app, AS_APP_QUIRK_COMPULSORY);
+	return !gs_app_has_quirk (app, AS_APP_QUIRK_COMPULSORY) || gs_plugin_loader_app_has_endless_icon (app);
```
cosimoc: I think that it's the EOS plugin that adds the compulsory quirk [1] for all the non-flatpaks. At least this code should be moved in there, but it's not very clear to me why only non-compulsory apps don't show in the categories. Perhaps that is what should be fixed instead?

[1] https://github.com/endlessm/gnome-software/blob/master/src/plugins/gs-plugin-eos.c#L600

src/plugins/gs-plugin-appstream.c
```diff
@@ -745,8 +745,6 @@ gs_plugin_add_categories (GsPlugin *plugin,
 		app = g_ptr_array_index (array, i);
 		if (as_app_get_id (app) == NULL)
 			continue;
-		if (as_app_get_priority (app) < 0)
```
cosimoc: Is this change alone not sufficient for the core apps to load in the categories?
Would this not be needed if we had appstream data for the core apps?

GeorgesStavracas: It'd be fixed, I guess. I'm not particularly sure how AppStream calculates the priority of the apps.

joaquimrocha:
This was not the correct direction so I have closed the PR.

PR #114, opened 2016-09-20
Perform Flatpak updates according the download-updates setting

joaquimrocha:
Even if there is a download-updates setting, the Flatpak plugins are not
honoring it, meaning that Flatpak updates are always downloaded on
refresh time (and deployed later).

With these changes, the original behavior is still used if the mentioned
setting is true. Otherwise, it shows updates for refs that even if they
do not have downloaded updates and fetch those when the user
deliberately updates the apps.

https://phabricator.endlessm.com/T13407

### DIFF
--- a/src/plugins/gs-flatpak.c
+++ b/src/plugins/gs-flatpak.c
@@ -42,6 +42,7 @@ struct _GsFlatpak {
 	GFileMonitor		*monitor;
 	GsFlatpakScope		 scope;
 	GsPlugin		*plugin;
+	gboolean		 download_updates;
 };
 
 G_DEFINE_TYPE (GsFlatpak, gs_flatpak, G_TYPE_OBJECT)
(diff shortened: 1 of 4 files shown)


src/plugins/gs-plugin-flatpak-user.c
```diff
@@ -39,13 +39,31 @@ struct GsPluginData {
...
 void
 gs_plugin_initialize (GsPlugin *plugin)
 {
 	GsPluginData *priv = gs_plugin_alloc_data (plugin, sizeof(GsPluginData));
+	gboolean download_updates;
+
 	priv->flatpak = gs_flatpak_new (plugin, GS_FLATPAK_SCOPE_USER);
 	priv->settings = g_settings_new ("org.gnome.software");
 
+	gs_flatpak_set_download_updates (priv->flatpak, download_updates);
```
cosimoc: `download_updates` will be uninitialized here as far as I can see

src/plugins/gs-flatpak.c
```diff
@@ -1654,16 +1734,22 @@ gs_flatpak_update_app (GsFlatpak *self,
...
 
 	/* only process this app if was created by this plugin */
 	if (g_strcmp0 (gs_app_get_management_plugin (app),
 		       gs_plugin_get_name (self->plugin)) != 0)
 		return TRUE;
 
+	/* if 'download updates' is enabled, then we just deploy the installed
+	 * updates instead of fetching any eventual new ones */
+	if (self->download_updates)
+		update_flags = FLATPAK_UPDATE_FLAGS_NO_PULL;
```
cosimoc: Use `|=` for future-proofing

src/plugins/gs-flatpak.c
```diff
@@ -653,24 +687,66 @@ gs_flatpak_add_updates (GsFlatpak *self, GsAppList *list,
...
+	if (0&&!flatpak_installation_drop_caches (self->installation,
+						  cancellable,
+						  error)) {
+		return FALSE;
+	}
+
+	if (!gs_flatpak_add_downloaded_updates (self, list, cancellable,
+						error)) {
+		g_debug ("Failed to get downloaded updates: %s",
+			 (*error)->message);
```
cosimoc: `error` may not be set here. I would use a local error and `g_propagate_error()` if you want to print this here.

src/plugins/gs-flatpak.c
```diff
@@ -653,24 +687,66 @@ gs_flatpak_add_updates (GsFlatpak *self, GsAppList *list,
...
+	/* if 'download updates' is enabled, then we just deploy the installed
+	 * updates instead, so we should only show as updatable apps that have
+	 * downloaded updates */
+	if (self->download_updates)
+		return TRUE;
+
+	if (!gs_flatpak_add_available_updates (self, list, cancellable,
+					       error)) {
+		g_debug ("Failed to get available updates: %s",
+			 (*error)->message);
```
cosimoc: Same as above.

src/plugins/gs-flatpak.c
```diff
@@ -653,24 +687,66 @@ gs_flatpak_add_updates (GsFlatpak *self, GsAppList *list,
...
+gboolean
+gs_flatpak_add_updates (GsFlatpak *self, GsAppList *list,
+			GCancellable *cancellable,
+			GError **error)
+{
+	guint i;
+	g_autoptr(GPtrArray) xrefs = NULL;
+
+	/* manually drop the cache */
+	if (0&&!flatpak_installation_drop_caches (self->installation,
```
cosimoc: I know that `0&&` was here already... But is it intended?

src/plugins/gs-flatpak.c
```diff
@@ -653,24 +687,66 @@ gs_flatpak_add_updates (GsFlatpak *self, GsAppList *list,
...
 	return TRUE;
 }
 
+gboolean
+gs_flatpak_add_updates (GsFlatpak *self, GsAppList *list,
+			GCancellable *cancellable,
+			GError **error)
+{
+	guint i;
+	g_autoptr(GPtrArray) xrefs = NULL;
```
cosimoc: These variables look unused here

src/plugins/gs-flatpak.c
```diff
@@ -615,21 +616,54 @@ gs_flatpak_app_install_source (GsFlatpak *self, GsApp *app,
...
+								     error);
+
+	if (xrefs == NULL)
 		return FALSE;
+
+	for (i = 0; i < xrefs->len; i++) {
+		FlatpakInstalledRef *xref = g_ptr_array_index (xrefs, i);
+		g_autoptr(GsApp) app = NULL;
+		g_autoptr(GError) error_local = NULL;
+		g_autoptr(FlatpakInstalledRef) xref2 = NULL;
```
cosimoc: This is unused

src/plugins/gs-flatpak.c
```diff
@@ -2039,3 +2126,10 @@ gs_flatpak_app_is_runtime (GsApp *app)
 {
 	return g_strcmp0 (gs_app_get_flatpak_kind_as_str (app), "runtime") == 0;
 }
+
+void
+gs_flatpak_set_download_updates (GsFlatpak *self,
+				 gboolean download_updates)
+{
+	self->download_updates = download_updates;
```
cosimoc: Is any refresh needed at this point, if you set a different new value?

joaquimrocha: No refresh needs to happen IMO (as it won't really compromise anything: we always check the flag directly before performing ops). In any case a refresh would have to be done from outside this module.

src/plugins/gs-flatpak.c
joaquimrocha: Good question! I'd rather not change that as part of these changes and fix it separately after checking with Richard.

PR #125, opened 2016-10-14
Fix the creation of the extra AppStream files

joaquimrocha:
https://phabricator.endlessm.com/T12454


src/plugins/gs-plugin-external-appstream.c
```diff
@@ -181,6 +182,11 @@ gs_plugin_external_appstream_refresh_url (GsPlugin *plugin,
 		return FALSE;
 	}
 
+	/* close the file */
+	if (!g_output_stream_close (outstream, cancellable, error)) {
```
cosimoc: I would close this in any event right after the call to `g_output_stream_write_all()`, even when that fails, so that we can prevent `g_file_delete()` failing because of dangling reference.

src/plugins/gs-plugin-external-appstream.c
```diff
@@ -169,13 +170,22 @@ gs_plugin_external_appstream_refresh_url (GsPlugin *plugin,
...
-	if (!g_output_stream_write_all (outstream, msg->response_body->data,
-					msg->response_body->length, NULL,
-					cancellable, error)) {
+	file_written = g_output_stream_write_all (outstream,
+						  msg->response_body->data,
+						  msg->response_body->length,
+						  NULL, cancellable, error);
+
+	/* close the file */
+	if (!g_output_stream_close (outstream, cancellable, error)) {
```
cosimoc: You can't reuse `error` here, it may already have been set.
I would just always close the stream without error checking, not returning if it fails.

PR #128, opened 2016-10-19
T13763

rshuler:
https://phabricator.endlessm.com/T13763


joaquimrocha:
@rshuler , the labels patch looks fine. The buttons vs links one is not the right approach IMO. I'd rather change the shortcuts button to a button and move it to the right position, and the same to the website in two atomic patches.

PR #130, opened 2016-10-19
external-apps: Improve the removal of apps

joaquimrocha:
This patch changes the order of the removal for external apps: now the
headless app is removed before the external runtime because if it fails
or is cancelled for some reason, the app may still be used.
It also uses the helper script to remove the external runtime so that
the removal process doesn't ask for the user's password twice.

https://phabricator.endlessm.com/T13807

### DIFF
--- a/src/plugins/gs-plugin-external-apps.c
+++ b/src/plugins/gs-plugin-external-apps.c
@@ -638,39 +638,57 @@ gs_plugin_app_remove (GsPlugin *plugin,
 		      GError **error)
 {
 	GsApp *ext_runtime;
-	GsPluginData *priv;
+	const char *runtime_id;
 	GsFlatpak *flatpak = NULL;
+	g_autoptr(GError) local_error = NULL;
 
 	/* only process this app if was created by this plugin */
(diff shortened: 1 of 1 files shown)


src/plugins/gs-plugin-external-apps.c
```diff
@@ -638,39 +638,57 @@ gs_plugin_app_remove (GsPlugin *plugin,
...
+	}
+
+	g_debug ("Removing external runtime %s",
+		 gs_app_get_unique_id (ext_runtime));
+
+	if (!remove_external_runtime (ext_runtime, cancellable, &local_error)) {
+		g_debug ("Removed app %s but cannot remove external runtime "
+			 "'%s': %s.", gs_app_get_unique_id (app),
+			 gs_app_get_unique_id (ext_runtime),
+			 local_error->message);
```
cosimoc: Discussed with @joaquimrocha on Slack, we may want to set the app state to UNKNOWN for the external runtime when it's removed.

cosimoc: That is being addressed on the other PR for external apps.

PR #137, opened 2016-10-31
T13950

joaquimrocha:
https://phabricator.endlessm.com/T13950

### DIFF
--- a/src/gs-plugin-loader.c
+++ b/src/gs-plugin-loader.c
@@ -71,6 +71,12 @@ enum {
 	SIGNAL_LAST
 };
 
+enum {
+	PROP_0,
+	PROP_ALLOW_UPDATES,
+	PROP_LAST
+};
+
 static guint signals [SIGNAL_LAST] = { 0 };
 
 typedef void		 (*GsPluginFunc)		(GsPlugin	*plugin);
@@ -3297,6 +3303,13 @@ gs_plugin_loader_app_action_finish (GsPluginLoader *plugin_loader,
 
 /******************************************************************************/
 
+gboolean
+gs_plugin_loader_get_allow_updates (GsPluginLoader *plugin_loader)
+{
+	GsPluginLoaderPrivate *priv = gs_plugin_loader_get_instance_private (plugin_loader);
+	return g_settings_get_boolean (priv->settings, "allow-updates");
+}
+
 GsAppList *
 gs_plugin_loader_get_pending (GsPluginLoader *plugin_loader)
 {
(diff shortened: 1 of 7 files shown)


src/gs-plugin-loader.c
```diff
@@ -3835,6 +3849,34 @@ gs_plugin_loader_dump_state (GsPluginLoader *plugin_loader)
...
+		g_value_set_boolean (value, gs_plugin_loader_get_allow_updates (plugin_loader));
+		break;
+	default:
+		G_OBJECT_WARN_INVALID_PROPERTY_ID (object, prop_id, pspec);
+		break;
+	}
+}
+
+static void
+gs_plugin_loader_set_property (GObject *object, guint prop_id,
```
cosimoc: Does not look like this is needed.

src/gs-plugin-loader.c
```diff
@@ -3877,11 +3919,19 @@ gs_plugin_loader_finalize (GObject *object)
 static void
 gs_plugin_loader_class_init (GsPluginLoaderClass *klass)
 {
+	GParamSpec *pspec;
 	GObjectClass *object_class = G_OBJECT_CLASS (klass);
 
+	object_class->get_property = gs_plugin_loader_get_property;
+	object_class->set_property = gs_plugin_loader_set_property;
```
cosimoc: See comment above

src/gs-plugin-loader.c
```diff
@@ -3925,6 +3990,8 @@ gs_plugin_loader_init (GsPluginLoader *plugin_loader)
 	priv->auth_array = g_ptr_array_new_with_free_func ((GFreeFunc) g_object_unref);
 	priv->profile = as_profile_new ();
 	priv->settings = g_settings_new ("org.gnome.software");
+	g_signal_connect (priv->settings, "changed",
```
cosimoc: If you connect to `changed::allow-updates` you can remove the check in the callback (but it's OK if upstream has it this way and you don't want to change it)

src/gs-plugin-loader.c
```diff
@@ -3909,6 +3959,21 @@ gs_plugin_loader_class_init (GsPluginLoaderClass *klass)
 }
 
 static void
+gs_plugin_loader_settings_changed_cb (GSettings *settings,
+				      const gchar *key,
+				      GsPluginLoader *plugin_loader)
+{
+	GsPluginLoaderPrivate *priv = gs_plugin_loader_get_instance_private (plugin_loader);
+	if (g_strcmp0 (key, "allow-updates") == 0) {
+		priv->allow_updates = g_settings_get_boolean (priv->settings,
```
cosimoc: There's really no need to store this in a private field. Calling `g_settings_get()` is fast, and you can do it directly from the property getter.

src/gs-update-monitor.c
```diff
@@ -413,19 +415,57 @@ check_thrice_daily_cb (gpointer data)
...
+
 static gboolean
 check_updates_on_startup_cb (gpointer data)
 {
 	GsUpdateMonitor *monitor = data;
 
 	g_debug ("First hourly updates check");
-	check_updates (monitor);
-	get_upgrades (monitor);
+	restart_updates_check (monitor);
```
cosimoc: Why does this not honor the allow-updates check?

joaquimrocha: This commit is already upstream. The reason I don't make dependable on allow-updates is that this is not really a check_updates, it's something more of a _refresh_and_eventually_get_updates_. Richard and I have agreed to rename gs-updates-monitor to gs-schedule which is less misguiding ATM but that will happen later.

src/gs-plugin-loader.c
```diff
@@ -58,7 +58,7 @@ typedef struct
 
 	guint			 updates_changed_id;
 	guint			 reload_id;
-	gboolean		 online; 
+	gboolean		 online;
```
cosimoc: Nitpick: remove this spurious change

PR #138, opened 2016-11-02
eos-plugin: show desktop when adding a shortcut

GeorgesStavracas:
The previous app store used to show the desktop right after
adding a shortcut to an application, which showed a nice
animation and effectively teaches the user where the new
shortcut is.

After moving to GNOME Software, however, this behavior is gone.

Fix that by calling the FocusSearch DBus call (for the lack of
a better call), which in turn shows the desktop.

https://phabricator.endlessm.com/T13764


pwithnall:
Given that the behaviour of the app grid is changing a bit in the shell now, is this still wanted, @GeorgesStavracas?

GeorgesStavracas:
> Given that the behaviour of the app grid is changing a bit in the shell now, is this still wanted, @GeorgesStavracas?

IIRC this pull request was restoring the behavior that Endless OS had when installing apps from the old app center. I think this is still desired, but not a priority right now. I also do not have time to rebase this pull requests.

pwithnall:
OK. https://phabricator.endlessm.com/T13764 is still open, and let’s leave this PR open so someone can rebase it in future whenever they’re assigned to T13764.

mwleeds:
As with #346, this can be closed since we no longer have the ability to add or remove things to the desktop using the App Center.

PR #141, opened 2016-11-08
Handle default branches using flatpak's API only

mariospr:
Drop the need on our custom system based on the `/etc/gnome-software/flatpak-extra.conf` file and use flatpak's new APIs instead to check the default-branch attribute for a remote, pulling it from the server's summary file when needed.

https://phabricator.endlessm.com/T13403

### DIFF
--- a/src/plugins/gs-flatpak.c
+++ b/src/plugins/gs-flatpak.c
@@ -2333,6 +2333,33 @@ gs_flatpak_refine_metadata_from_installation (GsFlatpak *self,
 	return TRUE;
 }
 
+static void
+ensure_default_branches (GsFlatpak *self)
+{
+	g_autoptr(GPtrArray) xremotes = NULL;
+	guint i;
+
+	xremotes = flatpak_installation_list_remotes (self->installation, NULL, NULL);
+	if (xremotes == NULL || xremotes->len == 0)
+		return;
+
+	for (i = 0; i < xremotes->len; i++) {
+		FlatpakRemote *xremote = g_ptr_array_index (xremotes, i);
+		const gchar *remote_name = flatpak_remote_get_name (xremote);
+		g_autofree char *default_branch = NULL;
+		g_autoptr(GError) error_local = NULL;
+
+		default_branch = flatpak_remote_get_default_branch (xremote);
+		if (default_branch != NULL)
+			continue;
+
+		if (!flatpak_installation_update_remote_sync (self->installation, remote_name, NULL, &error_local)) {
+			g_warning ("failed to update metadata from remote %s: %s",
+				   remote_name, error_local->message);
+		}
+	}
+}
+
 void
 gs_flatpak_fill_default_branches (GsFlatpak *self, GHashTable *table)
 {
@@ -2345,6 +2372,9 @@ gs_flatpak_fill_default_branches (GsFlatpak *self, GHashTable *table)
 	xremotes = flatpak_installation_list_remotes (self->installation, NULL, NULL);
 	if (xremotes == NULL || xremotes->len == 0)
 		return;
+
+	/* Make sure that default branches are pulled from the server if needed */
+	ensure_default_branches (self);
 
 	for (i = 0; i < xremotes->len; i++) {
 		FlatpakRemote *xremote = g_ptr_array_index (xremotes, i);
--- a/src/plugins/gs-flatpak.c
+++ b/src/plugins/gs-flatpak.c
@@ -2332,3 +2332,32 @@ gs_flatpak_refine_metadata_from_installation (GsFlatpak *self,
 	gs_appstream_copy_metadata (app, as_app, TRUE);
 	return TRUE;
 }
+
+void
+gs_flatpak_fill_default_branches (GsFlatpak *self, GHashTable *table)
+{
+	g_autoptr(GPtrArray) xremotes = NULL;
+	guint i;
+
+	if (self->installation == NULL)
+		return;
+
+	xremotes = flatpak_installation_list_remotes (self->installation, NULL, NULL);
+	if (xremotes == NULL || xremotes->len == 0)
+		return;
+
+	for (i = 0; i < xremotes->len; i++) {
+		FlatpakRemote *xremote = g_ptr_array_index (xremotes, i);
+		char *default_branch = flatpak_remote_get_default_branch (xremote);
+		gchar *remote_name = NULL;
+
+		/* filter by branch */
+		if (default_branch == NULL)
+			continue;
+
+		remote_name = g_strdup (flatpak_remote_get_name (xremote));
+		g_hash_table_insert (table, remote_name, default_branch);
+		g_debug ("Found default branch '%s' for remote '%s'",
+			 default_branch, remote_name);
+	}
+}
--- a/src/plugins/gs-flatpak.h
+++ b/src/plugins/gs-flatpak.h
@@ -24,6 +24,7 @@
 #define __GS_FLATPAK_H
 
 #include <gnome-software.h>
+#include <flatpak.h>
 
 G_BEGIN_DECLS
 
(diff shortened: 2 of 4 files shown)


src/plugins/gs-flatpak.c
```diff
@@ -2332,3 +2332,9 @@ gs_flatpak_refine_metadata_from_installation (GsFlatpak *self,
 	gs_appstream_copy_metadata (app, as_app, TRUE);
 	return TRUE;
 }
+
+FlatpakInstallation *
+gs_flatpak_get_installation (GsFlatpak *self)
```
joaquimrocha: So far none the FlatpakInstallation is never exposed outside of GsFlatpak and I'd prefer to keep using this object to abstract the Flatpak stuff. Also, many of the code of GsFlatpak will be changed/removed so it makes sense to add this temporary code there.

The code will also be much simpler and easier to change when we rebase if we just call something like _gs_flatpak_app_is_default_branch_ or something.

src/plugins/gs-plugin-eos.c
```diff
@@ -106,6 +106,32 @@ on_desktop_apps_changed (GDBusConnection *connection,
...
+		return;
+
+	for (i = 0; i < xremotes->len; i++) {
+		FlatpakRemote *xremote = g_ptr_array_index (xremotes, i);
+		const gchar *remote_name = flatpak_remote_get_name (xremote);
+		g_autofree char *default_branch = NULL;
+		g_autoptr(GError) error_local = NULL;
+
+		default_branch = flatpak_remote_get_default_branch (xremote);
+		if (default_branch == NULL) {
```
joaquimrocha: We usually invert the check and continue in cases like this.
Either that or join both _if_ statements in one:
`if (default_branch == NULL && !flatpak_installation.... )`

src/plugins/gs-plugin-eos.c
```diff
@@ -105,79 +105,50 @@ on_desktop_apps_changed (GDBusConnection *connection,
...
+	for (i = 0; i < xremotes->len; i++) {
+		FlatpakRemote *xremote = g_ptr_array_index (xremotes, i);
+		char *default_branch = flatpak_remote_get_default_branch (xremote);
 
-		default_branch = g_key_file_get_string (config_file,
-							group,
-							FLATPAK_EXTRA_CONF_REMOTE_DEFAULT_BRANCH,
-							NULL);
+		/* filter by branch */
+		if (default_branch != NULL) {
```
joaquimrocha: We usually invert the if and call continue in these cases.

src/plugins/gs-plugin-eos.c
```diff
@@ -760,12 +737,21 @@ gs_plugin_eos_blacklist_by_branch_if_needed (GsPlugin *plugin, GsApp *app)
...
+	g_autoptr(GSettings) settings = NULL;
 
 	if (!gs_app_is_flatpak (app))
 		return FALSE;
 
 	priv = gs_plugin_get_data (plugin);
-	default_branch = g_hash_table_lookup (priv->default_branches, origin);
+
+	/* do we want to filter to the default branch */
+	if (!g_settings_get_boolean (priv->settings, "filter-default-branch"))
```
joaquimrocha: This should also belong to GsFlatpak AFAIU.

src/plugins/gs-plugin-eos.c
mariospr: Ok, I'll change that

src/plugins/gs-plugin-eos.c
mariospr: Ok. Will change that

src/plugins/gs-plugin-eos.c
mariospr: Again, this is something that will only be needed before the rebase. Once we rebase this check will come for free in gs-flatpak already, see https://git.gnome.org/browse/gnome-software/commit/?id=5fa03af8

mariospr: That's a good point indeed. I added this because in principle this was meant to be a backport of Richard's patch upstream... then ended up being a complete different thing.

I'll remove that and also the new schema changes

src/plugins/gs-flatpak.h
```diff
@@ -24,6 +24,7 @@
 #define __GS_FLATPAK_H
 
 #include <gnome-software.h>
+#include <flatpak.h>
```
cosimoc: This does not look needed anymore

src/plugins/gs-plugin-eos.c
```diff
@@ -760,12 +737,21 @@ gs_plugin_eos_blacklist_by_branch_if_needed (GsPlugin *plugin, GsApp *app)
 	const char *branch = NULL;
 	const char *origin = gs_app_get_origin (app);
 	GsPluginData *priv = NULL;
+	g_autoptr(GSettings) settings = NULL;
```
cosimoc: This comment still applies

src/plugins/gs-plugin-eos.c
```diff
@@ -105,79 +103,28 @@ on_desktop_apps_changed (GDBusConnection *connection,
...
-					    "flatpak-extra.conf", NULL);
-	config_file = load_branches_config_file (extra_conf_file);
-
-	if (!config_file) {
-		/* Try loading the file from our pkgdatadir instead */
-		g_free (extra_conf_file);
-		extra_conf_file = g_build_filename (GS_DATA, "flatpak-extra.conf", NULL);
+	GHashTable *default_branches = NULL;
+	GsFlatpak *flatpak = NULL;
+	guint i;
```
cosimoc: This looks unused

src/plugins/gs-flatpak.c
```diff
@@ -2333,6 +2333,30 @@ gs_flatpak_refine_metadata_from_installation (GsFlatpak *self,
...
 }
 
+static void
+ensure_default_branches (GsFlatpak *self)
+{
+	g_autoptr(GPtrArray) xremotes = NULL;
+	guint i;
+
+	xremotes = flatpak_installation_list_remotes (self->installation, NULL, NULL);
+	for (i = 0; i < xremotes->len; i++) {
```
cosimoc: You should check for `xremotes != NULL` and return early, since `flatpak_installation_list_remotes()` can fail

src/plugins/gs-flatpak.c
```diff
@@ -2333,6 +2333,30 @@ gs_flatpak_refine_metadata_from_installation (GsFlatpak *self,
...
+		const gchar *remote_name = flatpak_remote_get_name (xremote);
+		g_autofree char *default_branch = NULL;
+		g_autoptr(GError) error_local = NULL;
+
+		default_branch = flatpak_remote_get_default_branch (xremote);
+		if (default_branch != NULL)
+			continue;
+
+		if (!flatpak_installation_update_remote_sync (self->installation, remote_name, NULL, &error_local)) {
+			g_warning ("failed to update metatada from remote %s: %s",
```
cosimoc: `metadata`

src/plugins/gs-plugin-eos.c
mariospr: Yes, it's a leftover

src/plugins/gs-plugin-eos.c
mariospr: Another one, oops

src/plugins/gs-flatpak.h
mariospr: Actually it is needed for `FlatpakProgressCallback`. I'm not 100% sure how this built before, though, but it definitely does not build if I don't add it.

src/plugins/gs-flatpak.c
mariospr: You're right, good point

src/plugins/gs-flatpak.c
mariospr: ok

PR #142, opened 2016-11-08
Set minimum required version of flatpak to 0.6.13

mariospr:
This is required for flatpak_installation_update_remote_sync ()

https://phabricator.endlessm.com/T13403-debian


joaquimrocha:
This is good to go but only makes sense after the pending #141 so I will wait till I merge the former.

PR #144, opened 2016-11-08
external-apps: Double-check/reset a cached external runtime's state

joaquimrocha:
This is defensive code that should be able to fix an issue where an
installed external runtime has its state as "unknown".

https://phabricator.endlessm.com/T14075

### DIFF
--- a/src/plugins/gs-plugin-external-apps.c
+++ b/src/plugins/gs-plugin-external-apps.c
@@ -558,11 +558,22 @@ gs_plugin_refine_app (GsPlugin *plugin,
 		 * for download and this refine was not requested by the
 		 * details view, then we hide it as it will not be usable */
 		if ((flags & GS_PLUGIN_REFINE_FLAGS_DETAILS_VIEW) == 0) {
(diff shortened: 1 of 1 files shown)


src/plugins/gs-plugin-external-apps.c
```diff
@@ -375,10 +388,19 @@ get_external_runtime_from_json (GsPlugin *plugin,
 
 	runtime = gs_plugin_cache_lookup (plugin, full_id);
 	if (runtime) {
-		g_debug ("Found cached '%s'", full_id);
 		gs_app_set_management_plugin (runtime,
 					      gs_plugin_get_name (plugin));
 
+		if (gs_flatpak_is_installed (priv->sys_flatpak, runtime, NULL,
```
cosimoc: I think that this block of code and the one at the end of this function should be refactored together.

So basically you'd have
```
if (runtime) {
  g_debug("...");
} else {
  // create runtime here
}

// check installed state and set mgmt plugin
return runtime;
```

PR #145, opened 2016-11-08
Fix stale screenshots when they are cached

joaquimrocha:
https://phabricator.endlessm.com/T13859

### DIFF
--- a/src/gs-screenshot-image.c
+++ b/src/gs-screenshot-image.c
@@ -199,6 +199,85 @@ gs_screenshot_image_show_blurred (GsScreenshotImage *ssimg,
 	}
 }
 
+static gboolean
+gs_screenshot_image_save_downloaded_img (GsScreenshotImage *ssimg,
+					 GdkPixbuf *pixbuf,
+					 GError **error)
+{
+	g_autoptr(AsImage) im = NULL;
+	gboolean ret;
+	const GPtrArray *images;
+	g_autoptr(GError) local_error = NULL;
+	g_autofree char *filename = NULL;
+	g_autofree char *size_dir = NULL;
+	g_autofree char *cache_kind = NULL;
+	g_autofree char *basename = NULL;
+	guint width = ssimg->width;
+	guint height = ssimg->height;
+
+	/* save to file, using the same code as the AppStream builder
+	 * so the preview looks the same */
+	im = as_image_new ();
+	as_image_set_pixbuf (im, pixbuf);
+	ret = as_image_save_filename (im, ssimg->filename,
+				      ssimg->width * ssimg->scale,
+				      ssimg->height * ssimg->scale,
+				      AS_IMAGE_SAVE_FLAG_PAD_16_9,
+				      error);
+
+	if (!ret)
+		return FALSE;
+
+	if (ssimg->screenshot == NULL)
+		return TRUE;
+
+	images = as_screenshot_get_images (ssimg->screenshot);
+	if (images->len > 1)
+		return TRUE;
+
+	if (width == AS_IMAGE_THUMBNAIL_WIDTH &&
+	    height == AS_IMAGE_THUMBNAIL_HEIGHT) {
+		width = AS_IMAGE_NORMAL_WIDTH;
+		height = AS_IMAGE_NORMAL_HEIGHT;
+	} else {
+		width = AS_IMAGE_THUMBNAIL_WIDTH;
+		height = AS_IMAGE_THUMBNAIL_HEIGHT;
+	}
+
+	width *= ssimg->scale;
+	height *= ssimg->scale;
+	basename = g_path_get_basename (ssimg->filename);
+	size_dir = g_strdup_printf ("%ux%u", width, height);
+	cache_kind = g_build_filename ("screenshots", size_dir, NULL);
+	filename = gs_utils_get_cache_filename (cache_kind, basename,
+						GS_UTILS_CACHE_FLAG_WRITEABLE,
+						&local_error);
+
+        if (filename == NULL) {
+		/* if we cannot get a cache filename, warn about that but do not
+		 * set a user's visible error because this is a complementary
+		 * operation */
+                g_warning ("Failed to get cache filename for counterpart "
+                           "screenshot '%s' in folder '%s': %s", basename,
+                           cache_kind, local_error->message);
+                return TRUE;
+        }
+
+	ret = as_image_save_filename (im, filename, width, height,
+				      AS_IMAGE_SAVE_FLAG_PAD_16_9,
+				      &local_error);
+
+	if (!ret) {
+		/* if we cannot save this screenshot, warn about that but do not
+		 * set a user's visible error because this is a complementary
+		 * operation */
+                g_warning ("Failed to save screenshot '%s': %s", filename,
+                           local_error->message);
+        }
+
+	return TRUE;
+}
+
 static void
 gs_screenshot_image_complete_cb (SoupSession *session,
 				 SoupMessage *msg,
@@ -255,19 +334,10 @@ gs_screenshot_image_complete_cb (SoupSession *session,
 			gs_screenshot_image_set_error (ssimg, error->message);
 			return;
 		}
-	} else {
-		/* save to file, using the same code as the AppStream builder
-		 * so the preview looks the same */
-		im = as_image_new ();
-		as_image_set_pixbuf (im, pixbuf);
-		ret = as_image_save_filename (im, ssimg->filename,
-					      ssimg->width * ssimg->scale,
-					      ssimg->height * ssimg->scale,
-					      AS_IMAGE_SAVE_FLAG_PAD_16_9, &error);
-		if (!ret) {
-			gs_screenshot_image_set_error (ssimg, error->message);
-			return;
-		}
+	} else if (!gs_screenshot_image_save_downloaded_img (ssimg, pixbuf,
+							     &error)) {
+		gs_screenshot_image_set_error (ssimg, error->message);
+		return;
 	}
 
 	/* got image, so show */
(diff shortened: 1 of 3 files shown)


src/gs-utils.c
```diff
@@ -503,4 +503,31 @@ gs_utils_get_wilson_rating (guint64 star1,
 	return (gint) ceil (val);
 }
 
+gboolean
```
cosimoc: This should either have an out `GError **` parameter or return `void`

src/gs-screenshot-image.c
```diff
@@ -214,7 +280,26 @@ gs_screenshot_image_complete_cb (SoupSession *session,
...
+		g_debug ("Screenshot for '%s' has not been modified; not "
+			 "downloading again", url);
+		return;
+	}
 	if (msg->status_code != SOUP_STATUS_OK) {
+		SoupURI *uri = soup_message_get_uri (msg);
+		g_autofree char *url = soup_uri_to_string (uri, FALSE);
+		g_warning ("Result of downloading screenshot from "
+			   "'%s': %s", url,
+			   soup_status_get_phrase (msg->status_code));
```
cosimoc: Documentation of this function says
>There is no reason for you to ever use this function. If you wanted the textual description for the “status_code” of a given SoupMessage, you should just look at the message's “reason_phrase”.

src/gs-screenshot-image.c
```diff
@@ -214,7 +280,26 @@ gs_screenshot_image_complete_cb (SoupSession *session,
...
+
+		/* if we are already showing a screenshot, do not show and
+		 * error */
+		if (ssimg->status == GS_SCREENSHOT_IMAGE_STATUS_IMAGE)
+			return;
```
cosimoc: It may be worth to wrap this check inside `gs_screenshot_image_set_error()` itself? Or in another function that calls into it?

src/gs-screenshot-image.c
```diff
@@ -368,26 +471,42 @@ gs_screenshot_image_load_async (GsScreenshotImage *ssimg,
...
 		gs_screenshot_image_set_error (ssimg, _("Could not create cache"));
 		return;
 	}
 
-	/* does local file already exist */
-	if (g_file_test (ssimg->filename, G_FILE_TEST_EXISTS)) {
-		as_screenshot_show_image (ssimg);
-		return;
+	if (ssimg->filename != NULL) {
+		g_free (ssimg->filename);
```
cosimoc: `g_free()` works fine on `NULL`

src/gs-screenshot-image.c
```diff
@@ -368,26 +471,42 @@ gs_screenshot_image_load_async (GsScreenshotImage *ssimg,
...
-	if (ssimg->filename == NULL) {
+
+	/* request the writable cache filename before trying the system's one
+	 * because the former should be the more recent */
+	writable_filename = gs_utils_get_cache_filename (cache_kind,
+							 basename,
+							 GS_UTILS_CACHE_FLAG_WRITEABLE,
+							 NULL);
+
+	if (writable_filename == NULL) {
```
cosimoc: Don't you still want to check the system cache here?

src/gs-screenshot-image.c
cosimoc: But how can `ssimg->filename` not be `NULL` here? The only place we set it is above, but we either return or clear it...

src/gs-screenshot-image.c
```diff
@@ -368,26 +471,42 @@ gs_screenshot_image_load_async (GsScreenshotImage *ssimg,
...
+	if (ssimg->filename != NULL) {
+		g_free (ssimg->filename);
+	}
+
+	/* if the user's cache file does not exist, try to use the system one */
+	if (!g_file_test (writable_filename, G_FILE_TEST_EXISTS)) {
+		ssimg->filename = gs_utils_get_cache_filename (cache_kind,
+							       basename,
+							       GS_UTILS_CACHE_FLAG_NONE,
+							       NULL);
```
cosimoc: Should we do anything when this returns `NULL`?

src/gs-screenshot-image.c
```diff
@@ -368,26 +471,42 @@ gs_screenshot_image_load_async (GsScreenshotImage *ssimg,
...
+	} else {
+		ssimg->filename = g_steal_pointer (&writable_filename);
 	}
 
-	/* can we load a blurred smaller version of this straight away */
-	if (ssimg->width > AS_IMAGE_THUMBNAIL_WIDTH &&
-	    ssimg->height > AS_IMAGE_THUMBNAIL_HEIGHT) {
+	/* show the screenshot if its cached version exists, otherwise try to
+	 * load a blurred smaller version while we fetch the new image */
+	if (g_file_test (ssimg->filename, G_FILE_TEST_EXISTS)) {
```
cosimoc: You probably only want to do this check when in the code path of the first case in the above if

src/gs-screenshot-image.c
```diff
@@ -435,12 +552,33 @@ gs_screenshot_image_load_async (GsScreenshotImage *ssimg,
...
+			g_warning ("Could not setup a new soup_message for the "
+				   "URL '%s'. Showing cached image '%s'.", url,
+				   ssimg->filename);
+			return;
+		}
+
+		/* silently return as we have already a screenshot set, even if
+		 * eventually outdated */
+		if (showing_image)
+			return;
```
cosimoc: Either this or the block above should be removed

src/gs-screenshot-image.c
```diff
@@ -435,12 +552,33 @@ gs_screenshot_image_load_async (GsScreenshotImage *ssimg,
...
+
 		/* TRANSLATORS: this is when networking is not available */
 		gs_screenshot_image_set_error (ssimg, _("Screenshot not available"));
 		return;
 	}
 
+	if (showing_image)
+		gs_utils_set_modified_request (ssimg->message, ssimg->filename);
+
+	if (writable_filename != NULL) {
```
cosimoc: Didn't we do this already above? Why do you need to do it again?

src/gs-screenshot-image.c
```diff
@@ -481,6 +619,7 @@ gs_screenshot_image_init (GsScreenshotImage *ssimg)
 		atk_object_set_role (accessible, ATK_ROLE_IMAGE);
 		atk_object_set_name (accessible, _("Screenshot"));
 	}
+	ssimg->status = GS_SCREENSHOT_IMAGE_STATUS_UNKNOWN;
```
cosimoc: You should not need to explicitly initialize this, since its value is zero

src/gs-screenshot-image.c
```diff
@@ -250,18 +337,30 @@ gs_screenshot_image_complete_cb (SoupSession *session,
...
 		 * so the preview looks the same */
 		im = as_image_new ();
 		as_image_set_pixbuf (im, pixbuf);
+
 		ret = as_image_save_filename (im, ssimg->filename,
 					      ssimg->width * ssimg->scale,
 					      ssimg->height * ssimg->scale,
-					      AS_IMAGE_SAVE_FLAG_PAD_16_9, &error);
+					      AS_IMAGE_SAVE_FLAG_PAD_16_9,
+					      &error);
```
cosimoc: This change does not seem needed

src/gs-screenshot-image.c
```diff
@@ -250,18 +337,30 @@ gs_screenshot_image_complete_cb (SoupSession *session,
...
 			return;
 		}
+
+		/* if the kind is unknown then we save also the counterpart size
+		 * image in order to save an extra download when the user
+		 * requests that size later */
+		kind = as_image_get_kind (im);
+		if (kind == AS_IMAGE_KIND_UNKNOWN) {
+			gs_screenshot_image_save_counterpart (ssimg, im);
+		}
```
cosimoc: Remove curly braces for single line blocks.

src/gs-screenshot-image.c
```diff
@@ -250,18 +337,30 @@ gs_screenshot_image_complete_cb (SoupSession *session,
...
+					      AS_IMAGE_SAVE_FLAG_PAD_16_9,
+					      &error);
 		if (!ret) {
 			gs_screenshot_image_set_error (ssimg, error->message);
 			return;
 		}
+
+		/* if the kind is unknown then we save also the counterpart size
+		 * image in order to save an extra download when the user
+		 * requests that size later */
```
cosimoc: Why does the "unknown" kind imply that we want to save both versions? Also I'm a bit confused why we need to do this at all. Won't it potentially upscale images, making them look bad?

src/gs-screenshot-image.c
```diff
@@ -196,6 +209,59 @@ gs_screenshot_image_show_blurred (GsScreenshotImage *ssimg,
...
+	    height == AS_IMAGE_THUMBNAIL_HEIGHT) {
+		width = AS_IMAGE_NORMAL_WIDTH;
+		height = AS_IMAGE_NORMAL_HEIGHT;
+	} else {
+		width = AS_IMAGE_THUMBNAIL_WIDTH;
+		height = AS_IMAGE_THUMBNAIL_HEIGHT;
+	}
+
+	size_dir = g_strdup_printf ("%ux%u", width * ssimg->scale,
+				    height * ssimg->scale);
```
cosimoc: You could save scaled width and height in separate variables, since they're used more than once

src/gs-screenshot-image.c
cosimoc: Is it because we would end up scaling and saving them in any case, when the kind is unknown?

src/gs-screenshot-image.c
joaquimrocha: Last minute refining rush, sorry for the mistake...

src/gs-screenshot-image.c
joaquimrocha: No. We do this when the writable_filename is valid _and_ exists. So at this point, the writable_filename may point to the user's cache path _but_ we're actually showing the system cache's image (and have filename assigned as such).

src/gs-screenshot-image.c
joaquimrocha: I prefer to be explicit in such cases :smiley:

PR #146, opened 2016-11-09
T13211

GeorgesStavracas:
https://phabricator.endlessm.com/T13211

### DIFF
--- a/src/gs-shell.c
+++ b/src/gs-shell.c
@@ -784,6 +784,20 @@ gs_shell_set_mode (GsShell *shell, GsShellMode mode)
 }
 
 GsShellMode
+gs_shell_get_previous_mode (GsShell *shell)
+{
+	GsShellPrivate *priv = gs_shell_get_instance_private (shell);
+	BackEntry *entry;
+
+	entry = g_queue_peek_head (priv->back_entry_stack);
+
+	if (entry == NULL)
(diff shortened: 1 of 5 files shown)


src/gs-shell.c
```diff
@@ -784,6 +784,20 @@ gs_shell_set_mode (GsShell *shell, GsShellMode mode)
...
+
+	entry = g_queue_peek_head (priv->back_entry_stack);
+
+	if (entry == NULL)
+		return GS_SHELL_MODE_UNKNOWN;
+
+	return entry->mode;
+}
+
+GsShellMode
```
joaquimrocha: Should we just abstract the code in _gs_shell_back_button_cb_ and call it when the user exits search?

joaquimrocha:
Hi @GeorgesStavracas ! Thanks for this nice change.
Please change the commits' titles to replace the module/subject word by a full sentence as that's the way it's done upstream. Downstream we only do that for plugins: e.g. _external-apps_, _eos_, etc.

src/gs-shell.c
GeorgesStavracas: I don't think it'll be worth it. The back button code goes beyond the :get_previous_mode() method as it 1. removes the queue's head and 2. use more members of the queue data other than the mode.

PR #148, opened 2016-11-11
T14102

joaquimrocha:
https://phabricator.endlessm.com/T14102

### DIFF
--- a/src/plugins/eos-external-apps-build-install
+++ b/src/plugins/eos-external-apps-build-install
@@ -110,7 +110,6 @@ sed -i s/Application/Runtime/ "$BUNDLEDIR/metadata"
 
 # Extract the asset into the build files
 mkdir -p "$BUNDLEDIR/usr"
-cd $BUILDDIR
 
 if [ "$ASSETTYPE" = "deb" ]; then
     ar x "$assetfile"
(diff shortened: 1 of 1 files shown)


src/plugins/eos-external-apps-build-install
```diff
@@ -121,6 +121,11 @@ else
     exit 1
 fi
 
+# Fix permissions on export dir because some debs' tarballs base directory
+# has read permissions only for the user, and thus wouldn't be usable as a
+# Flatpak runtime
+chmod a+rx "$EXPORTDIR"
```
cosimoc: Nitpick: you could omit `a`. But do you really also need execution permissions, in addition to read?

src/plugins/eos-external-apps-build-install
```diff
@@ -82,10 +82,11 @@ EXTAPPSBUILDDIR="$HOME/external-apps-build"
 APPDIR="$EXTAPPSBUILDDIR/$APPNAME"
 BUILDDIR="$APPDIR/build"
 BUNDLEDIR="$APPDIR/bundle"
+EXPORTDIR="$BUNDLEDIR/usr"
```
cosimoc: Looks like this belongs to the next commit?

src/plugins/eos-external-apps-build-install
```diff
@@ -82,10 +82,11 @@ EXTAPPSBUILDDIR="$HOME/external-apps-build"
...
 
 # Remove any previous left build dir and create it again
 rm -rf "$BUNDLEDIR"
-mkdir -p "$BUILDDIR" "$BUNDLEDIR"
+mkdir -p "$BUILDDIR"
```
cosimoc: Not sure this agrees with the commit message... You mention it was being called with the build directory twice, but it was actually called with both the build and bundle directories.

src/plugins/eos-external-apps-build-install
joaquimrocha: The execution bit on a directory is for listing and we need it.

PR #152, opened 2016-11-16
eos: Use fallback branches when no default ones are fetched

joaquimrocha:
This is done only for the Endless remotes and is a solution for when
existing users upgrade the OS and boot into an offline state (meaning
they won't be able to fetch the default branches).

https://phabricator.endlessm.com/T14206

### DIFF
--- a/src/plugins/gs-plugin-eos.c
+++ b/src/plugins/gs-plugin-eos.c
@@ -121,9 +121,16 @@ reload_default_branches (GsPlugin *plugin, GsFlatpakScope scope)
 	g_hash_table_remove_all (default_branches);
 	gs_flatpak_fill_default_branches (flatpak, default_branches);
 
-	if (g_hash_table_size (default_branches) == 0)
-		g_debug ("No default branches configured for %s!",
-			 scope == GS_FLATPAK_SCOPE_USER ? "USER" : "SYSTEM");
+	if (scope == GS_FLATPAK_SCOPE_SYSTEM &&
+	    !g_hash_table_lookup (default_branches, EOS_APPS_REMOTE_NAME)) {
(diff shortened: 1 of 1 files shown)


src/plugins/gs-plugin-eos.c
```diff
@@ -121,9 +121,14 @@ reload_default_branches (GsPlugin *plugin, GsFlatpakScope scope)
...
-		g_debug ("No default branches configured for %s!",
-			 scope == GS_FLATPAK_SCOPE_USER ? "USER" : "SYSTEM");
+	if (g_hash_table_size (default_branches) == 0) {
+		g_warning ("No default branches configured for Flatpak '%s' "
+			   "installation! Using fallback branches for Endless "
+			   "remotes (eos3)...",
+			   scope == GS_FLATPAK_SCOPE_USER ? "user" : "system");
+		g_hash_table_insert (priv->sys_default_branches,
+				     g_strdup ("eos-apps"), g_strdup ("eos3"));
+	}
```
mariospr: I don't think this will work as intended, since you won't insert the fallback in case there is already some default branch defined for a different remote (e.g. I have one default-branch defined for a non-endless remote), as the `g_hash_table_size (default_branches) == 0` condition won't evaluate to `TRUE` (even if there's no default-branch defined for `eos-apps` yet).

Btw, what about a simpler approach such as https://github.com/endlessm/gnome-software/pull/153/commits/47d3ea3918964d0049c3ecd7dc69d792939aa6df (I've tested it and works nicely too)? The link in the commit message is wrong because I wrote it before realizing you did this, but if you like it I can change that.

mariospr: Sounds good to me too. Already closed my other PR in favour of this one

PR #153, opened 2016-11-16
Fallback to using the 'eos3' branch as the default for the eos-apps remote

mariospr:
This is a small patch to make sure users upgrading from EOS < 3.1 don't ever
see multiple versions of the apps if, for any reason, they run then new App
Center while being offline and before having successfully pulled the default
branch extended attribute from the summary file, in the OSTree server.

NOTE: This must be removed at some point before releasing EOS 4, of course.

https://phabricator.endlessm.com/T14167


mariospr:
Closing this PR in favor of a reviesed version of https://github.com/endlessm/gnome-software/pull/152 after talking to @joaquimrocha, who made me realize my approach is quite inefficient

PR #156, opened 2016-11-17
eos: Blacklist app if incompatible with the OS version

joaquimrocha:
Even though we use Flatpak for managing our applications and that
is independent from the OS version, there are some applications that
may use in a hybrid way: we use Flatpak for the distribution but we
launch them from outside which means that perhaps the OS in question
doesn't have the app helper that launched it. For that reason we want
to be able to blacklist apps if they are not compatible with the OS.

https://phabricator.endlessm.com/T14235

### DIFF
--- a/src/plugins/gs-plugin-eos.c
+++ b/src/plugins/gs-plugin-eos.c
@@ -59,6 +59,7 @@ struct GsPluginData
 	GHashTable *usr_default_branches;
 	GHashTable *sys_default_branches;
 	char *personality;
+	char *os_version_id;
 	GsFlatpak	*usr_flatpak;
 	GsFlatpak	*sys_flatpak;
 };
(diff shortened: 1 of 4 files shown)


src/plugins/gs-plugin-eos.c
```diff
@@ -514,6 +533,25 @@ app_is_proxy (GsApp *app)
...
+		return TRUE;
+
+	app_available_since =
+		gs_app_get_metadata_item (app, "EndlessOS::available-since");
+	if (!app_available_since)
+		return TRUE;
+
+	/* if the OS version is greater than the app "available-since"
+	 * metadata item, it means it is incompatible */
+	return as_utils_vercmp (priv->os_version_id, app_available_since) > 0;
```
cosimoc: `as_utils_vercmp()` is documented as returning positive values when the first argument is a greater version than the second.
So the code gets the check right (if OS version is e.g. 3.1 it means it's compatible with an app available since 3.0.7), but the comment is wrong. Additionally, if both of those are the same version (function returning zero) it should still say the app is compatible. So the check should be `>= 0` I think

joaquimrocha: Good catch. T

PR #159, opened 2016-11-22
Return auth error in the Flatpak plugins when auth fails on removal

joaquimrocha:
If the authentication fails when the user is removing a Flatpak, we need
to return a plugin "authentication failed" error as any other error will
be ignored by the plugin loader.

https://phabricator.endlessm.com/T14240

### DIFF
--- a/src/plugins/gs-flatpak.c
+++ b/src/plugins/gs-flatpak.c
@@ -1630,6 +1630,8 @@ gs_flatpak_app_remove (GsFlatpak *self,
 		       GCancellable *cancellable,
 		       GError **error)
 {
+	GError *local_error = NULL;
+
 	/* only process this app if was created by this plugin */
 	if (g_strcmp0 (gs_app_get_management_plugin (app),
 		       gs_plugin_get_name (self->plugin)) != 0)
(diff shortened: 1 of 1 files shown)


src/plugins/gs-flatpak.c
```diff
@@ -1657,8 +1659,24 @@ gs_flatpak_app_remove (GsFlatpak *self,
...
+				     G_DBUS_ERROR_AUTH_FAILED) ||
+		    g_error_matches (local_error, G_DBUS_ERROR,
+				     G_DBUS_ERROR_ACCESS_DENIED)) {
+			g_set_error (error, GS_PLUGIN_ERROR,
+				     GS_PLUGIN_ERROR_AUTH_REQUIRED,
+				     "Authentication problem when uninstalling "
+				     "'%s': %s", gs_app_get_unique_id (app),
+				     local_error->message);
+		}
+		g_propagate_error (error, local_error);
```
cosimoc: This should be in an else block, or it will undo the effects of the block above

joaquimrocha: Yeah, sorry.

PR #163, opened 2016-12-06
Add plugins for cleaning up the legacy external runtimes

joaquimrocha:
The plugin "legacy-external-apps" gets the info about the external
runtime from the installed AppStream data and runs before the "flatpak"
plugin so we can get this info before the app is changed
(removed/updated). That info is set as a metadata item in the GsApp
object.

The plugin "external-runtimes-cleaner" uses the info mentioned above
to remove the external runtimes once an app is updated or removed. It
runs after the Flatpak plugin and ensures that such runtimes are only
removed if the app has been successfully updated/removed.

https://phabricator.endlessm.com/T14442

### DIFF
--- a/src/plugins/Makefile.am
+++ b/src/plugins/Makefile.am
@@ -101,6 +101,8 @@ endif
 
 if HAVE_FLATPAK
 plugin_LTLIBRARIES += libgs_plugin_flatpak.la
+plugin_LTLIBRARIES += libgs_plugin_legacy-external-apps.la
+plugin_LTLIBRARIES += libgs_plugin_external-runtimes-cleaner.la
 endif
 
 if HAVE_OSTREE
@@ -227,7 +229,33 @@ libgs_plugin_flatpak_la_SOURCES =			\
 libgs_plugin_flatpak_la_LIBADD = $(GS_PLUGIN_LIBS) $(FLATPAK_LIBS)
 libgs_plugin_flatpak_la_LDFLAGS = -module -avoid-version
 libgs_plugin_flatpak_la_CFLAGS = $(GS_PLUGIN_CFLAGS) $(WARN_CFLAGS)
-endif
+
+libgs_plugin_legacy_external_apps_la_SOURCES =	\
+	gs-appstream.c				\
+	gs-appstream.h				\
+	gs-flatpak.c				\
+	gs-flatpak.h				\
+	gs-flatpak-symlinks.c			\
+	gs-flatpak-symlinks.h			\
+	gs-legacy-external-apps.h		\
+	gs-plugin-legacy-external-apps.c
+libgs_plugin_legacy_external_apps_la_LIBADD = $(GS_PLUGIN_LIBS) $(FLATPAK_LIBS) $(JSON_GLIB_LIBS)
+libgs_plugin_legacy_external_apps_la_LDFLAGS = -module -avoid-version
+libgs_plugin_legacy_external_apps_la_CFLAGS = $(GS_PLUGIN_CFLAGS) $(JSON_GLIB_CFLAGS) $(WARN_CFLAGS)
+
+libgs_plugin_external_runtimes_cleaner_la_SOURCES =		\
+	gs-appstream.c				\
+	gs-appstream.h				\
+	gs-flatpak.c				\
+	gs-flatpak.h				\
+	gs-flatpak-symlinks.c			\
+	gs-flatpak-symlinks.h			\
+	gs-legacy-external-apps.h		\
+	gs-plugin-external-runtimes-cleaner.c
+libgs_plugin_external_runtimes_cleaner_la_LIBADD = $(GS_PLUGIN_LIBS) $(FLATPAK_LIBS) $(JSON_GLIB_LIBS)
+libgs_plugin_external_runtimes_cleaner_la_LDFLAGS = -module -avoid-version
+libgs_plugin_external_runtimes_cleaner_la_CFLAGS = $(GS_PLUGIN_CFLAGS) $(JSON_GLIB_CFLAGS) $(WARN_CFLAGS)
+endif # HAVE_FLATPAK
 
 if HAVE_OSTREE
 libgs_plugin_ostree_la_SOURCES = gs-plugin-ostree.c
--- a/src/plugins/gs-legacy-external-apps.h
+++ b/src/plugins/gs-legacy-external-apps.h
@@ -1,0 +1,29 @@
+/* -*- Mode: C; tab-width: 8; indent-tabs-mode: t; c-basic-offset: 8 -*-
+ *
+ * Copyright (C) 2016 Endless Mobile, Inc
+ *
+ * Author: Joaquim Rocha <jrocha@endlessm.com>
+ *
+ * Licensed under the GNU General Public License Version 2
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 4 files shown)


src/plugins/Makefile.am
```diff
@@ -221,7 +223,31 @@ libgs_plugin_flatpak_la_SOURCES =			\
...
+	gs-flatpak.c				\
+	gs-flatpak.h				\
+	gs-flatpak-symlinks.c			\
+	gs-flatpak-symlinks.h			\
+	gs-plugin-legacy-external-apps.c
+libgs_plugin_legacy_external_apps_la_LIBADD = $(GS_PLUGIN_LIBS) $(FLATPAK_LIBS) $(JSON_GLIB_LIBS)
+libgs_plugin_legacy_external_apps_la_LDFLAGS = -module -avoid-version
+libgs_plugin_legacy_external_apps_la_CFLAGS = $(GS_PLUGIN_CFLAGS) $(JSON_GLIB_CFLAGS) $(WARN_CFLAGS)
+
+libgs_plugin_external_runtimes_cleaner_la_SOURCES =		\
```
cosimoc: Can you add the word "legacy" also in the name of this plugin? To make it clear it applies to the same legacy ext apps as the other plugin.

src/plugins/gs-plugin-external-runtimes-cleaner.c
```diff
@@ -0,0 +1,239 @@
...
+#include <config.h>
+#include <flatpak.h>
+#include <glib.h>
+#include <gnome-software.h>
+#include <string.h>
+
+#include "gs-appstream.h"
+#include "gs-flatpak.h"
+
+#define RUNTIME_INSTALLED_MTD_KEY "EndlessOS::legacy-ext-runtime-installed"
```
cosimoc: Nit: use same define name you use in other plugin (or even better, move to a common header?)

src/plugins/gs-plugin-legacy-external-apps.c
```diff
@@ -0,0 +1,220 @@
...
+	GsPluginData *priv = gs_plugin_get_data (plugin);
+	g_object_unref (priv->installation);
+}
+
+static gboolean
+app_is_legacy_external_app (GsApp *app)
+{
+	guint i;
+	for (i = 0; LEGACY_EXTERNAL_APPS[i] != NULL; ++i) {
+		const char *curren_id = LEGACY_EXTERNAL_APPS[i];
```
cosimoc: Nit: `current_id`

src/plugins/gs-plugin-legacy-external-apps.c
```diff
@@ -0,0 +1,220 @@
...
+
+	/* we setup the version of the external runtime used by this external
+	 * app so it is later used by "external-apps-cleaner" plugin when
+	 * removing those runtimes; we use a new key and not the one that is
+	 * already set in the metadata so we verify that this key has been
+	 * set by this plugin (and thus, by what was installed) and not by
+	 * the general AppStream data */
+	gs_app_set_metadata (app, LEGACY_RUNTIME_INSTALLED_MTD_KEY, NULL);
+	gs_app_set_metadata (app, LEGACY_RUNTIME_INSTALLED_MTD_KEY,
+			     runtime_version);
```
cosimoc: Any reason why you need to call this twice in a row?

src/plugins/Makefile.am
joaquimrocha: I pondered that and I decided to not include it because the name would be too long and also external runtimes is a term only used on our legacy external apps (i.e. making _legacy_ redundant). That with _cleaner_ seemed enough for me to identify the plugin as something not to keep forever (there's also a short description in the license comment). If you still prefer it, I can of course add it.

src/plugins/gs-plugin-external-runtimes-cleaner.c
joaquimrocha: It seemed a bit overkill to add another header just for one single define but I can do that.

src/plugins/gs-plugin-legacy-external-apps.c
joaquimrocha: Typo + emacs completion. Thanks.

src/plugins/gs-plugin-legacy-external-apps.c
joaquimrocha: This is because Richard likes these safety checks for some operations. It prevents assigning metadata already assigned.
I disagree with it and have a recurring discussion about it with Richard but for now we have to go with that.

src/plugins/Makefile.am
```diff
@@ -221,7 +223,33 @@ libgs_plugin_flatpak_la_SOURCES =			\
...
-endif
+
+libgs_plugin_legacy_external_apps_la_SOURCES =	\
+	gs-appstream.c				\
+	gs-appstream.h				\
+	gs-flatpak.c				\
+	gs-flatpak.h				\
+	gs-flatpak-symlinks.c			\
+	gs-flatpak-symlinks.h			\
+	gs-legacy-external-apps.h		\
```
cosimoc: I think you forgot to add this file to the PR.

src/plugins/Makefile.am
cosimoc: OK, sounds good then.

src/plugins/gs-plugin-legacy-external-apps.c
```diff
@@ -0,0 +1,220 @@
...
+			   GsApp *app,
+			   GCancellable *cancellable,
+			   GError **error)
+{
+	const char *runtime_version;
+	g_autoptr(GError) local_error = NULL;
+	g_autoptr(AsApp) as_app = NULL;
+
+	if (!app_is_legacy_external_app (app))
+		return TRUE;
```
cosimoc: Would it not be enough to just check for the presence of `LEGACY_RUNTIME_MTD_KEY` instead of hardcoding this list?

src/plugins/Makefile.am
joaquimrocha: Right. Sorry about that.

src/plugins/gs-plugin-legacy-external-apps.c
joaquimrocha: That would mean that the key would/could come from the general appstream which is not sync'ed with the app's own appstream. Since we can drop it in the future the general appstream while users may still have it in the installed one, we need to use the installed one. And using the installed one means reading from disk, for all apps; to save that I filter the apps by the ones that need it.

src/plugins/gs-legacy-external-apps.h
```diff
@@ -0,0 +1,29 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#ifndef __GS_SNAPD_H__
```
cosimoc: Whoops! :-)

src/plugins/gs-plugin-legacy-external-apps.c
cosimoc: OK... yeah I guess it would be slower to perform the installed check, but it would not be for all apps, right? Only for the app being updated or removed?

joaquimrocha: Only catching up with this now.
Yes, you are right and we're comparing the IDs of all apps that are being installed/updated/remove, however my reluctance is that we're talking about IO rather than just some check in memory and IMO the code complexity wins the trade-off here.

cosimoc: Okay, let's keep it like this

PR #165, opened 2016-12-15
eos: Use a proxy app for updating a couple of applications

joaquimrocha:
This way we can have many apps that can be confusing for the user, such
as service apps, that are updated seamlessly and possibly together with
others.

Currently the applications used in the proxy are currently the
com.endlessm.Platform runtime and the com.endlessm.EknServices app.

https://phabricator.endlessm.com/T12538

### DIFF
--- a/src/plugins/gs-plugin-eos.c
+++ b/src/plugins/gs-plugin-eos.c
@@ -41,6 +41,8 @@
 #define EOS_IMAGE_VERSION_ALT_PATH "/"
 
 #define METADATA_SYS_DESKTOP_FILE "flatpak-3rdparty::system-desktop-file"
+#define EOS_PROXY_APP_PREFIX ENDLESS_ID_PREFIX "proxy"
+
 /*
  * SECTION:
  * Plugin to improve GNOME Software integration in the EOS desktop.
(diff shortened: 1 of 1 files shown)


src/plugins/gs-plugin-eos.c
```diff
@@ -873,3 +882,81 @@ gs_plugin_launch (GsPlugin *plugin,
...
+	gs_app_list_add (list, updates_proxy_app);
+
+	return TRUE;
+}
+
+gboolean
+gs_plugin_add_updates (GsPlugin *plugin,
+			       GsAppList *list,
+			       GCancellable *cancellable,
+			       GError **error)
```
cosimoc: Indentation

src/plugins/gs-plugin-eos.c
```diff
@@ -873,3 +882,81 @@ gs_plugin_launch (GsPlugin *plugin,
...
+{
+	/* only the gs_plugin_add_updates_pending should be used in EOS
+	 * but in case the user has changed the "download-updates" then this
```
cosimoc: Nit: `the "download-updates` -> `the "download-updates" setting`

PR #167, opened 2016-12-16
T14596

joaquimrocha:
https://phabricator.endlessm.com/T14596

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -258,7 +258,7 @@ AC_ARG_ENABLE(flatpak,
               enable_flatpak=maybe)
 AS_IF([test "x$enable_flatpak" != "xno"], [
     PKG_CHECK_MODULES(FLATPAK,
-                      [flatpak >= 0.6.12],
+                      [flatpak >= 0.6.13],
                       [have_flatpak=yes],
                       [have_flatpak=no])
 ], [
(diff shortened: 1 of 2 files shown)


mariospr:
This is a backport of the changes from https://github.com/endlessm/gnome-software/commit/05795ec876e06333621f524391becef2dbff1cc4 split in two commits, so surely they look good to me. 

The only nitpick I have is that, being exactly the same changes I would have kept the authorship when backporting them (at least the one with actual code changes) by committing with `--author="Mario Sanchez Prada <mario@endlessm.com>`, mainly because it helps identifying who did that when running `git blame`, and also because it's common practice while backporting patches.

If you do that, I'd remove the "This is based on a patch by Mario Sanchez related to T13403" comment in the commit log, and maybe expand that comment a bit to explain the why, maybe something like this: _Ensure that the default-branch attribute is pulled from the server's summary file before finding the default branches for the available remotes, so that GNOME Software can filter out flatpaks from other branches._

@joaquimrocha Would you mind doing those small changes? Otherwise it looks good to me

cosimoc:
@mariospr Joaquim is off until Tuesday so can you please implement the suggestion and update the PR?

mariospr:
Sure, done, now merging

PR #177, opened 2017-01-06
Install needed Flatpak runtimes when updating an app

joaquimrocha:
When an app is updated it may require a new runtime (or new branch),
thus we need to also check for that and install any new runtimes
accordingly.


rshuler:
For the record, the branch number was incorrect.  This PR should have been associated with https://phabricator.endlessm.com/T14732

PR #178, opened 2017-01-06
Blacklist more GNOME flatpaks: Iagno, Maps, Polari, Documents and Epiphany

mariospr:
Iagno is also shipped by our OS. Maps, Polari, Documents and Epiphany
all present issues that need to be addressed upstream before being
actually usable as flatpaks.

https://phabricator.endlessm.com/T14720


src/plugins/gs-flatpak.c
```diff
@@ -318,9 +339,10 @@ gs_flatpak_add_apps_from_xremote (GsFlatpak *self,
 			continue;
 		}
 
-		/* filter duplicated apps that we provide via OSTree */
-		if (app_is_duplicated_gnome_flatpak (app, self->scope, xremote)) {
-			g_debug ("not adding duplicated app from GNOME repository: %s",
+		g_debug ("[AAA] %s", as_app_get_unique_id (app));
```
rshuler: Is this a stray extra debug message that was only intended during development?

mariospr: Oops! Fixed

PR #180, opened 2017-01-10
Temporarily added GNOME Weather to the list of blacklisted apps

mariospr:
The idea is to use GNOME Weather from the GNOME flatpak repository
but we can't do it now because we need it to be patched so that it
allows overriding its application ID (needed for Endless Coding).

See the upstream bug: https://bugzilla.gnome.org/show_bug.cgi?id=776315

In the meantime we are creating our own flatpak for GNOME Weather,
so we need to blacklist the one from GNOME to avoid duplicates.

https://phabricator.endlessm.com/T14669


mariospr:
In the end we committed this change associating it to T14720, so this PR is closed

PR #181, opened 2017-01-10
eos: Fix launching Google Chrome

joaquimrocha:
Google Chrome needs to be launched from outside the Flatpak sandbox so
the EOS plugin now adds a special key to the Chrome app which indicates
that it uses a special desktop file in order to be launched.
It also overrides the launch method in the EOS plugin to be able to
launch Chrome.

https://phabricator.endlessm.com/T14847

### DIFF
--- a/src/plugins/gs-plugin-eos.c
+++ b/src/plugins/gs-plugin-eos.c
@@ -40,7 +40,7 @@
 #define EOS_IMAGE_VERSION_PATH "/sysroot"
 #define EOS_IMAGE_VERSION_ALT_PATH "/"
 
-#define METADATA_SYS_DESKTOP_FILE "flatpak-3rdparty::system-desktop-file"
+#define METADATA_SYS_DESKTOP_FILE "EndlessOS::system-desktop-file"
 #define EOS_PROXY_APP_PREFIX ENDLESS_ID_PREFIX "proxy"
 
 /*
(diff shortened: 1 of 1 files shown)


src/plugins/gs-plugin-eos.c
```diff
@@ -863,18 +871,47 @@ gs_plugin_app_install (GsPlugin *plugin,
...
 }
 
+static gboolean
+launch_with_sys_desktop_file (GsApp *app,
+                              GError **error)
+{
+        GdkDisplay *display;
+	g_autoptr(GAppLaunchContext) context = NULL;
+	g_autoptr(GDesktopAppInfo) app_info = get_desktop_app_info (app);
+	g_autoptr(GError) local_error = NULL;
```
mariospr: Nit : indentation in the previous 3 line is broken

src/plugins/gs-plugin-eos.c
```diff
@@ -863,18 +871,47 @@ gs_plugin_app_install (GsPlugin *plugin,
...
+
+        display = gdk_display_get_default ();
+        context = G_APP_LAUNCH_CONTEXT (gdk_display_get_app_launch_context (display));
+        ret = g_app_info_launch (G_APP_INFO (app_info), NULL, context, &local_error);
+
+	if (!ret) {
+		g_warning ("Could not launch %s: %s", gs_app_get_unique_id (app),
+			   local_error->message);
+		g_set_error (error, GS_PLUGIN_ERROR, GS_PLUGIN_ERROR_FAILED,
+			     _("Could not launch this application."));
```
mariospr: Nit : indentation in the previous 3 line is also broken

PR #185, opened 2017-01-13
Fix launching apps after updating

joaquimrocha:
When an app has an update and the new updated app needs a different
runtime, unless the AppStream of the repo has been very recently
updated, the app not realize that it needs to eventually install a
new runtime.

These changes get the runtime information from the remote ref (meaning
the real runtime for the app) before deciding whether the runtime
needs to be installed which prevents the problem above.

https://phabricator.endlessm.com/T15062

### DIFF
--- a/src/plugins/gs-flatpak.c
+++ b/src/plugins/gs-flatpak.c
@@ -1597,6 +1597,29 @@ gs_plugin_refine_item_state (GsFlatpak *self,
 	return TRUE;
 }
 
+static GsApp *
+gs_flatpak_create_runtime_from_metadata (GsFlatpak *self,
+					 const GsApp *app,
+					 const gchar *data,
+					 const gsize length,
+					 GError **error)
+{
+	g_autofree gchar *runtime = NULL;
+	g_autoptr(GKeyFile) kf = NULL;
+
+	kf = g_key_file_new ();
+	if (!g_key_file_load_from_data (kf, data, length, G_KEY_FILE_NONE, error)) {
+		gs_utils_error_convert_gio (error);
+		return FALSE;
+	}
+	runtime = g_key_file_get_string (kf, "Application", "runtime", error);
+	if (runtime == NULL) {
+		gs_utils_error_convert_gio (error);
+		return FALSE;
+	}
+	return gs_appstream_create_runtime (self->plugin, app, runtime);
+}
+
 static gboolean
 gs_flatpak_set_app_metadata (GsFlatpak *self,
 			     GsApp *app,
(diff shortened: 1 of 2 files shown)


src/plugins/gs-flatpak.c
```diff
@@ -1669,6 +1692,42 @@ gs_flatpak_set_app_metadata (GsFlatpak *self,
 	return TRUE;
 }
 
+static GBytes *
+fetch_remote_metadata (GsFlatpak *self,
+		       GsApp *app,
+		       GCancellable *cancellable,
+		       GError **error)
+{
+	gchar *str;
```
hughsie: unused?

src/plugins/gs-flatpak.c
```diff
@@ -2143,7 +2178,33 @@ install_runtime_for_app (GsFlatpak *self,
 			 GError **error)
 {
 	GsApp *runtime;
-	runtime = gs_app_get_runtime (app);
+	gsize len;
+	const gchar *str;
+	GError *local_error = NULL;
```
hughsie: error_local please

src/plugins/gs-flatpak.c
```diff
@@ -2143,7 +2178,33 @@ install_runtime_for_app (GsFlatpak *self,
...
+	g_autoptr(GBytes) data = fetch_remote_metadata (self, app, cancellable,
+							&local_error);
+
+	if (!data) {
```
hughsie: == NULL pls, it's not a boolean

src/plugins/gs-flatpak.c
```diff
@@ -2143,7 +2178,33 @@ install_runtime_for_app (GsFlatpak *self,
...
+
+	str = g_bytes_get_data (data, &len);
+	runtime = create_runtime_from_flatpak_metadata (self, app, str, len,
+							error);
+
+	if (runtime != NULL) {
+		gs_app_set_runtime (app, runtime);
+	} else {
+		g_warning ("Cannot create runtime from metadata for app %s: %s",
+			   gs_app_get_unique_id (app), local_error->message);
```
hughsie: local_error? How would that be set?

src/plugins/gs-flatpak.c
```diff
@@ -1597,6 +1597,29 @@ gs_plugin_refine_item_state (GsFlatpak *self,
...
+	g_autoptr(GKeyFile) kf = NULL;
+
+	kf = g_key_file_new ();
+	if (!g_key_file_load_from_data (kf, data, length, G_KEY_FILE_NONE, error)) {
+		gs_plugin_flatpak_error_convert (error);
+		return FALSE;
+	}
+	runtime = g_key_file_get_string (kf, "Application", "runtime", error);
+	if (runtime == NULL) {
+		gs_plugin_flatpak_error_convert (error);
```
hughsie: gs_utils_error_convert_gio surely?

src/plugins/gs-flatpak.c
```diff
@@ -1597,6 +1597,29 @@ gs_plugin_refine_item_state (GsFlatpak *self,
...
+				      const gchar *data,
+				      const gsize length,
+				      GError **error)
+{
+	g_autofree gchar *runtime = NULL;
+	g_autoptr(GKeyFile) kf = NULL;
+
+	kf = g_key_file_new ();
+	if (!g_key_file_load_from_data (kf, data, length, G_KEY_FILE_NONE, error)) {
+		gs_plugin_flatpak_error_convert (error);
```
hughsie: gs_utils_error_convert_gio() surely?

src/plugins/gs-flatpak.c
```diff
@@ -2143,7 +2178,33 @@ install_runtime_for_app (GsFlatpak *self,
...
 	GsApp *runtime;
-	runtime = gs_app_get_runtime (app);
+	gsize len;
+	const gchar *str;
+	GError *local_error = NULL;
+	g_autoptr(GBytes) data = fetch_remote_metadata (self, app, cancellable,
+							&local_error);
+
+	if (!data) {
+		g_warning ("Cannot get remote metadata for app %s: %s",
```
hughsie: why printing here? Surely just return FALSE and print the error in the caller?

PR #188, opened 2017-01-13
Fix launching apps after updating

joaquimrocha:
When an app has an update and the new updated app needs a different
runtime, unless the AppStream of the repo has been very recently
updated, the app not realize that it needs to eventually install a
new runtime.

These changes get the runtime information from the remote ref (meaning
the real runtime for the app) before deciding whether the runtime
needs to be installed which prevents the problem above.

https://phabricator.endlessm.com/T15062


joaquimrocha:
Merged it myself because @hughsie reviewed it. Many thanks Richard!

PR #191, opened 2017-01-18
eos: Remove shortcuts directly when applications are removed

joaquimrocha:
This is done automatically by the Shell for all applications except
Chrome as it uses a special desktop file instead of the one provided
by Flatpak. Nonetheless, instead of using a different logic only for
Chrome, this patch always attempts to remove the shortcut for any
Flatpak app that is removed (as it's future proof in case more apps
have special desktop files in the future).

https://phabricator.endlessm.com/T15130


rshuler:
Merging per the CR pass from @mariospr

PR #193, opened 2017-01-25
T14960

GeorgesStavracas:
This PR makes GNOME Software usable on low resolution displays through the following changes:
 * Make the main stack non-homogeneous
 * Make the header buttons non-homogeneous on low resolution displays
 * Allow the sidebar to reduce its width when needed

https://phabricator.endlessm.com/T14960

### DIFF
--- a/src/gnome-software.ui
+++ b/src/gnome-software.ui
@@ -452,6 +452,7 @@
                 <child>
                   <object class="GtkStack" id="stack_main">
                     <property name="visible">True</property>
+                    <property name="hhomogeneous">False</property>
                     <child>
                       <object class="GsShellOverview" id="shell_overview">
                         <property name="visible">True</property>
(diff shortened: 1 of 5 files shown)


src/gs-utils.h
```diff
@@ -78,6 +78,7 @@ gboolean	 gs_utils_error_convert_gdbus	(GError		**perror);
 gboolean	 gs_utils_error_convert_gdk_pixbuf(GError	**perror);
 gboolean	 gs_utils_error_convert_json_glib (GError	**perror);
 gboolean	 gs_utils_error_convert_appstream (GError	**perror);
+gboolean	 gs_utils_is_low_resolution	(void);
```
joaquimrocha: You should align the opening parenthesis with the ones above it.

src/gnome-software.ui
```diff
@@ -452,6 +452,7 @@
                 <child>
```
joaquimrocha: Commenting here because I cannot comment on the git message.
There's a small typo in _appearently_.

PR #196, opened 2017-01-30
Fix leaks when using flatpak_remote_get_url

joaquimrocha:
This function transfers the ownership of the memory to the caller
code which was not freeing it.

https://phabricator.endlessm.com/T15300


pwithnall:
Reworked to apply to the latest copy of `master`. There was only one fix left.

pwithnall:
Also squashed into the corresponding upstream MR (not yet merged upstream): https://gitlab.gnome.org/pwithnall/gnome-software/commits/usb-app-installation

PR #198, opened 2017-02-08
Blacklist GNOME Games: flatpak from stable channel is buggy and crashes

mariospr:
The org.gnome.Games flatpak in the stable channel is buggy and easily
crashes when retrieving covers, as well as when trying to run some
ROMS that rely on libretro-specific plugins, as those are not installed
in the right path under /app.

This is currently fixed in the nightly channel, so we will hopefully be
able to un-blacklist once those versions make it to the stable channel.

https://phabricator.endlessm.com/T15446


mariospr:
This has been fixed in the flatpak published in the stable repo, so there's no need to black list it anymore.

PR #201, opened 2017-02-17
Don't blacklist GNOME Builder

GeorgesStavracas:
In the past, GNOME Builder didn't properly installed
the appstream data, and GNOME Software wouldn't show
it. Sometime later, we blacklisted Builder and, even
after the appstream issue was fixed, Software wouldn't
show it.

Fix that by not blacklisting GNOME Builder.

https://phabricator.endlessm.com/T14858


cosimoc:
@mariospr You're correct; we are still shipping Builder from our own repository, and we should fix whatever problem that bundle has with e.g. appdata that prevents it from showing up in the store.

PR #204, opened 2017-03-08
eos: Make sure all added shortcuts are in the top level

joaquimrocha:
When adding a shortcut, instead of checking if it is already in the grid
(which means it can be hidden inside a folder), we verify whether it is
in the top level of it and add it if it is not. This ensures the icons
are always added in a visible way from GNOME Software, even if they're
configured to be inside a certain folder.

https://phabricator.endlessm.com/T15941

### DIFF
--- a/src/plugins/gs-plugin-eos.c
+++ b/src/plugins/gs-plugin-eos.c
@@ -809,7 +809,6 @@ add_app_to_shell (GsPlugin	*plugin,
 		  GCancellable	*cancellable,
 		  GError	**error_out)
 {
-	GHashTable *apps_with_shortcuts;
 	GError *error = NULL;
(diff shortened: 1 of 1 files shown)


src/plugins/gs-plugin-eos.c
```diff
@@ -825,19 +825,12 @@ add_app_to_shell (GsPlugin	*plugin,
 	}
 
 	app_id = g_app_info_get_id (G_APP_INFO (app_info));
-	apps_with_shortcuts = get_applications_with_shortcuts (plugin, NULL, NULL);
```
cosimoc: `apps_with_shortcuts` will be unused after this commit

joaquimrocha: You're right. Sorry I didn't notice it.

PR #207, opened 2017-03-27
T16302

joaquimrocha:
https://phabricator.endlessm.com/T16302

### DIFF
--- a/src/gs-shell-overview.c
+++ b/src/gs-shell-overview.c
@@ -142,6 +142,17 @@ sort_with_popular_background (GsApp *a, GsApp *b, gpointer user_data)
 	return cmp;
 }
 
+static gint
+sort_by_alphabetical_order (GsApp *a, GsApp *b, gpointer user_data)
+{
+	gint cmp = sort_with_popular_background (a, b, user_data);
+
+	if (cmp == 0)
+		return g_strcmp0 (gs_app_get_name (a), gs_app_get_name (b));
+
+	return cmp;
+}
+
 static void
 gs_shell_overview_decrement_action_cnt (GsShellOverview *self)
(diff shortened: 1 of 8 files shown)


src/plugins/gs-plugin-eos.c
```diff
@@ -1019,3 +1019,43 @@ gs_plugin_add_updates (GsPlugin *plugin,
...
+
+static gboolean
+ensure_only_fnde_popular_apps (GsPlugin *plugin,
+			       GsAppList *list,
+			       GCancellable *cancellable,
+			       GError **error)
+{
+	guint i;
+	GsApp *app;
+	const char *prefixes[] = {"com.endlessm.fnde_", "com.endlessm.teachers_",
```
rshuler: I think it would be better to be more specific and use "com.endlessm.teachers_portal_".

src/gs-shell-overview.c
```diff
@@ -194,9 +206,15 @@ gs_shell_overview_get_popular_cb (GObject *source_object,
 	}
 	gs_app_list_randomize (list);
 
-	/* Sort the randomized list based on the presence of the popular
+	sort_alphabetically = g_settings_get_boolean (priv->settings,
+						      "sort-overview-alphabetically");
+	/* In any case always sort the randomized list based on the presence of the popular
 	 * background so it gives precedence to apps that have that asset */
-	gs_app_list_sort (list, sort_with_popular_background, NULL);
+	if (!sort_alphabetically) {
```
rshuler: Normally, I would probably invert the polarity of the check to check for `if (sort_alphabetically)`, but not a big deal, so I'm going to merge as is.

PR #208, opened 2017-03-27
Retry installing/updating a flatpak after 3 seconds on error

mariospr:
This is an ugly hack to try to overcome network issues that sometimes
lead to hitting errors when installing/updating apps that would lead
to having the process failing, when just retrying would likely work.

It is of course not a real fix and it's probably only useful with the
G_TLS_ERROR_NO_TLS that sometimes glib-networking spits out for no
reason, but we want to try it to see if at least reduces failures.

https://phabricator.endlessm.com/T16317

### DIFF
--- a/src/plugins/gs-flatpak.c
+++ b/src/plugins/gs-flatpak.c
@@ -36,6 +36,8 @@
 #include "gs-flatpak.h"
 #include "gs-flatpak-symlinks.h"
 
+#define RETRY_DELAY_SECONDS 3
+
 struct _GsFlatpak {
 	GObject			 parent_instance;
 	FlatpakInstallation	*installation;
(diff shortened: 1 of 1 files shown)


src/plugins/gs-flatpak.c
```diff
@@ -2329,11 +2349,26 @@ gs_flatpak_app_install (GsFlatpak *self,
...
+								    gs_app_get_local_file (app),
+								    gs_flatpak_progress_cb,
+								    app,
+								    cancellable, error);
+			if (xref != NULL || !keep_trying)
+				break;
+
+			keep_trying = FALSE;
+
+			if (xref == NULL) {
```
joaquimrocha: If _xref_ were not NULL then it would never reach this block, right? So we can drop the check.

I'd probably just use this block with _&& keep_trying_, and set _keep_trying_ inside of it (makes the code more clear IMO).

mariospr: > If _xref_ were not NULL then it would never reach this block, right? So we can drop the check.

Oh, right... you're correct that this check is superfluous

> I'd probably just use this block with _&& keep_trying_, and set _keep_trying_ inside of it (makes the code more clear IMO).

Maybe, but once I remove the extra check I think I actually like it more having a clear early break on `xref != NULL || !keep_trying`, so I'll leave that part that way if you don't have a strong opinion on it. Thanks

PR #210, opened 2017-04-01
Make loading tiles less proeminent

GeorgesStavracas:
The loading tiles are too dark and create too much
contrast with the new light theme.

Fix that by using subtler colors, per proposed mockups.

https://phabricator.endlessm.com/T16197


mariospr:
I find it a bit confusing that you set the temporary class `loading` permanently in the UI file and then remove it dynamically after loading is done. Isn't there a better way to achieve the same result, without setting the `loading` class there (e.g. setting it dynamically only while it's loading)?

PR #217, opened 2017-05-02
Retry update without static deltas

joaquimrocha:
When a file is missing from a commit, updates to the respective app
will fail if using static deltas.
This patch looks for such errors and retries the update without static
deltas.

https://phabricator.endlessm.com/T14158

### DIFF
--- a/src/plugins/gs-flatpak.c
+++ b/src/plugins/gs-flatpak.c
@@ -2418,6 +2418,26 @@ gs_flatpak_update_app (GsFlatpak *self,
 					    gs_app_get_flatpak_branch (app),
 					    gs_flatpak_progress_cb, app,
 					    cancellable, &local_error);
+	/* workaround for when file objects are missing and a static delta
+	 * cannot be used */
+	if (xref == NULL && g_error_matches (local_error, G_IO_ERROR,
+					     G_IO_ERROR_NOT_FOUND)) {
+		g_debug ("Error due to missing file when updating '%s': %s. "
+			 "Retrying the update without static deltas...",
+			 gs_app_get_unique_id (app),
+			 local_error->message);
+		g_clear_error (&local_error);
+
+		/* pull without static deltas */
+		xref = flatpak_installation_update (self->installation,
+						    FLATPAK_UPDATE_FLAGS_NO_STATIC_DELTAS,
+						    gs_app_get_flatpak_kind (app),
+						    gs_app_get_flatpak_name (app),
+						    gs_app_get_flatpak_arch (app),
+						    gs_app_get_flatpak_branch (app),
+						    gs_flatpak_progress_cb, app,
+						    cancellable, &local_error);
+	}
 	if (xref == NULL) {
 		gboolean already_installed =
 			g_error_matches (local_error,


src/plugins/gs-flatpak.c
```diff
@@ -2418,6 +2431,25 @@ gs_flatpak_update_app (GsFlatpak *self,
 					    gs_app_get_flatpak_branch (app),
 					    gs_flatpak_progress_cb, app,
 					    cancellable, &local_error);
+	/* workaround for when file objects are missing and a static delta
+	 * cannot be used */
+	if (xref == NULL && error_is_ostree_missing_file_obj (local_error)) {
+		g_warning ("Error due to missing file when updating '%s': %s. "
```
cosimoc: Not sure this should be a warning...

src/plugins/gs-flatpak.c
```diff
@@ -2393,6 +2393,19 @@ gs_flatpak_app_install (GsFlatpak *self,
...
+error_is_ostree_missing_file_obj (GError *error)
+{
+	if (!g_error_matches (error, G_IO_ERROR,
+			      G_IO_ERROR_NOT_FOUND))
+		return FALSE;
+
+	/* XXX: matching the error messages is not ideal but we need to further
+	 * identify the error we are interested in */
+	return (g_strstr_len (error->message, -1, "No such file object") != NULL) ||
+		(g_strstr_len (error->message, -1, "Couldn't find file object") != NULL);
```
cosimoc: Ugh, I understand why you're doing this (flatpak could return the same error code) but it's really ugly! I am amendable to accepting this for 3.1, but we should have a better plan going forward. Ideally returning a different error from ostree when this is the case.
Can you talk with upstream and open an issue about it?

cosimoc:
The change that I'd like to see before merging this is the demotion of the warning to something less spammy; also, before we accept this, an upstream issue should be opened to avoid looking at the error code.

src/plugins/gs-flatpak.c
cosimoc: Copying @cgwalters here in case he  has an opinion.

src/plugins/gs-flatpak.c
joaquimrocha: @cosimoc My reasoning here was that this should not be reached normally. If it is, we have this workaround in place but it'd still be nice for it to show up in the journal, so we get a notion of whether this is frequently applied when users update. And knowing that is useful after we have a way to prevent the issue IMO.
If you still want to downgrade it to a debug message I understand, just let me know.

src/plugins/gs-flatpak.c
joaquimrocha: @cosimoc yeah, I had mentioned that it was an ugly thing to do. If you prefer we can always allow the retrial when the error simply matches, potentially ending up doing a retrial when the cause is not exactly a missing file object, but the downside of doing a retry when the problem may not be solved is not a big issue...

cgwalters:
I remember we discussed this but I forget the details - under what scenario could this happen?  Is this content missing on the local machine somehow?  Or a partial mirror?

cosimoc:
@cgwalters Yeah, this is probably a consequence of https://github.com/ostreedev/ostree/pull/813. We end up with missing objects when that happens.

PR #219, opened 2017-05-04
Do not try to fetch the default branch for every repo

joaquimrocha:
In order to ensure we have the default branches for the eos-apps repo we
are updating the metadata of every repo that doesn't have that info set
yet.
Since updating the metadata takes a couple of seconds and there are
repos that do not have a default branch set, this operation is taking
several seconds, delaying the launch of GNOME Software.

These changes restrict the check for the default branch only for the
eos-apps repo (as this was the motivation for the logic initially in any
case).

https://phabricator.endlessm.com/T16924

### DIFF
--- a/src/plugins/gs-flatpak.c
+++ b/src/plugins/gs-flatpak.c
@@ -214,7 +214,7 @@ app_is_blacklisted_gnome_flatpak (AsApp *app, AsAppScope scope, FlatpakRemote *x
 	};
 
 	g_autoptr(SoupURI) soup_uri = NULL;
-	const char *remote_url = NULL;
+	g_autofree char *remote_url = NULL;
 	const char *remote_host = NULL;
 	const char *remote_path = NULL;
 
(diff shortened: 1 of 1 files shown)


src/plugins/gs-flatpak.c
```diff
@@ -244,6 +244,31 @@ app_is_blacklisted_gnome_flatpak (AsApp *app, AsAppScope scope, FlatpakRemote *x
...
+{
+	g_autoptr(SoupURI) soup_uri = NULL;
+	const char *remote_host = NULL;
+	const char *remote_path = NULL;
+	const char *remote_url = NULL;
+
+	if (g_strcmp0 (flatpak_remote_get_name (remote), "eos-apps") != 0)
+		return FALSE;
+
+	remote_url = flatpak_remote_get_url (remote);
```
cosimoc: This one is `transfer full` so you  should `g_autofree` it

PR #223, opened 2017-05-16
Assign the branch to Flatpak apps coming from installed desktop files

joaquimrocha:
When an app's non-default branch is installed and the default one isn't,
then the latter will show up as installed because the apps are getting
mixed up.
This is due to merging apps that are created by parsing the installed
desktop files without setting their branch.

These changes set the branch for the mentioned apps which not only fixes
the state for the default branch app in the example mentioned above, but
also show the non-default branch app (as it should be because it is
installed).

https://phabricator.endlessm.com/T17055

### DIFF
--- a/plugins/flatpak/gs-flatpak.c
+++ b/plugins/flatpak/gs-flatpak.c
@@ -30,6 +30,7 @@
 #include <config.h>
 
 #include <flatpak.h>
+#include <string.h>
 
 #include "gs-appstream.h"
 #include "gs-flatpak.h"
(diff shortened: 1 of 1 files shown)


plugins/flatpak/gs-flatpak.c
```diff
@@ -338,6 +339,22 @@ gs_flatpak_add_apps_from_xremote (GsFlatpak *self,
...
+gs_flatpak_discard_desktop_suffix (const gchar *app_id)
+{
+	const guint app_id_len = strlen (app_id);
+	const gchar *desktop_suffix = ".desktop";
+	const guint app_suffix_len = strlen (desktop_suffix);
+	guint prefix_len = app_id_len - app_suffix_len;
+
+	/* if .desktop is not a suffix of the app_id, we return the whole app id */
+	if (app_id_len <= app_suffix_len ||
+	    (g_strcmp0 ((app_id + prefix_len), desktop_suffix) != 0))
```
cosimoc: Why not just use `g_str_has_suffix()`?

PR #225, opened 2017-05-24
T16949

joaquimrocha:
https://phabricator.endlessm.com/T16949

### DIFF
--- a/plugins/flatpak/gs-flatpak.c
+++ b/plugins/flatpak/gs-flatpak.c
@@ -394,6 +394,60 @@ remote_is_eos_apps (FlatpakRemote *remote)
 }
 
 static gboolean
+gs_flatpak_mark_apps_from_usb_remote (GsFlatpak *self,
+				      FlatpakRemote *remote,
+				      gboolean *found_usb_apps,
+				      GCancellable *cancellable,
+				      GError **error)
+{
+	g_autoptr(GPtrArray) refs = NULL;
+	g_autofree gchar *remote_url = flatpak_remote_get_url (remote);
+
+	refs = flatpak_installation_list_remote_refs_sync (self->installation,
+							   remote_url,
+							   cancellable,
+							   error);
+
+	if (refs == NULL)
+		return FALSE;
+
+	for (guint i = 0; i < refs->len; ++i) {
+		FlatpakRef *ref = FLATPAK_REF (g_ptr_array_index (refs, i));
+		g_autoptr(GsApp) app = gs_flatpak_create_app (self, ref);
+		const gchar *ref_collection_id = flatpak_ref_get_collection_id (ref);
+		const gchar *app_collection_id = NULL;
+		AsApp *as_app = as_store_get_app_by_unique_id (self->store,
+							       gs_app_get_unique_id (app),
+							       AS_STORE_SEARCH_FLAG_USE_WILDCARDS);
+
+		if (as_app == NULL)
+			continue;
+
+		app_collection_id = as_app_get_metadata_item (as_app, "flatpak::CollectionId");
+		if (g_strcmp0 (ref_collection_id, app_collection_id) != 0) {
+			g_debug ("Not marking app %s as coming from USB at %s, "
+				 "since the app and ref collection IDs don't "
+				 "match (%s != %s)",
+				 as_app_get_unique_id (as_app), remote_url,
+				 app_collection_id, ref_collection_id);
+			continue;
+		}
+		as_app_add_category (as_app, "USB");
+
+		/* add a keyword so users can search for "usb" */
+		as_app_add_keyword (as_app, NULL, "usb");
+
+		/* adding the priority so these apps are not filtered out in the
+		 * category view when there are others with the same ID */
+		as_app_set_priority (as_app, 100);
+
+		if (found_usb_apps)
+			*found_usb_apps = TRUE;
+	}
+	return TRUE;
+}
+
+static gboolean
 gs_flatpak_add_apps_from_xremote (GsFlatpak *self,
 				  FlatpakRemote *xremote,
 				  GCancellable *cancellable,
@@ -411,6 +465,8 @@ gs_flatpak_add_apps_from_xremote (GsFlatpak *self,
 	g_autoptr(GFile) file = NULL;
 	g_autoptr(GSettings) settings = NULL;
 	g_autoptr(GPtrArray) app_filtered = NULL;
+	g_autofree gchar *remote_url = NULL;
+	g_autofree gchar *collection_id = NULL;
 
 	/* profile */
 	ptask = as_profile_start (gs_plugin_get_profile (self->plugin),
@@ -418,6 +474,17 @@ gs_flatpak_add_apps_from_xremote (GsFlatpak *self,
 				  gs_flatpak_get_id (self),
 				  flatpak_remote_get_name (xremote));
 	g_assert (ptask != NULL);
+
+	remote_url = flatpak_remote_get_url (xremote);
+	if (g_str_has_prefix (remote_url, "file:")) {
+		g_autoptr(GFile) remote_local_repo = g_file_new_for_uri (remote_url);
+		if (!g_file_query_exists (remote_local_repo, cancellable)) {
+			g_set_error (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND,
+				     "Local repo for remote %s not found: %s",
+				     flatpak_remote_get_name (xremote), remote_url);
+			return FALSE;
+		}
+	}
 
 	/* get the AppStream data location */
 	appstream_dir = flatpak_remote_get_appstream_dir (xremote, NULL);
@@ -487,6 +554,8 @@ gs_flatpak_add_apps_from_xremote (GsFlatpak *self,
 		default_branch = g_strdup ("eos3");
 	}
 
+	collection_id = flatpak_remote_get_collection_id (xremote);
+
 	/* get all the apps and fix them up */
 	app_filtered = g_ptr_array_new ();
 	for (i = 0; i < apps->len; i++) {
@@ -514,6 +583,11 @@ gs_flatpak_add_apps_from_xremote (GsFlatpak *self,
 		as_app_set_scope (app, self->scope);
 		as_app_set_origin (app, flatpak_remote_get_name (xremote));
 		as_app_add_keyword (app, NULL, "flatpak");
+
+		if (collection_id != NULL)
+			as_app_add_metadata (app, "flatpak::CollectionId",
+					     collection_id);
+
 		g_debug ("adding %s", as_app_get_unique_id (app));
 		g_ptr_array_add (app_filtered, app);
 	}
@@ -636,6 +710,8 @@ gs_flatpak_rescan_appstream_store (GsFlatpak *self,
 	guint i;
 	g_autoptr(AsProfileTask) ptask = NULL;
 	g_autoptr(GPtrArray) xremotes = NULL;
+	g_autoptr(GPtrArray) usb_remotes = NULL;
+	const FlatpakRemoteType types[] = { FLATPAK_REMOTE_TYPE_STATIC, FLATPAK_REMOTE_TYPE_USB };
 
 	/* profile */
 	ptask = as_profile_start (gs_plugin_get_profile (self->plugin),
@@ -647,16 +723,26 @@ gs_flatpak_rescan_appstream_store (GsFlatpak *self,
 	as_store_remove_all (self->store);
 
 	/* go through each remote adding metadata */
-	xremotes = flatpak_installation_list_remotes (self->installation,
-						      cancellable,
-						      error);
+	xremotes = flatpak_installation_list_remotes_by_type (self->installation,
+							      types,
+							      G_N_ELEMENTS (types),
+							      cancellable, error);
 	if (xremotes == NULL) {
 		gs_flatpak_error_convert (error);
 		return FALSE;
 	}
+
+	usb_remotes = g_ptr_array_new ();
+
 	for (i = 0; i < xremotes->len; i++) {
 		g_autoptr(GError) local_error = NULL;
 		FlatpakRemote *xremote = g_ptr_array_index (xremotes, i);
+		/* gather the USB type remotes for marking the apps as coming
+		 * from it later */
+		if (flatpak_remote_get_remote_type (xremote) == FLATPAK_REMOTE_TYPE_USB) {
+			g_ptr_array_add (usb_remotes, xremote);
+			continue;
+		}
 		if (flatpak_remote_get_disabled (xremote))
 			continue;
 		g_debug ("found remote %s",
@@ -664,6 +750,17 @@ gs_flatpak_rescan_appstream_store (GsFlatpak *self,
 		if (!gs_flatpak_add_apps_from_xremote (self, xremote, cancellable,
 						       &local_error)) {
 			g_debug ("Failed to add apps from remote: %s; skipping",
+				 local_error->message);
+		}
+	}
+
+	for (i = 0; i < usb_remotes->len; ++i) {
+		FlatpakRemote *remote = g_ptr_array_index (usb_remotes, i);
+		g_autoptr(GError) local_error = NULL;
+		if (!gs_flatpak_mark_apps_from_usb_remote (self, remote, NULL,
+							   cancellable, &local_error)) {
+			g_debug ("Failed to mark apps coming from USB remote at %s: %s",
+				 flatpak_remote_get_url (remote),
 				 local_error->message);
 		}
 	}
--- a/plugins/flatpak/gs-flatpak.c
+++ b/plugins/flatpak/gs-flatpak.c
@@ -41,12 +41,15 @@ struct _GsFlatpak {
 	GsFlatpakFlags		 flags;
 	FlatpakInstallation	*installation;
 	GHashTable		*broken_remotes;
+	GHashTable		*usb_remotes;
 	GFileMonitor		*monitor;
 	AsAppScope		 scope;
 	GsPlugin		*plugin;
 	AsStore			*store;
 	gchar			*id;
 	guint			 changed_id;
+	GVolumeMonitor		*volume_monitor;
+	guint			 mount_removed_handler_id;
 };
 
 G_DEFINE_TYPE (GsFlatpak, gs_flatpak, G_TYPE_OBJECT)
@@ -413,6 +416,8 @@ gs_flatpak_mark_apps_from_usb_remote (GsFlatpak *self,
 
 	for (guint i = 0; i < refs->len; ++i) {
 		FlatpakRef *ref = FLATPAK_REF (g_ptr_array_index (refs, i));
+		GPtrArray *usb_apps_for_remote = g_hash_table_lookup (self->usb_remotes,
+								      remote_url);
 		g_autoptr(GsApp) app = gs_flatpak_create_app (self, ref);
 		const gchar *ref_collection_id = flatpak_ref_get_collection_id (ref);
 		const gchar *app_collection_id = NULL;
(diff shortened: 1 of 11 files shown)


plugins/flatpak/gs-flatpak.c
```diff
@@ -214,6 +215,46 @@ remote_is_eos_apps (FlatpakRemote *remote)
 }
 
 static gboolean
+gs_flatpak_is_removable_remote (GsFlatpak *self,
```
cosimoc: I believe that you could also implement this by using `g_file_find_enclosing_mount()` on the remote URL and avoid enumerating all the drives from the volume monitor.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3468,6 +3537,7 @@ gs_flatpak_init (GsFlatpak *self)
 	self->broken_remotes = g_hash_table_new_full (g_str_hash, g_str_equal,
 						      g_free, NULL);
 	self->store = as_store_new ();
+	self->volume_monitor = g_volume_monitor_get ();
```
cosimoc: Not sure if you will end up needing this, but if so it's leaked.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3144,9 +3215,27 @@ gs_flatpak_file_to_app_repo (GsFlatpak *self,
...
+	file_type = g_file_query_file_type (file, G_FILE_QUERY_INFO_NONE, cancellable);
+	if (file_type == G_FILE_TYPE_DIRECTORY) {
+		g_autoptr(GError) local_error = NULL;
+		app = gs_flatpak_create_app_from_repo_dir (self, file, cancellable,
+							   &local_error);
+		if (app == NULL) {
+			g_autofree gchar *path = g_file_get_path (file);
+			g_debug ("Failed to get app from dir %s: %s; skipping", path,
+				 local_error->message);
+			return TRUE;
```
cosimoc: Pretty sure you should return `FALSE` and set the error here.

src/gs-shell.c
```diff
@@ -1931,14 +1974,80 @@ gs_shell_show_search (GsShell *shell, const gchar *search)
...
+
+	/* print what we've got */
+	tmp = gs_app_to_string (app);
+	g_debug ("App from file:\n%s", tmp);
+
+	removable_media_cat = gs_app_get_metadata_item (app, "EndlessOS::RemovableMediaCategory");
+	if (removable_media_cat != NULL) {
+		GsCategory *category;
+		if (gs_app_get_metadata_item (app, "EndlessOS::ReloadOverview") == NULL) {
+			reload_overview_and_select_category (shell, removable_media_cat);
```
cosimoc: Why do you need to reload the overview in this case? The metadata that calls for that was not set; perhaps you can just select the category.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2761,6 +2761,44 @@ gs_flatpak_create_app_from_repo_dir (GsFlatpak *self,
 	return app;
 }
 
+static void
+gs_flatpak_mount_removed (GVolumeMonitor *volume_monitor,
+			  GDrive *mount,
```
cosimoc: This should be a `GMount *`

plugins/flatpak/gs-flatpak.c
joaquimrocha: Thanks. It seemed like a lot of work for what I needed but I failed to find a better solution. I will give your suggestion a try.

plugins/flatpak/gs-flatpak.c
joaquimrocha: Not a critical leak but sorry about that. Good catch!

plugins/flatpak/gs-flatpak.c
joaquimrocha: I added this on purpose because IIRC this would e.g. fail in the system instance of gs-flatpak and prevent the further processing by the user's gs-flatpak instance.
Maybe something more clever could be done but I was in a rush and decided to let it silently fail (the reasoning being: in comparison with the regular file-to-app logic, having an //.ostree// dir in a USB drive is a less strict action than having a //.flatpakref// file, so silently failing it not as critical).

cosimoc: Sounds good if the error ends up blocking further processing.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3178,6 +3184,66 @@ gs_flatpak_create_runtime_repo (GsFlatpak *self,
...
+
+	uri = g_file_get_uri (file);
+
+	for (guint i = 0; i < remotes->len; ++i) {
+		FlatpakRemote *remote = g_ptr_array_index (remotes, i);
+		g_autofree gchar *url = flatpak_remote_get_url (remote);
+		g_autoptr(GError) local_error = NULL;
+
+		/* skip if the remote does not have the dir's path as a prefix */
+		if (!g_str_has_prefix (url, uri))
```
pwithnall: The use of `uri` and `url` as separate variables is a bit confusing. How about renaming `uri` to `file_uri`, and `url` to `remote_uri`?

Also, be wary of string comparisons on URIs, since you can get false positives if the prefix doesn’t end in a slash. Use `g_file_has_parent()` instead?

plugins/flatpak/gs-flatpak.c
```diff
@@ -3178,6 +3184,66 @@ gs_flatpak_create_runtime_repo (GsFlatpak *self,
...
+		if (!gs_flatpak_mark_apps_from_usb_remote (self, remote, &reload_overview,
+							   cancellable, &local_error)) {
+			g_debug ("Failed to mark USB apps from remote at %s: %s",
+				 url, local_error->message);
+			continue;
+		}
+	}
+
+	if (!dir_has_repo) {
+		g_set_error (error, G_IO_ERROR_FAILED, G_IO_ERROR_NOT_SUPPORTED,
```
pwithnall: Would `G_IO_ERROR_NOT_FOUND` be a more appropriate error code?

plugins/flatpak/gs-plugin-flatpak.c
```diff
@@ -424,6 +424,24 @@ gs_plugin_flatpak_file_to_app_repo (GsPlugin *plugin,
 	GsPluginData *priv = gs_plugin_get_data (plugin);
 	g_autoptr(GsApp) app_tmp = NULL;
 	g_autoptr(GsAppList) list_tmp = NULL;
+	GFileType file_type = G_FILE_TYPE_UNKNOWN;
+
+	file_type = g_file_query_file_type (file, G_FILE_QUERY_INFO_NONE, cancellable);
+
+	/* check if this is rather a directory*/
```
pwithnall: Nitpick: Missing space before `*/`.

plugins/flatpak/gs-plugin-flatpak.c
```diff
@@ -424,6 +424,24 @@ gs_plugin_flatpak_file_to_app_repo (GsPlugin *plugin,
...
+			g_autoptr(GError) local_error = NULL;
+			GsFlatpak *flatpak = g_ptr_array_index (priv->flatpaks, i);
+			app_tmp = gs_flatpak_create_app_from_repo_dir (flatpak, file, cancellable,
+								       &local_error);
+			if (app_tmp)
+				break;
+		}
+		if (app_tmp != NULL)
+			gs_app_list_add (list, app_tmp);
+		return TRUE;
```
pwithnall: Is it appropriate to return `TRUE` here if (`app_tmp == NULL`)?

src/gs-overview-page.c
```diff
@@ -1137,6 +1140,13 @@ gs_overview_page_class_init (GsOverviewPageClass *klass)
 			      NULL, NULL, g_cclosure_marshal_VOID__VOID,
 			      G_TYPE_NONE, 0);
 
+	signals [SIGNAL_CATEGORIES_LOADED] =
+		g_signal_new ("categories-loaded",
+			      G_TYPE_FROM_CLASS (object_class), G_SIGNAL_RUN_LAST,
+			      G_STRUCT_OFFSET (GsOverviewPageClass, refreshed),
```
pwithnall: This struct offset doesn’t look valid.

src/gs-shell.c
```diff
@@ -582,6 +588,43 @@ initial_overview_load_done (GsOverviewPage *overview_page, gpointer data)
 	g_signal_emit (shell, signals[SIGNAL_LOADED], 0);
 }
 
+static void
+on_overview_categories_refreshed (GsOverviewPage *page, gpointer user_data)
```
pwithnall: Nitpick: Would it be better to call this `on_overview_categories_loaded()` so it matches the signal name? Especially since `refreshed` is a separate signal.

src/gs-shell.c
```diff
@@ -2008,15 +2051,96 @@ gs_shell_show_search (GsShell *shell, const gchar *search)
...
 gs_shell_show_local_file (GsShell *shell, GFile *file)
 {
 	g_autoptr(GsApp) app = gs_app_new (NULL);
-	save_back_entry (shell);
+	g_autoptr(GsPluginJob) plugin_job = NULL;
+	GsShellPrivate *priv = gs_shell_get_instance_private (shell);
+	GFileType file_type = g_file_query_file_type (file, G_FILE_QUERY_INFO_NONE,
+						      NULL);
+
+	/* check if this is rather a directory*/
```
pwithnall: Nitpick: Missing space before `*/`

plugins/flatpak/gs-flatpak.c
```diff
@@ -394,6 +395,23 @@ remote_is_eos_apps (FlatpakRemote *remote)
...
+	g_autoptr(GFile) mount_root = NULL;
+	g_autofree gchar *mount_uri = NULL;
+
+	if (!mount)
+		return FALSE;
+
+	mount_root = g_mount_get_root (mount);
+	mount_uri = g_file_get_uri (mount_root);
+
+	if (g_str_has_prefix (uri, mount_uri))
```
pwithnall: I think it would be better to operate on `GFile`s and use `g_file_has_parent()` here, otherwise you run into problems if the prefix string doesn’t end with a slash.

plugins/flatpak/gs-flatpak.c
```diff
@@ -454,6 +472,7 @@ gs_flatpak_mark_apps_from_usb_remote (GsFlatpak *self,
 	return TRUE;
 }
 
+
```
pwithnall: Nitpick: Spurious new line?

plugins/flatpak/gs-flatpak.c
pwithnall: You should also disconnect your signal handler when dropping your reference to the `GVolumeMonitor`, in case you don’t hold the last reference to it.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3244,6 +3263,30 @@ gs_flatpak_create_app_from_repo_dir (GsFlatpak *self,
 	return app;
 }
 
+static void
+gs_flatpak_mount_removed (GVolumeMonitor *volume_monitor,
+			  GMount *mount,
```
pwithnall: Nitpick: Indentation problem with arguments.

pwithnall: Although I’ve just realised that all these indentation problems may be caused by github not aligning the tabs the same way as your editor. If so, ignore my comments.

plugins/flatpak/gs-flatpak.c
```diff
@@ -655,13 +655,17 @@ gs_flatpak_rescan_appstream_store (GsFlatpak *self,
...
 		if (flatpak_remote_get_disabled (xremote))
 			continue;
 		g_debug ("found remote %s",
 			 flatpak_remote_get_name (xremote));
-		if (!gs_flatpak_add_apps_from_xremote (self, xremote, cancellable, error))
-			return FALSE;
+		if (!gs_flatpak_add_apps_from_xremote (self, xremote, cancellable,
+						       &local_error)) {
+			g_debug ("Failed to add apps from remote: %s; skipping",
+				 local_error->message);
```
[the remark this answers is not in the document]
joaquimrocha: That's reasonable. Thanks.

plugins/flatpak/gs-flatpak.c
```diff
@@ -393,6 +393,61 @@ remote_is_eos_apps (FlatpakRemote *remote)
 		g_str_has_suffix (remote_host, "ostree.endlessm.com");
 }
 
+static gboolean
+gs_flatpak_mark_apps_from_usb_remote (GsFlatpak *self,
+				      FlatpakRemote *remote,
```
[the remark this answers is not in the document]
joaquimrocha: I think this is Github/Chrome's formatting. g-s uses tabs (-_-), and such display mismatches have happened before (on Emacs it LGTM).

plugins/flatpak/gs-flatpak.c
```diff
@@ -393,6 +393,61 @@ remote_is_eos_apps (FlatpakRemote *remote)
...
+	g_autoptr(GPtrArray) refs = NULL;
+	g_autofree gchar *remote_url = flatpak_remote_get_url (remote);
+
+	refs = flatpak_installation_list_remote_refs_sync (self->installation,
+							   remote_url,
+							   cancellable,
+							   &error_local);
+
+	if (refs == NULL)
+		return FALSE;
```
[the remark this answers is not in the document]
joaquimrocha: Yup! Good catch!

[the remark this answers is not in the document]
joaquimrocha: It should actually just use the error out arg.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3178,6 +3184,66 @@ gs_flatpak_create_runtime_repo (GsFlatpak *self,
 	return g_steal_pointer (&app);
 }
 
+GsApp *
+gs_flatpak_create_app_from_repo_dir (GsFlatpak *self,
+				     GFile *file,
```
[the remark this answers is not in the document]
joaquimrocha: It's again the tabs...

plugins/flatpak/gs-flatpak.c
joaquimrocha: Good catch! Thanks.

plugins/flatpak/gs-flatpak.c
joaquimrocha: It was meant to say that loading a repo from such a dir is not supported (because we don't know the dir's format).
But I don't mind changing it to `NOT_FOUND` since we cannot find a remote in it...

plugins/flatpak/gs-flatpak.c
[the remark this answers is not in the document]
pwithnall: OK. Maybe add a `FIXME` comment about it then.

plugins/flatpak/gs-flatpak.c
```diff
@@ -4059,6 +4125,8 @@ gs_flatpak_init (GsFlatpak *self)
 {
 	self->broken_remotes = g_hash_table_new_full (g_str_hash, g_str_equal,
 						      g_free, NULL);
+	self->loaded_usb_remotes = g_hash_table_new_full (g_str_hash, g_str_equal,
```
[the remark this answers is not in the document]
joaquimrocha: This `loaded_usb_remotes` doesn't even belong to this patch... (it was a complicated change :smiley_cat: )

plugins/flatpak/gs-flatpak.c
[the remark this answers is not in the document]
joaquimrocha: I talked to Alan Day and it's not clear whether we'll touch this issue here, as it's apparently enough to ensure that if there are no apps from any plugin, we get a proper empty page message in g-s. I will keep a note to discuss this in the future with Richard too.

plugins/flatpak/gs-flatpak.c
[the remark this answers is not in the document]
pwithnall: Indeed, sorry for the noise.

plugins/flatpak/gs-flatpak.c
```diff
@@ -419,6 +484,17 @@ gs_flatpak_add_apps_from_xremote (GsFlatpak *self,
 				  flatpak_remote_get_name (xremote));
 	g_assert (ptask != NULL);
 
+	remote_url = flatpak_remote_get_url (xremote);
+	if (g_str_has_prefix (remote_url, "file://")) {
+		remote_local_repo = g_file_new_for_uri (remote_url);
```
pwithnall: Similarly, I’d drop `g_str_has_prefix()` here, hoist the `g_file_new_for_uri()` call, and then check the scheme using `g_file_has_uri_scheme()`.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3075,6 +3185,106 @@ gs_flatpak_create_runtime_repo (GsFlatpak *self,
...
+	FlatpakRemoteType usb_type[] = { FLATPAK_REMOTE_TYPE_USB };
+
+	remotes = flatpak_installation_list_remotes_by_type (self->installation,
+							     usb_type,
+							     G_N_ELEMENTS (usb_type),
+							     cancellable, error);
+	if (remotes == NULL)
+		return NULL;
+
+	file_uri = g_file_get_uri (file);
```
pwithnall: This could be moved into the error path which is the only user of it.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3935,11 +4145,19 @@ gs_flatpak_finalize (GObject *object)
 		self->changed_id = 0;
 	}
 
+	if (self->mount_removed_handler_id > 0) {
```
pwithnall: `mount_removed_handler_id` never seems to be set. Was this supposed to be for the `mount-removed` signal on `self->volume_monitor` (not `self->monitor`)?

plugins/flatpak/gs-flatpak.c
```diff
@@ -3075,6 +3185,106 @@ gs_flatpak_create_runtime_repo (GsFlatpak *self,
...
+				     GFile *file,
+				     GCancellable *cancellable,
+				     GError **error)
+{
+	g_autoptr(GPtrArray) remotes = NULL;
+	g_autofree gchar *file_uri = NULL;
+	gboolean dir_has_repo = FALSE;
+	gboolean reload_overview = FALSE;
+	GsApp *app = NULL;
+	FlatpakRemoteType usb_type[] = { FLATPAK_REMOTE_TYPE_USB };
```
pwithnall: This could be `const`.

plugins/flatpak/gs-flatpak.c
joaquimrocha: Yeah. I was late for picking up my daguther :/

plugins/flatpak/gs-flatpak.c
pwithnall: Are you going to do this, or do you disagree?

plugins/flatpak/gs-flatpak.c
pwithnall: I don’t mind if a PR is delayed by a few hours so you can pick your daughter up and then test the PR confidently before submitting it!

plugins/flatpak/gs-flatpak.c
joaquimrocha: Are there any advantages to doing so?
I'd prefer not to initialize an object just to realize whether it's needed or not (only needed if it's a `file://` scheme). So even though your proposal makes things look a bit more self contained, I don't find it good enough a reason to have the object always allocated (even if the impact is minimal).

I have updated it in the meanwhile to declare the `remote_local_repo` inside the same body where it's used, so it's easier to read and has a more limited scope.

pwithnall: Advantages: you don’t have to worry about the difference between `file:`, `file:/`, `file://` and `file:///`; all the string parsing is contained in one place (GLib) which has already been tested.

joaquimrocha: I believe I have seen flatpak checking for `file://` internally, so I don't think we should be covered with the prefix matching, and I prefer this to initializing an object I may not need. So unless I am failing to see the big advantage I actually prefer to keep it like this for now.

pwithnall: In that case, please change the prefix match to `file:` rather than `file://`, so that it at least matches the rest of the checks in `flatpak-dir.c`.

joaquimrocha: Okay. That makes sense. Thanks.

PR #230, opened 2017-06-20
T17490

smspillaz:
Add a mechanism to filter out non-discovery feed apps from the discovery feed query. Some of this API could be upstreamed.

https://phabricator.endlessm.com/T17490


lib/gs-app-list.h
```diff
@@ -32,6 +32,9 @@ G_BEGIN_DECLS
 
 G_DECLARE_FINAL_TYPE (GsAppList, gs_app_list, GS, APP_LIST, GObject)
 
+typedef gboolean (*GsAppListFilterFunc)		(GsApp		*app,
+						 gpointer	 user_data);
```
erikos: Alignment.

lib/gs-plugin-loader.c
```diff
@@ -1177,6 +1189,33 @@ gs_plugin_loader_get_app_is_compatible (GsApp *app, gpointer user_data)
 	return FALSE;
 }
 
+static gboolean
+gs_plugin_loader_filter_app_list (GsPluginLoaderJob *job,
+				  GsAppList	    *app_list,
```
erikos: Alignment.

lib/gs-plugin-types.h
```diff
@@ -204,6 +204,21 @@ typedef enum {
 } GsPluginRefreshFlags;
 
 /**
+ * GsPluginFilterFlags:
+ * @GS_PLUGIN_FILTER_FLAGS_NONE:		No filtering
+ * @GS_PLUGIN_FILTER_FLAGS_DISCOVERY_FEED:	Only show apps with discovery feed content
+ *
+ * The flags used for search filtering. Plugins might be want to restrict
```
erikos: ...might want to...

lib/gs-plugin-types.h
```diff
@@ -204,6 +204,21 @@ typedef enum {
...
+ * the apps returned in a search query based on the flags provided.
+ *
+ **/
+typedef enum {
+	GS_PLUGIN_FILTER_FLAGS_NONE			= 0,
```
erikos: the 0 looks quite distant.

lib/gs-app-list.h
joaquimrocha: I believe the alignment must be okay outside of GitHub's view, as it also showed like this in the code @smspillaz removed, and shows okay in my terminal.

smspillaz:
> I want to keep our changes to GNOME Software's core to a minimum and I think that this can be accomplished by a simply implementing the gs_plugin_add_search in our EOS plugin and just remove the apps that do not match our criteria from the list of apps arg. Since our plugin runs after all the main ones, this ought to work.

Hmm, I'm not sure about that. Refinement runs after the search hooks have completed, which means that we wouldn't typically have access to the flatpak metadata at that point (https://github.com/endlessm/gnome-software/blob/master/lib/gs-plugin-loader.c#L2001).

I'm also not sure how we could instruct the `search` hook to filter the app list unless we passed some flags to it indicating that we should do that. One method to avoid the API change is to have a "special" token like `has_metadata:Endless::DiscoveryFeed` which our search plugin can detect and act on. But then it still needs to get access to the flatpak metadata which AFAICT is only available through `gs-flatpak`. We *could* set the metadata early, but that feels like a layering violation to me: https://github.com/endlessm/gnome-software/blob/master/plugins/flatpak/gs-flatpak.c#L1429

joaquimrocha:
The metadata is set in the GsApp object and comes originally from the app-data. So all you need to do is to perform the search for Endless' apps as you did but then, you implement _gs_plugin_add_search_ and remove the apps you do not want from the list that's given as an argument. We do many things like this in our _EOS_ plugin in order to avoid complex changes to the core and I think this works. Let's discuss it on Slack.

smspillaz:
New approach in https://github.com/endlessm/gnome-software/pull/231

PR #231, opened 2017-06-22
T17490

smspillaz:
https://phabricator.endlessm.com/T17490

### DIFF
--- a/src/gs-discovery-feed-content-provider.c
+++ b/src/gs-discovery-feed-content-provider.c
@@ -135,6 +135,52 @@ get_app_thumbnail_cached_filename (GsApp *app)
 }
 
 static void
+generate_random_indices_array (guint *array,
+			       guint length,
+			       guint min,
+			       guint max)
+{
+	guint i;
+	for (i = 0; i < length; ++i) {
+		/* Need to loop here in case we pick a dupe */
+		gboolean again = TRUE;
+		while (again) {
+			guint index = g_random_int_range(0, length);
+			guint j = 0;
+
+			/* We might pick something unique this time */
+			again = FALSE;
+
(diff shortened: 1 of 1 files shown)


joaquimrocha:
A couple of things that I think may improve this:
- You add the search terms in this [line](https://github.com/endlessm/gnome-software/blob/d81bb6344f30c87afff5c18513252cb57ab1ada3/src/gs-discovery-feed-content-provider.c#L302) so it should be possible to add a keyword like _eos-discovery-feed_ to the terms, but for it to work you also need to add the keyword to the apps from the plugins side (hopefully in _EOS_'s refine it'll work). Still, doing it all in the discovery feed content provider may be easier and a better idea if we just want this in the shell search's results.

In any case I think that the current approach of randomizing the results if not right. I do understand the reasoning of giving every app a chance of showing up in the results, however I see two problems:
1) if a user searches for an app and finds it, they will assume that the next time they do it, they'll find it again; yet the more apps we have beyond MAX_INSTALLED_APPS, the less chances that will happen;
2) AFAICS you always randomize the results, not only if they are more than the maximum allowed number; so we're creating confusion for no good reason.

I would assume that the results are limited because there's no point in showing 500 apps as users will never go beyond a certain number and what they do is to add more terms to refine the search: 'Game' -> (1000 results?) : 'Game + Tux' -> 5 results = success.
If we disagree here, I think it's maybe worth asking our Design team for the right approach to take.

Anyway if you really need to randomize the results, instead of the approach you took there's a private function _gs_app_list_randomize_ that will shuffle the items inside a _GsAppList_; it's probably worth making it public so we can use one line for randomizing the results.

smspillaz:
Superseded this by #233 and #232

PR #233, opened 2017-06-22
feed: Select up to MAX_INSTALLABLE_APPS at random

smspillaz:
We have to build up an array of unique indices and then pick those
ones out of the list.

https://phabricator.endlessm.com/T21028

### DIFF
--- a/src/gs-discovery-feed-content-provider.c
+++ b/src/gs-discovery-feed-content-provider.c
@@ -64,12 +64,6 @@ typedef struct {
 	guint weight;
 } DiscoveryFeedKudoWeight ;
 
-static gint
-get_day_of_week (void)
-{
-  g_autoptr(GDateTime) datetime = g_date_time_new_now_local ();
-  return g_date_time_get_day_of_week (datetime);
-}
 
 static const DiscoveryFeedKudoWeight discovery_feed_app_kudos[] =
 {
@@ -136,6 +130,38 @@ get_app_thumbnail_cached_filename (GsApp *app)
 	return NULL;
 }
 
+static gboolean
+filter_for_discovery_feed_apps (GsApp *app, gpointer user_data)
+{
+	g_autofree gchar *app_thumbnail_filename = NULL;
+	GdkPixbuf *pixbuf;
+	g_autoptr(GVariant) icon_serialized = NULL;
+
+	if (gs_app_get_state (app) != AS_APP_STATE_AVAILABLE)
+		return FALSE;
+
+	/* This app must have an icon which is serializable */
+	pixbuf = gs_app_get_pixbuf (app);
+	if (pixbuf != NULL)
+		icon_serialized = g_icon_serialize (G_ICON (pixbuf));
+
+	if (!icon_serialized)
+		return FALSE;
+
+	/* Check to make sure that the app has an available thumbnail
+	 * otherwise, we need to skip it completely */
+	app_thumbnail_filename = get_app_thumbnail_cached_filename (app);
+
+	if (!app_thumbnail_filename)
+		return FALSE;
+
+	/* App must also be tagged as having discovery feed content */
+	if (!gs_app_get_metadata_item (app, "Endless::HasDiscoveryFeedContent"))
+		return FALSE;
+
+	return TRUE;
+}
+
 static void
 search_done_cb (GObject *source,
 		GAsyncResult *res,
@@ -146,7 +172,6 @@ search_done_cb (GObject *source,
 	guint i;
 	GVariantBuilder builder;
 	guint app_list_length;
-	guint count = 0;
 	g_autoptr(GsAppList) list = NULL;
 
 	list = gs_plugin_loader_job_process_finish (self->plugin_loader, res, NULL);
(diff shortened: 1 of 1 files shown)


smspillaz:
Just copy-pasting @joaquimrocha 's comment here so that it doesn't get lost as we split the branch:

    In any case I think that the current approach of randomizing the results if not right. I do understand the reasoning of giving every app a chance of showing up in the results, however I see two problems:

    if a user searches for an app and finds it, they will assume that the next time they do it, they'll find it again; yet the more apps we have beyond MAX_INSTALLED_APPS, the less chances that will happen;
AFAICS you always randomize the results, not only if they are more than the maximum allowed number; so we're creating confusion for no good reason.
    I would assume that the results are limited because there's no point in showing 500 apps as users will never go beyond a certain number and what they do is to add more terms to refine the search: 'Game' -> (1000 results?) : 'Game + Tux' -> 5 results = success.
    If we disagree here, I think it's maybe worth asking our Design team for the right approach to take.

    Anyway if you really need to randomize the results, instead of the approach you took there's a private function gs_app_list_randomize that will shuffle the items inside a GsAppList; it's probably worth making it public so we can use one line for randomizing the results.

lib/gs-app-list-private.h
```diff
@@ -65,7 +65,6 @@ GsAppList	*gs_app_list_copy		(GsAppList	*list);
 guint		 gs_app_list_get_size_peak	(GsAppList	*list);
 void		 gs_app_list_filter_duplicates	(GsAppList	*list,
 						 GsAppListFilterFlags flags);
-void		 gs_app_list_randomize		(GsAppList	*list);
```
cosimoc: Not sure you need to move this to the public header actually... `src/gs-discovery-feed-content-provider.c` already includes `lib/gs-app-list-private.h` as far as I can see.

src/gs-discovery-feed-content-provider.c
```diff
@@ -136,6 +130,38 @@ get_app_thumbnail_cached_filename (GsApp *app)
...
+	GdkPixbuf *pixbuf;
+	g_autoptr(GVariant) icon_serialized = NULL;
+
+	if (gs_app_get_state (app) != AS_APP_STATE_AVAILABLE)
+		return FALSE;
+
+	/* This app must have an icon which is serializable */
+	pixbuf = gs_app_get_pixbuf (app);
+	if (pixbuf != NULL)
+		icon_serialized = g_icon_serialize (G_ICON (pixbuf));
```
cosimoc: Serialization can be a pretty expensive operation, and it is very unlikely that it will fail. I would lean towards assuming that if `gs_app_get_pixbuf()` returns non-NULL, it will succeed, and so perhaps only add a warning if serialization fails later when building the variant.

Related: since you search specifically for apps that have an icon (https://github.com/endlessm/gnome-software/blob/e46ff4fb7841008b94cde0394c2b6991a59d944f/src/gs-discovery-feed-content-provider.c#L248), do we need the pixbuf check here?

src/gs-discovery-feed-content-provider.c
```diff
@@ -136,6 +130,38 @@ get_app_thumbnail_cached_filename (GsApp *app)
...
 }
 
+static gboolean
+filter_for_discovery_feed_apps (GsApp *app, gpointer user_data)
+{
+	g_autofree gchar *app_thumbnail_filename = NULL;
+	GdkPixbuf *pixbuf;
+	g_autoptr(GVariant) icon_serialized = NULL;
+
+	if (gs_app_get_state (app) != AS_APP_STATE_AVAILABLE)
```
cosimoc: Would it be possible to express this constraint when doing the search instead of at filtering time?

lib/gs-app-list-private.h
smspillaz: Yeah, I saw that now actually. I'll remove the commit, looks like this is not necessary \o/

src/gs-discovery-feed-content-provider.c
smspillaz: Doesn't look like it - or at least, the shell search provider also needs to make such a check: https://github.com/endlessm/gnome-software/blob/master/src/gs-shell-search-provider.c#L96

src/gs-discovery-feed-content-provider.c
smspillaz: Yeah, I did notice that.

I think as a defensive measure, we probably want to at least ensure that if an app does turn out to have a null pixbuf / is unserializable that we just emit a warning and move on.

src/gs-discovery-feed-content-provider.c
```diff
@@ -136,6 +130,29 @@ get_app_thumbnail_cached_filename (GsApp *app)
 	return NULL;
 }
 
+static gboolean
+filter_for_discovery_feed_apps (GsApp *app, gpointer user_data)
+{
+	g_autofree gchar *app_thumbnail_filename = NULL;
+	g_autoptr(GVariant) icon_serialized = NULL;
```
cosimoc: Remove unused variable

src/gs-discovery-feed-content-provider.c
```diff
@@ -159,54 +175,49 @@ search_done_cb (GObject *source,
...
+	 * on those indices */
 	g_variant_builder_init (&builder, G_VARIANT_TYPE ("aa{sv}"));
-	for (i = 0; i < app_list_length; i++) {
-		g_autofree gchar *app_thumbnail_filename = NULL;
-		GdkPixbuf *pixbuf;
+	for (i = 0; i < MIN (app_list_length, MAX_INSTALLABLE_APPS); ++i) {
+		GsApp *app = GS_APP (gs_app_list_index (list, i));
+		const gchar *app_id = gs_app_get_id (app);
+		g_autofree gchar *app_thumbnail_filename = get_app_thumbnail_cached_filename (app);
+		g_autoptr(GdkPixbuf) pixbuf = gs_app_get_pixbuf (app);
```
cosimoc: `gs_app_get_pixbuf()` is `transfer none`: https://github.com/endlessm/gnome-software/blob/master/lib/gs-app.c#L1634
So this code will remove a reference it does not own.

src/gs-discovery-feed-content-provider.c
smspillaz: Yup, good catch. Fixed.

src/gs-discovery-feed-content-provider.c
smspillaz: Also a good catch! Fixed.

PR #244, opened 2017-08-06
eos: Look up icon to replace and call ReplaceAppIfPossible

smspillaz:
When refining the apps, look for the metadata key
Endless::ReplacedByDesktopFile. This key indicates the desktop
file from another app that replaces this app on the desktop. We'll
use this functionality in order to replace the "Get X" placeholder
apps with their real application once the installation has
completed.

### DIFF
--- a/plugins/eos/gs-plugin-eos.c
+++ b/plugins/eos/gs-plugin-eos.c
@@ -39,6 +39,7 @@
 #define EOS_IMAGE_VERSION_ALT_PATH "/"
 
 #define METADATA_SYS_DESKTOP_FILE "EndlessOS::system-desktop-file"
+#define METADATA_REPLACED_BY_DESKTOP_FILE "EndlessOS::replaced-by-desktop-file"
 #define EOS_PROXY_APP_PREFIX ENDLESS_ID_PREFIX "proxy"
 
 /*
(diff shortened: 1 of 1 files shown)


plugins/eos/gs-plugin-eos.c
```diff
@@ -39,6 +39,7 @@
 #define EOS_IMAGE_VERSION_ALT_PATH "/"
 
 #define METADATA_SYS_DESKTOP_FILE "EndlessOS::system-desktop-file"
+#define METADATA_REPLACED_BY_DESKTOP_FILE "Endless::ReplacedByDesktopFile"
```
cosimoc: Please make both the namespace and the capitalization style of this key consistent with the other custom metadata key(s) we already have (e.g. the one defined just above)

plugins/eos/gs-plugin-eos.c
```diff
@@ -770,6 +775,19 @@ gs_plugin_eos_refine_popular_app (GsPlugin *plugin,
 	                            request_data);
 }
 
+static void
+gs_plugin_eos_update_replacement_app_lookup_table (GsApp      *app,
+						   GHashTable *replacement_app_lookup)
+{
+	const gchar *replaced_by = gs_app_get_metadata_item (app, METADATA_REPLACED_BY_DESKTOP_FILE);
+
+	if (replaced_by) {
```
cosimoc: Remove curly braces for single-line if

plugins/eos/gs-plugin-eos.c
```diff
@@ -770,6 +775,19 @@ gs_plugin_eos_refine_popular_app (GsPlugin *plugin,
 	                            request_data);
 }
 
+static void
+gs_plugin_eos_update_replacement_app_lookup_table (GsApp      *app,
+						   GHashTable *replacement_app_lookup)
```
cosimoc: The signature of this function should be
```
static void
gs_plugin_eos_update_replacement_app_lookup_table (GsPlugin *plugin,
                                                                                            GsApp *app)
```

plugins/eos/gs-plugin-eos.c
smspillaz: Good point, I suspected we might want to keep this consistent. There's a couple of other places where we're using the `Endless::` namespace with the `UpperCamelCase` notation, the first one that comes to mind is, `Endless::HasDiscoveryFeedContent`. Perhaps once we have that feature working again, we'll want to make that key consistent as well.

PR #248, opened 2017-08-28
T18657

smspillaz:

### DIFF
--- a/plugins/eos/gs-plugin-eos.c
+++ b/plugins/eos/gs-plugin-eos.c
@@ -182,6 +182,51 @@ get_os_version_id (GError **error)
 	return g_strdup (gs_os_release_get_version_id (os_release));
 }
 
+static void
+read_icon_replacement_overrides (GHashTable *replacement_app_lookup)
+{
+	const gchar * const *datadirs = g_get_system_data_dirs ();
+	g_autoptr(GError) error = NULL;
+
+	for (; *datadirs; ++datadirs) {
+		g_autofree gchar *candidate_path = g_build_filename (*datadirs,
+                                                                     "eos-application-tools",
+                                                                     "icon-overrides",
+                                                                     "eos-icon-overrides.ini",
+                                                                     NULL);
+		g_autoptr(GKeyFile) config = g_key_file_new ();
+		g_auto(GStrv) keys = NULL;
+		gsize n_keys = 0;
+		gsize key_iterator = 0;
+
+		if (!g_key_file_load_from_file (config, candidate_path, G_KEY_FILE_NONE, &error)) {
+			if (!g_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
+				g_warning ("Could not load icon overrides file %s: %s", candidate_path, error->message);
+			g_clear_error (&error);
+			continue;
+		}
+
+		if (!(keys = g_key_file_get_keys (config, "Overrides", &n_keys, &error))) {
+			g_warning ("Could not read keys from icon overrides file %s: %s", candidate_path, error->message);
+			g_clear_error (&error);
+			continue;
+		}
+
+		/* Now add all the key-value pairs to the replacement app lookup table */
+		for (; key_iterator != n_keys; ++key_iterator) {
+			g_hash_table_replace (replacement_app_lookup,
+					      g_strdup (keys[key_iterator]),
+					      g_key_file_get_string (config,
+								     "Overrides",
+								     keys[key_iterator],
+								     NULL));
+		}
+
+		/* First one takes priority, ignore the others */
+		break;
+	}
+}
+
 /**
  * gs_plugin_initialize:
  */
@@ -217,6 +262,10 @@ gs_plugin_initialize (GsPlugin *plugin)
 	                                                    gs_user_agent (),
 	                                                    NULL);
 	priv->personality = get_personality ();
+
+	/* Synchronous, but this guarantees that the lookup table will be
+	 * there when we call ReplaceApplication later on */
+	read_icon_replacement_overrides (priv->replacement_app_lookup);
 
 	if (!priv->personality)
 		g_warning ("No system personality could be retrieved!");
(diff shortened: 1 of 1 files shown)


plugins/eos/gs-plugin-eos.c
```diff
@@ -182,6 +182,48 @@ get_os_version_id (GError **error)
...
 }
 
+static void
+read_icon_replacement_overrides (GHashTable *replacement_app_lookup)
+{
+	const gchar * const *datadirs = g_get_system_data_dirs ();
+	g_autoptr(GError) error = NULL;
+
+	for (; *datadirs; ++datadirs) {
+		g_autofree gchar *candidate_path = g_build_filename (*datadirs, "eos-application-tools/icon-overrides/eos-icon-overrides.ini", NULL);
```
cosimoc: You should pass each path component as a distinct parameter.

plugins/eos/gs-plugin-eos.c
```diff
@@ -182,6 +182,48 @@ get_os_version_id (GError **error)
...
+		g_autoptr(GKeyFile) config = g_key_file_new ();
+		g_auto(GStrv) keys = NULL;
+		gsize n_keys = 0;
+		gsize key_iterator = 0;
+
+		if (!g_key_file_load_from_file (config, candidate_path, G_KEY_FILE_NONE, &error)) {
+			if (!(error->domain == G_FILE_ERROR && error->code == G_FILE_ERROR_NOENT)) {
```
cosimoc: Use `g_error_matches()`

plugins/eos/gs-plugin-eos.c
```diff
@@ -182,6 +182,48 @@ get_os_version_id (GError **error)
...
+			g_warning ("Could not read keys from icon overrides file %s: %s", candidate_path, error->message);
+			g_clear_error (&error);
+			continue;
+		}
+
+		/* Now add all the key-value pairs to the replacement app lookup table */
+		for (; key_iterator != n_keys; ++key_iterator) {
+			g_hash_table_replace (replacement_app_lookup,
+					      g_strdup (keys[key_iterator]),
+					      g_strdup (g_key_file_get_string (config,
```
cosimoc: `g_key_file_get_string()` already returns a newly allocated string

smspillaz: Huh, it does too. Okay, fixed!

plugins/eos/gs-plugin-eos.c
```diff
@@ -182,6 +182,52 @@ get_os_version_id (GError **error)
...
+                                                                     NULL);
+		g_autoptr(GKeyFile) config = g_key_file_new ();
+		g_auto(GStrv) keys = NULL;
+		gsize n_keys = 0;
+		gsize key_iterator = 0;
+
+		if (!g_key_file_load_from_file (config, candidate_path, G_KEY_FILE_NONE, &error)) {
+			if (!g_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_NOENT)) {
+				g_warning ("Could not load icon overrides file %s: %s", candidate_path, error->message);
+			}
```
joaquimrocha: Drop the brackets on 1 liner blocks.

smspillaz:
> Unless I missed something, it seems that as long as there's an entry for the apps' replacement icon in this new hashtable, the icon will always call the replacement logic. This means that if we later remove the icon from the desktop and try to add it again, it will still attempt to replace it again, right? If this is safe-checked within the Shell's icon replacement code (i.e. upon failure to find the icon to be replaced, it'll just append the new icon) then there's no need for two calls and we should always call the replacement code.

In this case, we can only call the `ReplaceApplication` function if we have something *to* replace, so we wouldn't be able to call it all the time.

>The EOS plugin keeps a desktop_apps hashtable with the apps in the desktop, so I think this table should be looked up before calling the replacement call.

I suppose that's a good point, though it would be a mostly pointless check since the shell will do it for us anyway.

I think the purpose of the hardcoded hashtable here is to say "is there something that we *could* replace? If so, replace that if possible, otherwise just append". Another options is to just allow the `ReplaceApplication` method to accept `null` as the "source" icon and deprecate `AddApplication`, though that's kind of an awkward API for people to use. @cosimoc  any thoughts on that?

joaquimrocha:
I think we had settled on that **if** the replace icon method does fallback to appending the icon, then we would always call that method. IMO the distinction makes the reader's expectations different than what the code really implies.

smspillaz:
Something @joaquimrocha and I discussed on slack was whether we want to call a different dbus method depending on whether we're replacing or appending an app. Since `ReplaceApplication` always falls back to appending the icon if a replacement was not possible, we could just call `ReplaceApplication` blindly here: https://github.com/endlessm/gnome-software/pull/248/files#diff-9cd85aa3e047f05095da410c840a9c56R1168 with the result of the hash table lookup. However, the downside is that this could confuse the reader who would wonder why we never checked if what we were replacing actually existed on the desktop. Any thoughts?

cosimoc:
I am trying to understand @joaquimrocha's concern here... It does not seem that this PR changes in any way the semantics of the `ReplaceApplication` and `AddApplication`, so it seems odd that's mentioned as a blocker for this change?

In any case I think you're mainly discussing two points:
- should we always call `ReplaceApplication()` instead of `AddApplication()` if `ReplaceApplication()` appends when there's nothing to replace? I don't really see a good reason for that; since the DBus interface is a private one, and all of its clients are inside the OSTree, if we wanted to make a change to only have one API, I'd probably rather extend `AddApplication()` to take an optional ID of what to replace and update all the clients at once. But I don't think that the current situation is particularly bad or confusing.
- should we check with the desktop hash table whether the app we're replacing is already on the desktop? I don't think so; doing it would be inherently racy in the same way that checking whether a file exists before doing something with it is, since the user or something else could have changed the desktop layout between the time we check and the DBus call, or the notification could have not gotten to us yet.

While I think the current patch is OK as is, I don't want to dismiss these concerns and would recommend that separate low-priority tickets are opened for further consideration without blocking this getting into 2017-09.
However, since the debate around this implementation has been quite heated, I will leave it to @joaquimrocha to give (or not give) the final approval for these changes.

PR #249, opened 2017-08-29
T18657: Add 'refine' action

smspillaz:
This changeset adds a "refine" action gnome-software which requires that a given app is refined. We'll use this from the placeholder applications to ensure that they are refined so that gnome-software knows which app to replace.


cosimoc:
Closed in favor of #248

PR #251, opened 2017-09-06
eos: Add Evergreen apps to the proxy app

joaquimrocha:
The apps are the Quote of the Day and Word of the Day.

https://phabricator.endlessm.com/T18944

### DIFF
--- a/plugins/eos/gs-plugin-eos.c
+++ b/plugins/eos/gs-plugin-eos.c
@@ -1329,6 +1329,8 @@ add_updates (GsPlugin *plugin,
 	const char *proxied_apps[] = {"com.endlessm.Platform",
 				      "com.endlessm.apps.Platform",
 				      "com.endlessm.EknServices.desktop",
+				      "com.endlessm.quote_of_the_day.en.desktop",
+				      "com.endlessm.word_of_the_day.en.desktop",
 				      NULL};
 
 	for (guint i = 0; i < gs_app_list_length (list); ++i) {


plugins/eos/gs-plugin-eos.c
```diff
@@ -1329,6 +1329,8 @@ add_updates (GsPlugin *plugin,
 	const char *proxied_apps[] = {"com.endlessm.Platform",
 				      "com.endlessm.apps.Platform",
 				      "com.endlessm.EknServices.desktop",
+				      "com.endlessm.quote_of_the_day.en.desktop",
+				      "com.endlessm.word_of_the_day.en.desktop",
```
cosimoc: This is still not enough; the applications will show up when searching for them.
Since I would not want to break the proxy app functionality at this time, I think all we can do for this release is to blacklist the evergreen apps, and have a proper solution for the next one.

joaquimrocha: I was aware of that but I was thinking that the blacklist would be applied in the appdata of those apps for some reason. I don't mind adding the blacklisting to the EOS plugin as we've done with multiple other apps though.
In any case, searching for those apps should be unrelated to being able to successfully update the apps in the proxy app, right? Or am I missing something?

cosimoc: Yeah, but those apps at the moment will show up launchable when found, which we don't want. The proper solution is being discussed in https://phabricator.endlessm.com/T19104 (which btw requires your input! ;-))

PR #254, opened 2017-09-15
Debian rebase gnome 3 26

joaquimrocha:
https://phabricator.endlessm.com/T19001


debian/control
```diff
@@ -6,15 +6,14 @@ Uploaders: Matthias Klumpp <mak@debian.org>,
            Laurent Bigonville <bigon@debian.org>, Michael Biebl <biebl@debian.org>
 Build-Depends: appstream,
                appstream-util,
-               debhelper (>= 9.0.0),
-               dh-autoreconf,
+               debhelper (>= 10.4),
```
cosimoc: Are these changes taken from the new version of the debian packaging, or did you make them yourself? We usually update the packaging to the new debian version first (I see 3.26 is in sid) and then apply other changes on top.

debian/changelog
```diff
@@ -1,3 +1,9 @@
+gnome-software (3.26.0-0) master; urgency=medium
```
cosimoc: I see that 3.26.0 is in sid, so we should look and update to that instead of our own packaging.

joaquimrocha: Right. I looked into the 3.25.92 version before the 3.26.0 was out but I found it so different (it now creates a new package for flatpak and some other plugins, as well as the complexity in the rules) that I preferred to just adapt our package (that has been manually updated by us for a long time now).

If there's a problem with the current debian package and/or you prefer nonetheless that I really use the one from upstream, I can do that though it'll take a couple of hours more to have everything tested again.

pwithnall:
Since I’m now [rebasing on 3.32](https://phabricator.endlessm.com/T25367), this can probably be closed.

PR #255, opened 2017-09-25
T19321

joaquimrocha:
https://phabricator.endlessm.com/T19321


src/gs-background-tile.c
```diff
@@ -0,0 +1,276 @@
...
+	gs_utils_widget_set_css (GTK_WIDGET (tile->image_box), css);
+}
+
+
+static void
+gs_background_tile_set_app (GsAppTile *app_tile, GsApp *app)
+{
+	GsBackgroundTile *tile = GS_BACKGROUND_TILE (app_tile);
+	const gchar *summary = NULL;
+	const gchar *css;
```
mariospr: Initialize to NULL

src/gs-background-tile.c
```diff
@@ -0,0 +1,276 @@
...
+	/* perhaps set custom css */
+	css = gs_app_get_metadata_item (app, "GnomeSoftware::BackgroundTile-css");
+	if (css != NULL)
+		gs_utils_widget_set_css (GTK_WIDGET (tile->image_box), css);
+	else
+		gs_background_tile_setup_colors_bg (tile);
+
+
+	if (gs_app_get_pixbuf (tile->app) != NULL)
+		gs_image_set_from_pixbuf (GTK_IMAGE (tile->image), gs_app_get_pixbuf (tile->app));
```
mariospr: Two calls to `gs_app_get_pixbuf()`? Maybe you could store the result in a variable and use that instead

lib/gs-app.c
```diff
@@ -4032,6 +4041,14 @@ gs_app_class_init (GsAppClass *klass)
 				     G_PARAM_READWRITE | G_PARAM_CONSTRUCT);
 	g_object_class_install_property (object_class, PROP_QUIRK, pspec);
 
+	/**
+	 * GsApp:key-colors:
+	 */
+	pspec = g_param_spec_string ("key-colors", NULL, NULL,
```
cosimoc: I think you want this to be a `g_param_spec_pointer`? Otherwise how can the calls to `g_value_get_pointer()` and `g_value_set_pointer()` above work at all?

src/gs-background-tile.h
```diff
@@ -0,0 +1,39 @@
+/* -*- Mode: C; tab-width: 8; indent-tabs-mode: t; c-basic-offset: 8 -*-
+ *
+ * Copyright (C) 2013 Matthias Clasen <mclasen@redhat.com>
```
cosimoc: Add your copyright here

src/gs-background-tile.c
```diff
@@ -0,0 +1,276 @@
...
+}
+
+static void
+app_background_tile_css_added (GsApp *app, const char *metadata, GsBackgroundTile *tile)
+{
+	const gchar *css = gs_app_get_metadata_item (app, "GnomeSoftware::BackgroundTile-css");
+	/* gs_background_tile_setup_colors_bg (tile); */
+	gs_utils_widget_set_css (GTK_WIDGET (tile->image_box), css);
+}
+
```
cosimoc: Remove empty extra line

src/gs-background-tile.c
```diff
@@ -0,0 +1,276 @@
...
+	/* this is because the state-change notification could be
+	 * coming from a different thread */
+	g_idle_add (app_state_changed_idle, g_object_ref (tile));
+}
+
+static void
+app_background_tile_css_added (GsApp *app, const char *metadata, GsBackgroundTile *tile)
+{
+	const gchar *css = gs_app_get_metadata_item (app, "GnomeSoftware::BackgroundTile-css");
+	/* gs_background_tile_setup_colors_bg (tile); */
```
cosimoc: Remove this line, or call it only when `css == NULL`

src/gs-background-tile.c
```diff
@@ -0,0 +1,276 @@
...
+	g_signal_connect (tile->app, "notify::state",
+			  G_CALLBACK (app_state_changed), tile);
+	g_signal_connect (tile->app, "notify::name",
+			  G_CALLBACK (app_state_changed), tile);
+	g_signal_connect (tile->app, "notify::summary",
+			  G_CALLBACK (app_state_changed), tile);
+	g_signal_connect (tile->app, "notify::key-colors",
+			  G_CALLBACK (app_state_changed), tile);
+	g_signal_connect (tile->app, "metadata-changed::GnomeSoftware::BackgroundTile-css",
+			  G_CALLBACK (app_background_tile_css_added), tile);
```
cosimoc: I wouldn't call the function `_css_added`, but `_css_changed`

src/gs-background-tile.c
```diff
@@ -0,0 +1,276 @@
...
+
+static gboolean
+app_state_changed_idle (gpointer user_data)
+{
+	GsBackgroundTile *tile = GS_BACKGROUND_TILE (user_data);
+	g_autofree gchar *name = NULL;
+	AtkObject *accessible = gtk_widget_get_accessible (GTK_WIDGET (tile));
+
+	if (gs_app_is_installed (tile->app))
+		name = g_strdup_printf (_("%s (Installed)"), gs_app_get_name (tile->app));
```
cosimoc: Needs a translator's comment

src/gs-background-tile.c
```diff
@@ -0,0 +1,276 @@
...
+	else if (gs_app_get_state (tile->app) == AS_APP_STATE_AVAILABLE)
+		name = g_strdup (gs_app_get_name (tile->app));
+
+	if (GTK_IS_ACCESSIBLE (accessible)) {
+		atk_object_set_name (accessible, name);
+		atk_object_set_description (accessible, gs_app_get_summary (tile->app));
+	}
+
+	if (gs_app_get_metadata_item (tile->app, "GnomeSoftware::BackgroundTile-css") == NULL)
+		gs_background_tile_setup_colors_bg (tile);
```
cosimoc: Do you really need to do this again here? You do it upon setup, and if you follow my suggestion of calling it also when `css == NULL` in the function below, unless there's a bug in how the app notifies the metadata change, it should not be required here.

src/gs-background-tile.c
```diff
@@ -0,0 +1,276 @@
...
+
+	if (gs_app_is_installed (tile->app)) {
+		gtk_widget_hide (tile->price);
+		gtk_widget_show (tile->installed_icon);
+		return;
+	}
+
+	price = gs_app_get_price (tile->app);
+	if (price == NULL || (gs_price_get_amount (price) == (gdouble) 0))
+		gtk_label_set_label (GTK_LABEL (tile->price), _("FREE"));
```
cosimoc: I'd add a translator's comment to keep this upper case, or even use a simple string here and then use e.g. `g_strup()` to make it upper case.

src/gs-background-tile.c
```diff
@@ -0,0 +1,276 @@
...
+	else
+		gtk_label_set_label (GTK_LABEL (tile->price), gs_price_to_string (price));
```
cosimoc: `gs_price_to_string()` allocates a new string that will be leaked here.

src/gs-background-tile.c
```diff
@@ -0,0 +1,276 @@
...
+	const guint num_colors = 4;
+	const GPtrArray *colors = gs_app_get_key_colors (app);
+	const gchar *gradients[num_colors];
+	g_autofree gchar *css = NULL;
+	gint degrees = 45;
+
+	if (colors->len == 0)
+		return;
+
+	/* apply one color gradient for each color from the cornors in the
```
cosimoc: Typo: `cornors`

src/gs-background-tile.c
```diff
@@ -0,0 +1,276 @@
...
+	 * - \   -  - \   -  - \   -  - \ / -
+	 * -------  -------  -------  -------
+	 */
+	for (guint i = 0; i < num_colors; ++i) {
+		GdkRGBA *rgba = g_ptr_array_index (colors, i % colors->len);
+		guint red, green, blue;
+
+		/* we use the values in the printf below instead of using the
+		 * convenient gdk_rgba_to_string to avoid extra allocations as
+		 * this function can be used many times */
```
cosimoc: I'm pretty sure that calling `gs_utils_widget_set_css()` will dominate any string allocations we can ever do here in terms of performance; I'd prefer to use the canonical `gdk_rgba_to_string()` unless you profile it to be an issue.

lib/gs-app.c
joaquimrocha: Wow. Sorry about that. I am surprised it all apparently worked...

src/gs-background-tile.c
joaquimrocha: I wonder if it's worth that, and whether that will drop some needed context for other languages. But maybe not.

src/gs-background-tile.c
joaquimrocha: This is in case we do things like refining the key-colors only after we have the tiles being shown (lazy loading) it's useful to have this.
I will add a flag to the the tile in order to skip setting the colors when they've already been added.

src/gs-background-tile.c
joaquimrocha: This was copied from the GsImageTile but I will change it.

src/gs-background-tile.c
cosimoc: Still needs translator's comment

src/gs-background-tile.c
cosimoc: My point is that if the key colors are changed, we should have received a `notify` callback already.
Right now the notify callback for this metadata key, always assumes that `css != NULL`; I think that's a bug, and if you fix the callback to call `gs_background_tile_setup_colors_bg()` when `css == NULL`, you would not need this here.
At that point you could also remove the hunk after the `/* perhaps set custom css */` comment from `gs_background_set_tile_app()` and replace it with a manual call to the notify callback.

src/gs-background-tile.c
```diff
@@ -0,0 +1,281 @@
...
+	gtk_widget_class_bind_template_child (widget_class, GsBackgroundTile, stack);
+}
+
+GtkWidget *
+gs_background_tile_new (GsApp *app)
+{
+	GsBackgroundTile *tile = g_object_new (GS_TYPE_BACKGROUND_TILE, NULL);
+
+	if (app != NULL)
+		gs_app_tile_set_app (GS_APP_TILE (tile), app);
```
cosimoc: Looks like `gs_app_tile_set_app()` can deal with `app == NULL` just fine...

src/gs-background-tile.c
```diff
@@ -0,0 +1,281 @@
...
+	GsApp *app = tile->app;
+	const guint num_colors = 4;
+	const GPtrArray *colors = gs_app_get_key_colors (app);
+	/* initialise with 1 extra member because we need the last element to be NULL*/
+	g_auto(GStrv) gradients = g_new0 (gchar *, num_colors + 1);
+	g_autofree gchar *css = NULL;
+	g_autofree gchar *gradients_str = NULL;
+	gint degrees = 45;
+
+	if (colors->len == 0 || tile->key_colors_set)
```
cosimoc: Sorry for the back and forth on this implementation, but this does not look quite right yet.

Right now `tile->key_colors_set` is only reset to `FALSE` when setting a new app on the tile, through `gs_background_tile_set_app()`, and set to `TRUE` at the end of this function. That means that any changes to the app's newly introduced `key-colors` property will effectively be ignored.

The code that looks at both the key colors and css and decides which one to set (right now part of `update_tile_info()`) should run when either of those notify a change.

src/gs-background-tile.c
```diff
@@ -0,0 +1,281 @@
...
+	/* apply one color gradient for each color from the corners in the
+	 * following fashion:
+	 * -------  -------  -------  -------
+	 * -     -  - /   -  - / \ -  - / \ -
+	 * - \   -  - \   -  - \   -  - \ / -
+	 * -------  -------  -------  -------
+	 */
+	for (guint i = 0; i < num_colors; ++i) {
+                GdkRGBA *rgba = g_ptr_array_index (colors, i % colors->len);
+		g_autofree gchar *rgba_str = NULL;
```
cosimoc: Indentation looks odd here -- maybe tabs and spaces are being mixed?

src/gs-background-tile.c
```diff
@@ -0,0 +1,281 @@
...
+	 * notification could be coming from a different thread */
+	g_idle_add (app_state_changed_idle, g_object_ref (tile));
+}
+
+static void
+app_background_tile_css_changed (GsApp *app, const char *metadata, GsBackgroundTile *tile)
+{
+	/* just call the app_state_changed function as it will deal with the tile
+	 * CSS already; calling it from idle as this can come from a different thread */
+	g_idle_add (app_state_changed_idle, g_object_ref (tile));
```
cosimoc: I don't believe metadata notification can happen in a different thread actually, or at least the previous code assumed that it did not.

src/gs-background-tile.c
joaquimrocha: This was because the notify code can also be triggered when the name, summary, and more importantly, the state changes. And we don't want to recalculate the gradients whenever all these change. That's why I was making sure that the gradients were only set once...

The key-colors are being set from the apps' icons, and icons don't usually change. When they do, because an app may have completely changed, a refresh of the catalog in g-s will happen, so we'll rebuild the widget.

I can have a dedicated callback for _notify::key-colors_ if you prefer, and we'll just stop having this treated in the same function as all the other states. Do you prefer it this way?

src/gs-background-tile.c
```diff
@@ -0,0 +1,303 @@
...
+}
+
+static void
+app_background_tile_css_changed (GsApp *app,
+				 const char *metadata,
+				 GsBackgroundTile *tile)
+{
+	/* just call the app_state_changed function as it will deal with the
+	 * tile CSS already */
+	app_state_changed_idle (g_object_ref (tile));
```
cosimoc: I think both this and `app_key_colors_changed()` should end up calling into a function that, after calling `gs_app_get_metadata_item (tile->app, "GnomeSoftware::BackgroundTile-css")` decides whether to set the widget's CSS to the custom background or the key colors one.
That function could be `update_tile_colors_bg()` itself, if it had that additional check, or e.g. another `update_tile_bg()` that does the check.

joaquimrocha: If we do what you say here, then as long as there's a "GnomeSoftware::BackgroundTile-css" metadata, the gradient colors will never be applied, and I think you had mentioned before that if a notify::key-colors occurred, then it should be set independently of what the CSS is. Sorry if I understood it wrong.

cosimoc: Correct, that's how I understood the intention of your logic; we either set the custom background, or use the gradient. Your commit message for `GsBackgroundTile` also mentions the same:
```
... has the ability to display either a thumbnail image or a gradient
background generated from an app's icon.
```

So yeah, when either the `GnomeSoftware::BackgroundTile-css` metadata piece or the `key-colors` property change, we should look at both and evaluate which one to use. https://github.com/endlessm/gnome-software/commit/e470c757c072af525ea06520b70d41c7173fc84a has the right logic, and I will comment there with a suggestion for an improvement.

src/gs-background-tile.c
```diff
@@ -0,0 +1,292 @@
...
+	gtk_stack_set_visible_child_name (GTK_STACK (tile->stack), "content");
+
+	g_signal_connect (tile->app, "notify::state",
+			  G_CALLBACK (app_state_changed), tile);
+	g_signal_connect (tile->app, "notify::name",
+			  G_CALLBACK (app_state_changed), tile);
+	g_signal_connect (tile->app, "notify::summary",
+			  G_CALLBACK (app_state_changed), tile);
+	g_signal_connect (tile->app, "notify::key-colors",
+			  G_CALLBACK (app_key_colors_changed), tile);
```
cosimoc: You could use `g_signal_connect_swapped()` and `call update_tile_background()` directly from this and the next callback.

lib/gs-app.c
```diff
@@ -3421,6 +3459,7 @@ gs_app_add_key_color (GsApp *app, GdkRGBA *key_color)
 	g_return_if_fail (GS_IS_APP (app));
 	g_return_if_fail (key_color != NULL);
 	g_ptr_array_add (priv->key_colors, gdk_rgba_copy (key_color));
+	g_object_notify (G_OBJECT (app), "key-colors");
```
cosimoc: This is notified directly here, but queued above; any reason for that?

lib/gs-app.c
```diff
@@ -3004,10 +3041,10 @@ gs_app_set_metadata (GsApp *app, const gchar *key, const gchar *value)
 	}
 	g_hash_table_insert (priv->metadata, g_strdup (key), g_strdup (value));
 
-	g_signal_emit (app,
-	               signals[SIGNAL_METADATA_CHANGED],
-	               g_quark_from_string (key),
-	               key);
+ emit_and_quit:
+	/* emit the signal from the mainloop */
+	helper = gs_metadata_signal_data_new (app, key);
```
cosimoc: While I like this idea, it did not look like we actually hit any issues with the previous approach or assuming that metadata updates happen in the main loop. I am OK with the change if you think it's a good idea though.

PR #256, opened 2017-09-26
flatpak: Improve performance of refining an app's state

joaquimrocha:
Instead of listing the installed refs and trying to match the app there,
it instead looks for the app's installed ref directly (which lets
Flatpak manage how to look for it and is much faster).

From some measurements this makes refining apps about 2x faster.

https://phabricator.endlessm.com/T19396

### DIFF
--- a/plugins/flatpak/gs-flatpak.c
+++ b/plugins/flatpak/gs-flatpak.c
@@ -1713,29 +1713,6 @@ gs_plugin_refine_item_origin (GsFlatpak *self,
 	return TRUE;
 }
 
-static gboolean
-gs_flatpak_app_matches_xref (GsFlatpak *self, GsApp *app, FlatpakRef *xref)
-{
-	g_autoptr(GsApp) app_tmp = gs_flatpak_create_app (self, xref);
-
-	/* check ID */
-	if (g_strcmp0 (gs_app_get_unique_id (app),
-	    gs_app_get_unique_id (app_tmp)) == 0)
-		return TRUE;
-
-	/* do all the metadata items match? */
-	if (g_strcmp0 (gs_flatpak_app_get_ref_name (app),
-		       flatpak_ref_get_name (xref)) == 0 &&
-	    g_strcmp0 (gs_flatpak_app_get_ref_arch (app),
-		       flatpak_ref_get_arch (xref)) == 0 &&
-	    g_strcmp0 (gs_flatpak_app_get_ref_branch (app),
-		       flatpak_ref_get_branch (xref)) == 0)
-		return TRUE;
-
-	/* sad panda */
-	return FALSE;
-}
-
 static FlatpakRef *
 gs_flatpak_create_fake_ref (GsApp *app, GError **error)
 {
@@ -1760,9 +1737,10 @@ gs_plugin_refine_item_state (GsFlatpak *self,
 			     GCancellable *cancellable,
 			     GError **error)
 {
-	guint i;
 	g_autoptr(GPtrArray) xrefs = NULL;
 	g_autoptr(AsProfileTask) ptask = NULL;
+	g_autoptr(FlatpakInstalledRef) ref = NULL;
+	g_autoptr(GError) error_local = NULL;
 
 	/* already found */
 	if (gs_app_get_state (app) != AS_APP_STATE_UNKNOWN)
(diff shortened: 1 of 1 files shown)


plugins/flatpak/gs-flatpak.c
```diff
@@ -1777,31 +1755,29 @@ gs_plugin_refine_item_state (GsFlatpak *self,
...
+		g_propagate_error (error, g_steal_pointer (&local_error));
+		gs_flatpak_error_convert (error);
+		return FALSE;
 	}
 
 	/* ensure origin set */
 	if (!gs_plugin_refine_item_origin (self, app, cancellable, error))
 		return FALSE;
 
+	g_clear_error (&local_error);
```
mariospr: This should be moved up, right after you know you won't use `local_error` again (line 1774)

plugins/flatpak/gs-flatpak.c
```diff
@@ -1777,31 +1755,29 @@ gs_plugin_refine_item_state (GsFlatpak *self,
...
+	if (ref != NULL) {
 		g_debug ("marking %s as installed with flatpak",
 			 gs_app_get_id (app));
-		gs_flatpak_set_metadata_installed (self, app, xref);
+		gs_flatpak_set_metadata_installed (self, app, ref);
 		if (gs_app_get_state (app) == AS_APP_STATE_UNKNOWN)
 			gs_app_set_state (app, AS_APP_STATE_INSTALLED);
+	} else if (!g_error_matches (local_error, FLATPAK_ERROR, FLATPAK_ERROR_NOT_INSTALLED)) {
+		/* if there was an error that wasn't simply that the app isn't installed, then
+		 * we need to return it*/
```
mariospr: This comment is redundant, it simply repeats what the code says. Either remove it or explain instead "why" only `FLATPAK_ERROR_NOT_INSTALLED` errors are not returned (that would be more useful).

plugins/flatpak/gs-flatpak.c
```diff
@@ -1760,9 +1737,10 @@ gs_plugin_refine_item_state (GsFlatpak *self,
 			     GCancellable *cancellable,
 			     GError **error)
 {
-	guint i;
 	g_autoptr(GPtrArray) xrefs = NULL;
 	g_autoptr(AsProfileTask) ptask = NULL;
+	g_autoptr(FlatpakInstalledRef) ref = NULL;
```
mariospr: Perhaps you could name this one `xref`, for consistency with the rest of the code (and the code being removed)

plugins/flatpak/gs-flatpak.c
```diff
@@ -1825,20 +1801,23 @@ gs_plugin_refine_item_state (GsFlatpak *self,
...
+
+			if (runtime_ref != NULL) {
+				g_debug ("marking runtime %s as installed in the "
+					 "counterpart installation", gs_app_get_id (app));
+				gs_flatpak_set_metadata_installed (self, app, runtime_ref);
+				if (gs_app_get_state (app) == AS_APP_STATE_UNKNOWN)
+					gs_app_set_state (app, AS_APP_STATE_INSTALLED);
+			} else if (!g_error_matches (local_error, FLATPAK_ERROR, FLATPAK_ERROR_NOT_INSTALLED)) {
+				/* if there was an error that wasn't simply that the app isn't installed, then
+				 * we need to return it*/
```
mariospr: Same comment on the usefulness of this comment

plugins/flatpak/gs-flatpak.c
```diff
@@ -1825,20 +1801,23 @@ gs_plugin_refine_item_state (GsFlatpak *self,
...
+				g_propagate_error (error, g_steal_pointer (&local_error));
 				gs_flatpak_error_convert (error);
 				return FALSE;
 			}
```
cosimoc: `local_error` should be cleared here too

plugins/flatpak/gs-flatpak.c
joaquimrocha: I think the _xref_ is legacy from when Flatpak was called xdg-app... But I can rename it...

plugins/flatpak/gs-flatpak.c
joaquimrocha: Why? It's an auto_ptr and no code uses it anymore. I only cleaned it above because it was going to be reused.

PR #260, opened 2017-10-04
T19490

joaquimrocha:
https://phabricator.endlessm.com/T19490

### DIFF
--- a/plugins/flatpak/gs-flatpak.c
+++ b/plugins/flatpak/gs-flatpak.c
@@ -2387,7 +2387,7 @@ gs_flatpak_refine_wildcard (GsFlatpak *self, GsApp *app,
 		/* new app */
 		g_debug ("found %s for wildcard %s",
 			 as_app_get_unique_id (item), id);
-		new = gs_appstream_create_app (self->plugin, item, NULL);
+		new = gs_appstream_get_or_create_app (self->plugin, item, NULL);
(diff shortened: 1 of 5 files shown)


plugins/flatpak/gs-flatpak.c
```diff
@@ -2387,9 +2387,17 @@ gs_flatpak_refine_wildcard (GsFlatpak *self, GsApp *app,
 		/* new app */
 		g_debug ("found %s for wildcard %s",
 			 as_app_get_unique_id (item), id);
-		new = gs_appstream_create_app (self->plugin, item, NULL);
+		new = gs_appstream_get_or_create_app (self->plugin, item, NULL);
 		if (new == NULL)
 			return FALSE;
+
+		/* if the app is not a FlatpakApp type one (meaning it was created by
+		 * another plugin) we need to create a new one */
```
cosimoc: Does this not degrade performance? It would be great if we did not have to create two objects.

joaquimrocha: I don't think there's a big impact performance wise.
Also, gnome-software was not using subclasses of GsApp per plugins and it relied on setting up a number of special metadata keys for proper identification and use, but Richard wanted it to be more organized and thus added the ability for plugins to declare a special type for apps. Of course this comes with some bugs like this one, I think it's better to keep it for now.

PR #261, opened 2017-10-05
T19395

joaquimrocha:
https://phabricator.endlessm.com/T19395

### DIFF
--- a/plugins/eos/gs-plugin-eos.c
+++ b/plugins/eos/gs-plugin-eos.c
@@ -1379,3 +1379,56 @@ gs_plugin_add_updates (GsPlugin *plugin,
 	 * this will still work correctly */
 	return add_updates (plugin, list, cancellable, error);
 }
+
+gboolean
+gs_plugin_add_popular (GsPlugin *plugin,
+		       GsAppList *list,
+		       GCancellable *cancellable,
+		       GError **error)
+{
+	g_autoptr(AsProfileTask) ptask = NULL;
+	GsAppList *new_list = NULL;
+	const gchar *popular_apps[] = {"com.google.Chrome.desktop",
+				       "com.spotify.Client.desktop",
+				       "com.transmissionbt.Transmission.desktop",
+				       "com.valvesoftware.Steam.desktop",
+				       "libreoffice-calc.desktop",
+				       "libreoffice-impress.desktop",
+				       "libreoffice-writer.desktop",
+				       "net.gcompris.Gcompris.desktop",
+				       "net.minetest.Minetest.desktop",
+				       "net.sourceforge.Audacity.desktop",
+				       "org.debian.alioth.tux4kids.Tuxmath.desktop",
+				       "org.gimp.Gimp.desktop",
+				       "org.inkscape.Inkscape.desktop",
+				       "org.mozilla.Firefox.desktop",
+				       "org.tuxpaint.Tuxpaint.desktop",
+				       "org.videolan.VLC.desktop",
+				       "simple-scan.desktop",
+				       NULL};
+
+	ptask = as_profile_start_literal (gs_plugin_get_profile (plugin),
+					  "eos::add-popular");
+
+	new_list = gs_app_list_new ();
+
+	/* add the hardcoded list of popular apps */
+	for (guint i = 0; popular_apps[i] != NULL; ++i) {
+		g_autoptr(GsApp) app = gs_app_new (popular_apps[i]);
+		gs_app_add_quirk (app, AS_APP_QUIRK_MATCH_ANY_PREFIX);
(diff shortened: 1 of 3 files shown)


plugins/eos/gs-plugin-eos.c
```diff
@@ -1374,3 +1374,56 @@ gs_plugin_add_updates (GsPlugin *plugin,
...
+				       "com.spotify.Client.desktop",
+				       "com.transmissionbt.Transmission.desktop",
+				       "com.valvesoftware.Steam.desktop",
+				       "libreoffice-calc.desktop",
+				       "libreoffice-impress.desktop",
+				       "libreoffice-writer.desktop",
+				       "net.minetest.Minetest.desktop",
+				       "net.sourceforge.Audacity.desktop",
+				       "org.debian.alioth.tux4kids.Tuxmath.desktop",
+				       "org.gnome.GIMP.desktop",
```
rshuler: Wrong desktop name: should be org.gimp.Gimp.desktop (note both gimp vs. gnome and capitalization of Gimp)

plugins/eos/gs-plugin-eos.c
```diff
@@ -1374,3 +1374,56 @@ gs_plugin_add_updates (GsPlugin *plugin,
...
+				       "org.inkscape.Inkscape.desktop",
+				       "org.mozilla.Firefox.desktop",
+				       "org.pitivi.Pitivi.desktop",
```
rshuler: I think we should not feature pitivi until we are configured to use the upstream version, as our version is a bit broken right now.

plugins/eos/gs-plugin-eos.c
```diff
@@ -1374,3 +1374,56 @@ gs_plugin_add_updates (GsPlugin *plugin,
...
+
+	ptask = as_profile_start_literal (gs_plugin_get_profile (plugin),
+					  "eos::add-popular");
+
+	new_list = gs_app_list_new ();
+
+	/* add the hardcoded list of popular apps */
+	for (guint i = 0; popular_apps[i] != NULL; ++i) {
+		g_autoptr(GsApp) app = gs_app_new (popular_apps[i]);
+		gs_app_add_quirk (app, AS_APP_QUIRK_MATCH_ANY_PREFIX);
```
rshuler: What is the purpose of this quirk?  Would it be obvious to most developers, or should we add a comment?

plugins/eos/gs-plugin-eos.c
```diff
@@ -1374,3 +1374,56 @@ gs_plugin_add_updates (GsPlugin *plugin,
...
+				       "net.sourceforge.Audacity.desktop",
+				       "org.debian.alioth.tux4kids.Tuxmath.desktop",
+				       "org.gnome.GIMP.desktop",
+				       "org.inkscape.Inkscape.desktop",
+				       "org.mozilla.Firefox.desktop",
+				       "org.pitivi.Pitivi.desktop",
+				       "org.tuxpaint.Tuxpaint.desktop",
+				       "org.videolan.VLC.desktop",
+				       "simple-scan.desktop",
+				       NULL};
```
rshuler: Missing from the list: net.gcompris.Gcompris

plugins/eos/gs-plugin-eos.c
joaquimrocha: Thanks.

plugins/eos/gs-plugin-eos.c
joaquimrocha: App IDs are composed of semantically different parts: scope (system/user), bundle type (package/flatpak/snap), etc.
Since we just set up the simple ID in the app, this quirk makes sure that the app will match e.g. all org.gimp.GNOME.desktop apps. In the context of GNOME Software and AppStream-glib it's not very complicated, so I don't think we need a comment for this (otherwise we'll fill the code with such comments).

plugins/eos/gs-plugin-eos.c
```diff
@@ -1374,3 +1374,56 @@ gs_plugin_add_updates (GsPlugin *plugin,
...
+				       "com.transmissionbt.Transmission.desktop",
+				       "com.valvesoftware.Steam.desktop",
+				       "libreoffice-calc.desktop",
+				       "libreoffice-impress.desktop",
+				       "libreoffice-writer.desktop",
+				       "net.gcompris.Gcompris.desktop",
+				       "net.minetest.Minetest.desktop",
+				       "net.sourceforge.Audacity.desktop",
+				       "org.debian.alioth.tux4kids.Tuxmath.desktop",
+				       "org.gimp.GIMP.desktop",
```
rshuler: Still has the wrong capitalization.  Should be "org.gimp.Gimp.desktop"

PR #266, opened 2017-10-10
T19617

joaquimrocha:
https://phabricator.endlessm.com/T19617

### DIFF
--- a/lib/gs-app.c
+++ b/lib/gs-app.c
@@ -124,6 +124,7 @@ typedef struct
 	AsContentRating		*content_rating;
 	GdkPixbuf		*pixbuf;
 	GsPrice			*price;
+	GCancellable		*cancellable;
 } GsAppPrivate;
 
 enum {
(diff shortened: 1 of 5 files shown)


lib/gs-app.c
```diff
@@ -4136,6 +4165,7 @@ gs_app_init (GsApp *app)
 	                                    g_free,
 	                                    g_free);
 	priv->allow_cancel = TRUE;
+	priv->cancellable = g_cancellable_new ();
```
cosimoc: You could avoid this if you added a `!priv->cancellable ||` condition to the `is_cancelled` check in `gs_app_get_cancellable()`

lib/gs-app.c
```diff
@@ -3791,6 +3792,34 @@ gs_app_get_priority (GsApp *app)
...
+ * Since: 3.28
+ **/
+GCancellable *
+gs_app_get_cancellable (GsApp *app)
+{
+	g_autoptr(GCancellable) cancellable = NULL;
+	GsAppPrivate *priv = gs_app_get_instance_private (app);
+
+	if (g_cancellable_is_cancelled (priv->cancellable)) {
+		cancellable = g_cancellable_new ();
```
cosimoc: Needs to be unreffed when the `GsApp` is finalized

src/gs-details-page.c
```diff
@@ -57,6 +57,7 @@ struct _GsDetailsPage
 	GsPluginLoader		*plugin_loader;
 	GtkBuilder		*builder;
 	GCancellable		*cancellable;
+	GCancellable		*app_cancellable;
```
cosimoc: I'd try to avoid storing this separately, since we store already a reference to the app. It should be enough to call `gs_app_get_cancellable()` where required.

src/gs-updates-page.c
```diff
@@ -610,6 +610,7 @@ _reboot_failed_cb (GObject *source, GAsyncResult *res, gpointer user_data)
 	GsUpdatesPage *self = GS_UPDATES_PAGE (user_data);
 	g_autoptr(GError) error = NULL;
 	g_autoptr(GsAppList) apps = NULL;
+	g_autoptr(GsApp) app = NULL;
```
cosimoc: I don't think this should be `g_autoptr` since `gs_app_list_index()` does not return a new reference

lib/gs-app.c
```diff
@@ -3965,7 +3995,8 @@ gs_app_finalize (GObject *object)
 		g_object_unref (priv->pixbuf);
 	if (priv->price != NULL)
 		g_object_unref (priv->price);
-
+	if (priv->price != NULL)
+		g_object_unref (priv->price);
```
cosimoc: This change looks wrong

src/gs-details-page.c
cosimoc: You could also add an API like gs_app_get_is_cancelled()... would simplify the code quite a bit not to have to store it tbh.

PR #267, opened 2017-10-10
updates: Set the refresh button insensitive when apps are updating

joaquimrocha:
Since refreshing GNOME Software influences the updates page, that option
should not be given to the user, and thus we should make the refresh
button insensitive when any app is being updated.

https://phabricator.endlessm.com/T19612

### DIFF
--- a/src/gs-updates-page.c
+++ b/src/gs-updates-page.c
@@ -324,6 +324,25 @@ gs_updates_page_get_state_string (GsPluginStatus status)
 	return _("Looking for new updates…");
 }
 
+static gboolean
+gs_shell_update_are_updates_in_progress (GsUpdatesPage *self)
+{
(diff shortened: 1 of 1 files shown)


src/gs-updates-page.c
```diff
@@ -919,6 +943,8 @@ _add_app_row (GsUpdatesPage *self, GsApp *app)
 		_create_listbox_section (self, section);
 
 	app_row = gs_app_row_new (app);
+	g_object_set_data_full (G_OBJECT (app_row), "page", g_object_ref (self),
+				g_object_unref);
```
cosimoc: I don't think the ref/unref are required here, since the app row is contained within the page hierarchy; when it's destroyed the app row will be destroyed too.

PR #269, opened 2017-10-18
core: Split the News subcategory

joaquimrocha:
The News category which comprised Network::News and Network::Feed is
moved under the the Utilities category, and a new Reference::News is
created.
The reason is that if we only have Network::News, then any app tagged
with this category will show up in the Utilities list as well, and this
is not desired for purely news reference apps.

https://phabricator.endlessm.com/T19566


plugins/core/gs-desktop-common.c
```diff
@@ -271,9 +275,8 @@ static const GsDesktopMap map_reference[] = {
 	{ "lifestyle",		NC_("Menu of Reference", "Lifestyle"),
 					{ "Reference::Lifestyle",
 					  NULL} },
-	{ "news",		NC_("Menu of Communication & News", "News"),
-					{ "Network::Feed",
-					  "Network::News",
+	{ "news",		NC_("Menu of Reference News", "News"),
```
rshuler: @joaquimrocha I think this should just say "Menu of Reference", for consistency with all the other subcategories.

plugins/core/gs-desktop-common.c
```diff
@@ -271,9 +275,8 @@ static const GsDesktopMap map_reference[] = {
...
+					{ "Reference::News",
```
rshuler: In addition to "Reference::News", we also need "Reference::Feed"

PR #274, opened 2017-10-27
T19711

joaquimrocha:
https://phabricator.endlessm.com/T19711

### DIFF
--- a/plugins/flatpak/gs-flatpak.c
+++ b/plugins/flatpak/gs-flatpak.c
@@ -2567,33 +2567,30 @@ gs_flatpak_refine_runtime_for_install (GsFlatpak *self,
 				       GError **error)
 {
 	GsApp *runtime;
+	gsize len;
+	const gchar *str;
+	g_autoptr(GBytes) data = NULL;
 
-	/* only required for ostree-based flatpak apps */
-	if (gs_flatpak_app_get_file_kind (app) == GS_FLATPAK_APP_FILE_KIND_BUNDLE) {
-		gsize len;
-		const gchar *str;
-		g_autoptr(GBytes) data = NULL;
-		g_autoptr(GsApp) runtime_new = NULL;
+	if (gs_app_get_kind (app) != AS_APP_KIND_DESKTOP)
+		return TRUE;
 
-		data = gs_flatpak_fetch_remote_metadata (self, app, cancellable, error);
-		if (data == NULL) {
-			gs_utils_error_add_unique_id (error, app);
-			return FALSE;
-		}
-
-		str = g_bytes_get_data (data, &len);
-		runtime_new = gs_flatpak_create_runtime_from_metadata (self, app,
-								       str, len,
-								       error);
-		if (runtime_new == NULL)
-			return FALSE;
-		gs_app_set_update_runtime (app, runtime_new);
+	/* ensure that we get the right runtime that will need to be installed */
+	data = gs_flatpak_fetch_remote_metadata (self, app, cancellable, error);
+	if (data == NULL) {
+		gs_utils_error_add_unique_id (error, app);
+		return FALSE;
 	}
 
-	/* no runtime required */
-	runtime = gs_app_get_update_runtime (app);
+	str = g_bytes_get_data (data, &len);
+	runtime = gs_flatpak_create_runtime_from_metadata (self, app,
+							   str, len,
+							   error);
+
+	/* apps need to have a runtime */
 	if (runtime == NULL)
-		return TRUE;
+		return FALSE;
+
+	gs_app_set_update_runtime (app, runtime);
 
 	/* the runtime could come from a different remote to the app */
 	if (!gs_refine_item_metadata (self, runtime, cancellable, error)) {
@@ -2617,8 +2614,6 @@ gs_flatpak_refine_runtime_for_install (GsFlatpak *self,
 		gs_utils_error_add_unique_id (error, runtime);
 		return FALSE;
 	}
-
-	gs_app_set_runtime (app, runtime);
 
 	return TRUE;
 }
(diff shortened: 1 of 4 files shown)


plugins/flatpak/gs-self-test.c
```diff
@@ -1398,6 +1399,11 @@ gs_plugins_flatpak_app_update_func (GsPluginLoader *plugin_loader)
 				  G_CALLBACK (update_app_progress_notify_cb),
 				  &progress_cnt);
 
+	/* check that the runtime is not the update's one */
```
mariospr: I'd personally add a new test (e.g. `gs_plugins_flatpak_app_update_runtime_changed_func` or similar) instead of adapting the currently existing test case, since you're testing different use cases, and both should be checked.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2567,34 +2567,31 @@ gs_flatpak_refine_runtime_for_install (GsFlatpak *self,
...
 
+	str = g_bytes_get_data (data, &len);
+	runtime = gs_flatpak_create_runtime_from_metadata (self, app,
+							   str, len,
+							   error);
+
 	/* no runtime required */
-	runtime = gs_app_get_update_runtime (app);
 	if (runtime == NULL)
 		return TRUE;
```
mariospr: Shouldn't this return `FALSE` now, as in https://github.com/endlessm/gnome-software/pull/274/commits/d7073b4e11d47e0f15da13076fbddc0fa1110862#diff-e64404ac7c111b350b85abaa4fb8084dL2589 ? It seems odd to me to return `TRUE` here if the call to `gs_flatpak_create_runtime_from_metadata()` failed (seemed fine with the old call to `gs_app_get_update_runtime()`, though)

joaquimrocha: Nope. The value is right because in certain functions in GNOME Software's plugins, the `TRUE` value is also for when we want to ignore/skip. As you can see, we are also returning `TRUE` in:
```
if (gs_app_get_kind (app) != AS_APP_KIND_DESKTOP)
		return TRUE;
```
So in this case, just like before, if the app in question has no runtime, we return `TRUE`.
I know this is counterintuitive but I prefer if we go with this, as a proper fix (let libflatpak care about installing the runtime itself) will anyhow involve new API that we're planning and thus IMO we shouldn't make deep changes in how g-s is working ATM.

plugins/flatpak/gs-self-test.c
joaquimrocha: The tests are actually not trivial to implement currently, so I preferred to update the current test, as also the difference won't make any of previous tests in this function invalid.
It's a complex update and I think it's fine to test it in the current update test.

plugins/flatpak/tests/app-update/org.test.Chiron/metadata
```diff
@@ -1,4 +1,4 @@
 [Application]
 name=org.test.Chiron
-runtime=org.test.Runtime/x86_64/master
+runtime=org.test.Runtime/x86_64/new_master
```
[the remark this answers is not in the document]
joaquimrocha: Nothing has changed in the previous use case, because the error we're fixing of course that it wouldn't care about the runtime in the update.
Changing the branch, as we do often in Endless is actually a better test because we don't give a chance for e.g. checking just the ID instead of the unique ID of apps.

plugins/flatpak/tests/build.py
```diff
@@ -80,7 +81,11 @@ def copy_repo(srcdir, destdir):
               'app-missing-runtime/repo')
 
 # app with an update
-copy_repo('app-with-runtime', 'app-update')
+build_flatpak('org.test.Runtime',
```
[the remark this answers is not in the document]
joaquimrocha: And this should be fine because it just makes the update more complex (not too much though) without compromising the previous tests.
We can (and should) add more tests that do test every single combination (same runtime when updating / new runtime when updating / deleting the runtime right before the update / etc.) but tests are very time taking as they're implemented right now, so I'd prefer to add this now and rework the tests when we have an easier solution to use.

plugins/flatpak/gs-flatpak.c
mariospr: > So in this case, just like before, if the app in question has no runtime, we return TRUE.
I know this is counterintuitive but I prefer if we go with this, as a proper fix (let libflatpak care about installing the runtime itself) will anyhow involve new API that we're planning and thus IMO we shouldn't make deep changes in how g-s is working ATM.

Yes, I get that. But in this case what I'm seeing is that you're returning TRUE when you **failed to create** a runtime for the app, not when a call to `gs_app_get_update_runtime()` has failed because there's no associated runtime, that's what I thought returning FALSE (as in other cases of failure before that point) made sense.

That said, I'm not extremely familiar with this code, so you probably know better and if you still think it's correct, I guess we can merge it. But first, can you confirm that?

plugins/flatpak/gs-self-test.c
mariospr: The difference is that you're testing different things: before you were testing updates that didn't require a different runtime, now you're testing updates that require a different runtime. I'm all for adding new tests everywhere, but not at the expense of stopping to test things that were tested before.

plugins/flatpak/tests/build.py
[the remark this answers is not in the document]
mariospr: As I said before, I don't think this particular thing should stop this PR, but in the future I think it would be better if we add new tests instead of replacing things :-)

plugins/flatpak/gs-flatpak.c
joaquimrocha: Ah, okay, I see it now. We should definitely do more than returning FALSE, I will need to check the implications of returning FALSE though. I will take a look. Thanks.

PR #276, opened 2017-10-30
eos: Refactor how the shortcuts are handled

joaquimrocha:
The EOS plugin needs to keep track of which apps started or stopped
having shortcuts in the desktop, in order to apply the right quirks to
the app objects.
However, the way this was being done was confusing (it was cleaning the
list of shortcuts on every refine) and had problems like: when the
desktop shortcut was added, removed, and re-added again, then the
correct quirk were not set because the list of apps with shortcuts
was not being correctly updated (i.e. the correct shortcut button would
not be shown).

These changes refactor the way the shortcuts are handled. Now EOS keeps
the list of shortcuts up to date, from the beginning, and every time
a shortcut is changed, which simplifies the code. The mentioned mistake
in updating the shortcut quirk is fixed as well.

https://phabricator.endlessm.com/T19698

### DIFF
--- a/plugins/eos/gs-plugin-eos.c
+++ b/plugins/eos/gs-plugin-eos.c
@@ -110,7 +110,7 @@ on_desktop_apps_changed (GDBusConnection *connection,
 			 GVariant	 *parameters,
 			 GsPlugin	 *plugin)
 {
-	GHashTable *apps;
+	g_autoptr(GHashTable) apps = NULL;
 	GHashTableIter iter;
 	gpointer key, value;
 	GsPluginData *priv = gs_plugin_get_data (plugin);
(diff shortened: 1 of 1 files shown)


plugins/eos/gs-plugin-eos.c
```diff
@@ -123,28 +123,35 @@ on_desktop_apps_changed (GDBusConnection *connection,
...
 	g_hash_table_iter_init (&iter, priv->desktop_apps);
 	while (g_hash_table_iter_next (&iter, &key, &value)) {
-		GsApp *app = gs_plugin_cache_lookup (plugin, key);
+		GsApp *app = NULL;
 
-		if (!g_hash_table_lookup (apps, key)) {
-			if (app)
-				gs_app_remove_quirk (app,
-						     AS_APP_QUIRK_HAS_SHORTCUT);
+		if (g_hash_table_lookup (apps, key)) {
```
cosimoc: You could just do
```
if (g_hash_table_remove (apps, key))
  continue;
```

plugins/eos/gs-plugin-eos.c
```diff
@@ -123,28 +123,35 @@ on_desktop_apps_changed (GDBusConnection *connection,
...
-				gs_app_add_quirk (app,
-						  AS_APP_QUIRK_HAS_SHORTCUT);
+		app = gs_plugin_cache_lookup (plugin, key);
+		if (app)
+			gs_app_remove_quirk (app, AS_APP_QUIRK_HAS_SHORTCUT);
 
-			g_hash_table_add (priv->desktop_apps,
-					  g_strdup (key));
-		}
+		g_hash_table_remove (priv->desktop_apps, key);
```
cosimoc: I think the correct way of doing this is `g_hash_table_iter_remove()`

plugins/eos/gs-plugin-eos.c
```diff
@@ -277,8 +294,7 @@ gs_plugin_initialize (GsPlugin *plugin)
 
 	priv->session_bus = g_application_get_dbus_connection (app);
 
-	priv->desktop_apps = g_hash_table_new_full (g_str_hash, g_str_equal,
-						    g_free, NULL);
+	priv->desktop_apps = NULL;
```
cosimoc: There should not be any need to initialize this explicitly.

PR #279, opened 2017-11-02
eos: Add com.endlessm.EknServices2 to proxy app

ptomato:
All of the possible versions of EknServices should be hidden by the
"Endless Platform" proxy app.

https://phabricator.endlessm.com/T19806

### DIFF
--- a/plugins/flatpak/gs-flatpak.c
+++ b/plugins/flatpak/gs-flatpak.c
@@ -1868,7 +1868,7 @@ gs_plugin_refine_item_state (GsFlatpak *self,
 }
 
 static GsApp *
-gs_flatpak_create_runtime (GsPlugin *plugin, GsApp *parent, const gchar *runtime)
+gs_flatpak_create_runtime (GsPlugin *plugin, const gchar *runtime)
 {
 	g_autofree gchar *source = NULL;
 	g_auto(GStrv) split = NULL;
@@ -1917,7 +1917,6 @@ gs_flatpak_create_runtime (GsPlugin *plugin, GsApp *parent, const gchar *runtime
 
 static GsApp *
 gs_flatpak_create_runtime_from_metadata (GsFlatpak *self,
-					 const GsApp *app,
 					 const gchar *data,
 					 const gsize length,
 					 GError **error)
@@ -1935,7 +1934,7 @@ gs_flatpak_create_runtime_from_metadata (GsFlatpak *self,
 		gs_utils_error_convert_gio (error);
 		return NULL;
 	}
-	return gs_flatpak_create_runtime (self->plugin, app, runtime);
+	return gs_flatpak_create_runtime (self->plugin, runtime);
 }
 
 static gboolean
@@ -2000,7 +1999,7 @@ gs_flatpak_set_app_metadata (GsFlatpak *self,
 
 	/* create runtime */
 	if (gs_app_get_runtime (app) == NULL) {
-		app_runtime = gs_flatpak_create_runtime (self->plugin, app, runtime);
+		app_runtime = gs_flatpak_create_runtime (self->plugin, runtime);
 		if (app_runtime != NULL) {
 			gs_plugin_refine_item_scope (self, app_runtime);
 			gs_app_set_runtime (app, app_runtime);
@@ -2582,7 +2581,7 @@ gs_flatpak_refine_runtime_for_install (GsFlatpak *self,
 	}
 
 	str = g_bytes_get_data (data, &len);
-	runtime = gs_flatpak_create_runtime_from_metadata (self, app,
+	runtime = gs_flatpak_create_runtime_from_metadata (self,
 							   str, len,
 							   error);
 
--- a/plugins/flatpak/gs-flatpak.c
+++ b/plugins/flatpak/gs-flatpak.c
@@ -2477,6 +2477,71 @@ gs_flatpak_app_remove_source (GsFlatpak *self,
 	return TRUE;
 }
 
+/*
+ * gs_flatpak_get_services_app_for_runtime:
+ * Specific to Endless runtimes, gets a GsApp reference for the related
+ * "EknServices" app that provides system services for apps that use the
+ * given runtime.
+ *
+ * Returns NULL if no services app was required or there was an error.
+ * Does not take a GError since any error should be recoverable. However, in
+ * order to make sure the operation was not cancelled, you should check the
+ * status of the GCancellable after calling this.
+ */
+static GsApp *
+gs_flatpak_get_services_app_for_runtime (GsFlatpak *self, GsApp *runtime,
+                                         GCancellable *cancellable)
+{
+	const gchar *runtime_id;
+	const gchar *runtime_branch;
+	const gchar *services_branch;
+	const gchar *arch;
+	g_autofree gchar *services_id = NULL;
+	g_autofree gchar *description = NULL;
+	g_autoptr(FlatpakRef) services_ref = NULL;
+	g_autoptr(GError) error = NULL;
+	g_autoptr(GsApp) services_app = NULL;
+
+	runtime_id = gs_app_get_id (runtime);
+	runtime_branch = gs_app_get_branch (runtime);
+
+	if ((g_strcmp0 (runtime_id, "com.endlessm.Platform") == 0 &&
+	     g_strcmp0 (runtime_branch, "eos3.1") == 0) ||
+	    (g_strcmp0 (runtime_id, "com.endlessm.apps.Platform") == 0 &&
+	     g_strcmp0 (runtime_branch, "1") == 0)) {
+		services_id = g_strdup ("com.endlessm.EknServices");
+		services_branch = "eos3";
+	} else if (g_strcmp0 (runtime_id, "com.endlessm.apps.Platform") == 0) {
+		if (g_strcmp0 (runtime_branch, "master") == 0)
+			services_id = g_strdup ("com.endlessm.EknServices2");
+		else
+			services_id = g_strdup_printf ("com.endlessm.EknServices%s",
+			                               runtime_branch);
+		services_branch = "stable";
+	} else {
+		/* Runtime doesn't require an EknServices app */
+		return NULL;
+	}
+
+	/* Construct a GsApp for the EknServices we determined we needed */
+	arch = gs_flatpak_app_get_ref_arch (runtime);
+	description = g_strdup_printf ("app/%s/%s/%s", services_id, arch, services_branch);
+	services_ref = flatpak_ref_parse (description, &error);
+	if (services_ref == NULL) {
+		g_critical ("Invalid flatpak ref description %s: %s", description,
+		            error->message);
+		return NULL;
+	}
+
+	services_app = gs_flatpak_create_app (self, services_ref);
+	if (!gs_plugin_refine_item_origin (self, services_app, cancellable, &error)) {
+		g_warning ("Couldn't discover origin for %s: %s", description,
+		           error->message);
+		return NULL;
+	}
+	return g_steal_pointer (&services_app);
+}
+
 static GsAppList *
 gs_flatpak_get_list_for_remove (GsFlatpak *self, GsApp *app,
 				GCancellable *cancellable, GError **error)
@@ -2617,6 +2682,23 @@ gs_flatpak_refine_runtime_for_install (GsFlatpak *self,
 	return TRUE;
 }
 
+static void
+gs_flatpak_add_app_to_list_if_not_installed (GsApp *app, GsAppList *list,
+					     const GHashTable *hash_installed)
+{
+	g_autofree gchar *ref_display = gs_flatpak_app_get_ref_display (app);
+	if (g_hash_table_contains (hash_installed, ref_display)) {
+		g_debug ("%s is already installed, so skipping",
+			 gs_app_get_id (app));
+		return;
+	}
+
+	g_debug ("%s/%s is not already installed, so installing",
+		 gs_flatpak_app_get_ref_name (app),
+		 gs_flatpak_app_get_ref_branch (app));
+	gs_app_list_add (list, app);
+}
+
 static GsAppList *
 gs_flatpak_get_list_for_install (GsFlatpak *self, GsApp *app,
 				 GCancellable *cancellable, GError **error)
@@ -2627,6 +2709,8 @@ gs_flatpak_get_list_for_install (GsFlatpak *self, GsApp *app,
 	g_autoptr(GPtrArray) xrefs_installed = NULL;
 	g_autoptr(GHashTable) hash_installed = NULL;
 	g_autoptr(GsAppList) list = gs_app_list_new ();
+	g_autoptr(GsApp) services_app = NULL;
+	g_autoptr(GsApp) services_runtime = NULL;
 
 	/* get the list of installed apps */
 	xrefs_installed = flatpak_installation_list_installed_refs (self->installation,
@@ -2647,18 +2731,22 @@ gs_flatpak_get_list_for_install (GsFlatpak *self, GsApp *app,
 	if (!gs_flatpak_refine_runtime_for_install (self, app, cancellable, error))
 		return FALSE;
 	runtime = gs_app_get_update_runtime (app);
-	if (runtime != NULL) {
-		g_autofree gchar *ref_display = NULL;
-		ref_display = gs_flatpak_app_get_ref_display (runtime);
-		if (g_hash_table_contains (hash_installed, ref_display)) {
-			g_debug ("%s is already installed, so skipping",
-				 gs_app_get_id (runtime));
-		} else {
-			g_debug ("%s/%s is not already installed, so installing",
-				 gs_flatpak_app_get_ref_name (runtime),
-				 gs_flatpak_app_get_ref_branch (runtime));
-			gs_app_list_add (list, runtime);
-		}
+	if (runtime != NULL)
+		gs_flatpak_add_app_to_list_if_not_installed (runtime, list, hash_installed);
+
+	/* add services flatpak */
+	services_app = gs_flatpak_get_services_app_for_runtime (self, runtime, cancellable);
+	if (g_cancellable_set_error_if_cancelled (cancellable, error))
+		return FALSE;
+	if (services_app != NULL) {
+		gs_flatpak_add_app_to_list_if_not_installed (services_app, list, hash_installed);
+
+		/* add services flatpak's runtime, if different from app's runtime */
+		if (!gs_flatpak_refine_runtime_for_install (self, services_app, cancellable, error))
+			return FALSE;
+		services_runtime = gs_app_get_update_runtime (services_app);
+		if (services_runtime != NULL)
+			gs_flatpak_add_app_to_list_if_not_installed (services_runtime, list, hash_installed);
 	}
 
 	/* lookup any related refs for this ref */
--- a/plugins/eos/gs-plugin-eos.c
+++ b/plugins/eos/gs-plugin-eos.c
@@ -1318,6 +1318,7 @@ add_updates (GsPlugin *plugin,
 	const char *proxied_apps[] = {"com.endlessm.Platform",
 				      "com.endlessm.apps.Platform",
 				      "com.endlessm.EknServices.desktop",
+				      "com.endlessm.EknServices2.desktop",
 				      "com.endlessm.quote_of_the_day.en.desktop",
 				      "com.endlessm.word_of_the_day.en.desktop",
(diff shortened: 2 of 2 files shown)


plugins/flatpak/gs-flatpak.c
```diff
@@ -1868,25 +1868,18 @@ gs_plugin_refine_item_state (GsFlatpak *self,
 }
 
 static GsApp *
-gs_flatpak_create_runtime (GsPlugin *plugin, GsApp *parent, const gchar *runtime)
+gs_flatpak_create_app_from_info (GsPlugin *plugin, const gchar *source,
```
cosimoc: Would be nice to split this refactor into a separate commit

plugins/flatpak/gs-flatpak.c
```diff
@@ -2618,11 +2629,60 @@ gs_flatpak_refine_runtime_for_install (GsFlatpak *self,
...
+	g_autofree gchar *source = NULL;
+
+	runtime_id = gs_app_get_id (runtime);
+	runtime_branch = gs_app_get_branch (runtime);
+
+	if ((g_strcmp0 (runtime_id, "com.endlessm.Platform") == 0 &&
+	     g_strcmp0 (runtime_branch, "eos3.1") == 0) ||
+	    (g_strcmp0 (runtime_id, "com.endlessm.apps.Platform") == 0 &&
+	     g_strcmp0 (runtime_branch, "1") == 0)) {
+		services_id = g_strdup("com.endlessm.EknServices");
```
cosimoc: Missing space before paren

plugins/flatpak/gs-flatpak.c
```diff
@@ -2618,11 +2629,60 @@ gs_flatpak_refine_runtime_for_install (GsFlatpak *self,
...
+		services_branch = "eos3";
+	} else if (g_strcmp0 (runtime_id, "com.endlessm.apps.Platform") == 0) {
+		services_id = g_strdup_printf("com.endlessm.EknServices%s", runtime_branch);
```
cosimoc: Missing space before paren

plugins/flatpak/gs-flatpak.c
```diff
@@ -2567,6 +2567,72 @@ gs_flatpak_related_should_download (GsFlatpak *self, GsApp *app, FlatpakRelatedR
...
+				           flatpak_ref_get_name (FLATPAK_REF (xref)),
+				           error->message);
+				g_clear_error (&error);
+				continue;
+			}
+
+			/* Special case for EknServices 1 and eosplatform 3.1 */
+			if (g_strcmp0 (services_name_ver, "1") == 0 &&
+			    g_strcmp0 (gs_app_get_id (app), "com.endlessm.Platform") == 0 &&
+			    g_strcmp0 (gs_app_get_branch (app), "eos3.1")) {
```
cosimoc: Should be `g_strcmp0 (...) == 0`, no? (If not, prefer the explicit `!= 0` notation.)

plugins/flatpak/gs-flatpak.c
```diff
@@ -2618,11 +2629,60 @@ gs_flatpak_refine_runtime_for_install (GsFlatpak *self,
...
+{
+	const gchar *runtime_id, *runtime_branch, *services_branch, *arch;
+	g_autofree gchar *services_id = NULL;
+	g_autofree gchar *source = NULL;
+
+	runtime_id = gs_app_get_id (runtime);
+	runtime_branch = gs_app_get_branch (runtime);
+
+	if ((g_strcmp0 (runtime_id, "com.endlessm.Platform") == 0 &&
+	     g_strcmp0 (runtime_branch, "eos3.1") == 0) ||
```
joaquimrocha: Shouldn't it be also for _eos3.2_?

plugins/flatpak/gs-flatpak.c
```diff
@@ -2618,11 +2629,60 @@ gs_flatpak_refine_runtime_for_install (GsFlatpak *self,
...
+		services_branch = "stable";
+	} else {
+		/* Runtime doesn't require an EknServices app */
+		return NULL;
+	}
+
+	/* Construct a GsApp for the EknServices we determined we needed */
+	arch = gs_flatpak_app_get_ref_arch (runtime);
+	source = g_strdup_printf ("app/%s/%s/%s", services_id, arch, services_branch);
+	return gs_flatpak_create_app_from_info (plugin, source, services_id, arch,
```
joaquimrocha: I don't think the app should be created this way. You should rather use `gs_flatpak_create_app` that takes a `FlatpakRef`. That way you also don't have to split up the runtime creation function into yet another app creation one.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2618,11 +2629,60 @@ gs_flatpak_refine_runtime_for_install (GsFlatpak *self,
...
+	                                        services_branch, AS_APP_KIND_GENERIC,
+	                                        FLATPAK_REF_KIND_APP);
+}
+
+static void
+gs_flatpak_add_app_to_list_if_not_installed (GsApp *app, GsAppList *list,
+					     const GHashTable *hash_installed)
+{
+	g_autofree gchar *ref_display = NULL;
+	ref_display = gs_flatpak_app_get_ref_display (app);
```
joaquimrocha: You could make these into the same line where you declare it.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2648,19 +2708,13 @@ gs_flatpak_get_list_for_install (GsFlatpak *self, GsApp *app,
...
-			gs_app_list_add (list, runtime);
-		}
-	}
+	if (runtime != NULL)
+		gs_flatpak_add_app_to_list_if_not_installed (runtime, list, hash_installed);
+
+	/* add services flatpak */
+	services = gs_flatpak_get_services_app_for_runtime (self->plugin, runtime);
+	if (services != NULL)
+		gs_flatpak_add_app_to_list_if_not_installed (services, list, hash_installed);
```
joaquimrocha: I understand the consistence with the code above and not repeating two blocks. However, this involves more changes to `gs-flatpak` than I really wanted. In this case, the code you're abstracting `gs_flatpak_add_app_to_list_if_not_installed` is only worth it because of the comments, but the comments in the function look out of place to me (they're referring to the installation). So just repeat the code, or maybe do use the function but return a boolean and use it to print the debug messages from outside.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2567,6 +2567,72 @@ gs_flatpak_related_should_download (GsFlatpak *self, GsApp *app, FlatpakRelatedR
...
+
+			/* FIXME: can you compare pointer equality like this
+			 * because of the cache? */
+			if (runtime == app)
+				found_app_that_uses_runtime = TRUE;
+		}
+
+		return found_app_that_uses_runtime;
+	}
+
```
joaquimrocha: I am afraid this will have an impact in performance: 1) Listing all installed refs with libflatpak is slow; 2) IO per each of those apps ditto.
I know this should happen only during installation/update, but still this code seems like a lot of complexity added "just" for skipping EknServices.

I understand that keeping a number of EknServices around may have an impact on the search, but I would rather include it's clean up as part of cleaning up unneeded legacy runtimes. So when we get rid of the eos3.2 Platform runtime, we should also uninstall EknServices (see `...get_list_for_remove`).

plugins/flatpak/gs-flatpak.c
ptomato: I don't think that's necessary, there are no knowledge apps published that use the `com.endlessm.Platform//eos3.2` runtime (and never will be.)

plugins/flatpak/gs-flatpak.c
ptomato: OK, I agree with what you say and will change it.

However, the situation that I was trying to avoid was this:
- user doesn't have any SDK2 apps installed
- user updates Endless Platform
- com.endlessm.EknServices2 is in the list
- com.endlessm.EknServices2 gets installed as part of the Endless Platform update
- com.endlessm.EknServices2 pulls in SDK2 into the update

Do you know if that situation will actually happen, and/or what's a better way to avoid it?

plugins/flatpak/gs-flatpak.c
ptomato: I'm not sure I understand what you're asking; I can't figure out which comments you're referring to.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2500,6 +2539,15 @@ gs_flatpak_get_list_for_remove (GsFlatpak *self, GsApp *app,
 		return FALSE;
 	}
 
+	/* If this is the SDK runtime, remove the associated EknServices.
+	 * Ignores EknServices associated with com.endlessm.Platform//eos3.1. */
+	if (g_strcmp0 (ref_name, "com.endlessm.apps.Platform") == 0) {
+		g_autoptr(GsApp) services = gs_flatpak_get_services_app_for_runtime (self, app);
+		if (services != NULL) {
```
cosimoc: Please remove curly braces for single-line blocks

plugins/flatpak/gs-flatpak.c
```diff
@@ -2647,19 +2711,13 @@ gs_flatpak_get_list_for_install (GsFlatpak *self, GsApp *app,
...
-				 gs_flatpak_app_get_ref_name (runtime),
-				 gs_flatpak_app_get_ref_branch (runtime));
-			gs_app_list_add (list, runtime);
-		}
-	}
+	if (runtime != NULL)
+		gs_flatpak_add_app_to_list_if_not_installed (runtime, list, hash_installed);
+
+	/* add services flatpak */
+	services = gs_flatpak_get_services_app_for_runtime (self, runtime);
```
cosimoc: Does `services` (or even `runtime`) not get leaked?

plugins/flatpak/gs-flatpak.c
```diff
@@ -2477,18 +2477,57 @@ gs_flatpak_app_remove_source (GsFlatpak *self,
...
+	g_autofree gchar *services_id = NULL;
+	g_autofree gchar *description = NULL;
+	g_autoptr(FlatpakRef) services_ref = NULL;
+	g_autoptr(GError) error = NULL;
+
+	runtime_id = gs_app_get_id (runtime);
+	runtime_branch = gs_app_get_branch (runtime);
+
+	if ((g_strcmp0 (runtime_id, "com.endlessm.Platform") == 0 &&
+	     g_strcmp0 (runtime_branch, "eos3.1") == 0) ||
```
cosimoc: Does `com.endlessm.Platform` version `eos3.2` not require `EknServices` version `eos3`?

ptomato: No, there are no knowledge apps that use `com.endlessm.Platform//eos3.2` and never will be.

plugins/flatpak/gs-flatpak.c
ptomato: I don't think so, I believe the `GsAppList` takes ownership.

plugins/flatpak/gs-flatpak.c
cosimoc: Sorry, did not see the previous similar comment from @joaquimrocha. Sounds good then.

plugins/flatpak/gs-flatpak.c
cosimoc: `gs_app_list_add()` adds a new ref as far as I can see.

ptomato: `runtime` is not leaked (comes from a transfer none function), but the `services` leak is fixed now.

cosimoc:
LGTM now but I will leave unmerged for @joaquimrocha to take another look too.

ptomato:
I would also prefer not to merge this until I've figured out how to test it.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2617,6 +2664,22 @@ gs_flatpak_refine_runtime_for_install (GsFlatpak *self,
...
 }
 
+static void
+gs_flatpak_add_app_to_list_if_not_installed (GsApp *app, GsAppList *list,
+					     const GHashTable *hash_installed)
+{
+	g_autofree gchar *ref_display = gs_flatpak_app_get_ref_display (app);
+	if (g_hash_table_contains (hash_installed, ref_display)) {
+		g_debug ("%s is already installed, so skipping",
+			 gs_app_get_id (app));
```
joaquimrocha: You could do an early return here to skip the else block.

plugins/flatpak/gs-flatpak.c
ptomato: OK, I tried repeating the code but it looks wrong to me... I still get the feeling I am misinterpreting your comment and looking at the wrong thing, but if you don't feel strongly about it then we can just drop it.

plugins/flatpak/gs-flatpak.c
ptomato: Fixed.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2477,18 +2477,82 @@ gs_flatpak_app_remove_source (GsFlatpak *self,
...
+	description = g_strdup_printf ("app/%s/%s/%s", services_id, arch, services_branch);
+	services_ref = flatpak_ref_parse (description, &error);
+	if (!services_ref) {
+		g_critical ("Invalid flatpak ref description %s: %s", description,
+		            error->message);
+		return NULL;
+	}
+
+	retval = gs_flatpak_create_app (self, services_ref);
+	gs_app_set_origin (retval, gs_app_get_origin (runtime));
```
joaquimrocha: If you do this then you'll make it look for the services app in the runtime's remote, and this is why you're getting the failure you've reported in [here](https://phabricator.endlessm.com/T19806#504381). Also, by setting the origin (and if the origin is not _NULL_), the next line will do nothing as it first verifies if an origin is already set before it proceeds to look for it.
So maybe all that's needed is to delete this line.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2477,18 +2477,82 @@ gs_flatpak_app_remove_source (GsFlatpak *self,
...
+static GsApp *
+gs_flatpak_get_services_app_for_runtime (GsFlatpak *self, GsApp *runtime,
+                                         GCancellable *cancellable)
+{
+	const gchar *runtime_id, *runtime_branch, *services_branch, *arch;
+	g_autofree gchar *services_id = NULL;
+	g_autofree gchar *description = NULL;
+	g_autoptr(FlatpakRef) services_ref = NULL;
+	g_autoptr(GError) error = NULL;
+	g_autoptr(GsApp) retval = NULL;
```
joaquimrocha: Let's call this `app` or `services_app` instead.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2477,18 +2477,82 @@ gs_flatpak_app_remove_source (GsFlatpak *self,
...
+		/* Runtime doesn't require an EknServices app */
+		return NULL;
+	}
+
+	/* Construct a GsApp for the EknServices we determined we needed */
+	arch = gs_flatpak_app_get_ref_arch (runtime);
+	description = g_strdup_printf ("app/%s/%s/%s", services_id, arch, services_branch);
+	services_ref = flatpak_ref_parse (description, &error);
+	if (!services_ref) {
+		g_critical ("Invalid flatpak ref description %s: %s", description,
```
joaquimrocha: I am afraid that due to a mistake we get it to abort here, rather than "simply" not getting the services app installed.
So I'd make this a warning instead... I know the way I am proposing makes it harder to notice when this happens but still, having no _EknServices_ and having users reporting that they cannot search in apps seems better than eventually braking the installation of apps.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2477,18 +2477,82 @@ gs_flatpak_app_remove_source (GsFlatpak *self,
...
+	} else {
+		/* Runtime doesn't require an EknServices app */
+		return NULL;
+	}
+
+	/* Construct a GsApp for the EknServices we determined we needed */
+	arch = gs_flatpak_app_get_ref_arch (runtime);
+	description = g_strdup_printf ("app/%s/%s/%s", services_id, arch, services_branch);
+	services_ref = flatpak_ref_parse (description, &error);
+	if (!services_ref) {
```
joaquimrocha: Richard doesn't like the `!pointer` notation (prefers `!= NULL`). So even though I prefer to do it as you did, I try to keep the coding style in core modules close to upstream (even if this will never be merged). Small nitpick though.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2477,18 +2477,82 @@ gs_flatpak_app_remove_source (GsFlatpak *self,
...
+ * Returns NULL if no services app was required or there was an error.
+ * Does not take a GError since any error should be recoverable. However, in
+ * order to make sure the operation was not cancelled, you should check the
+ * status of the GCancellable after calling this.
+ */
+static GsApp *
+gs_flatpak_get_services_app_for_runtime (GsFlatpak *self, GsApp *runtime,
+                                         GCancellable *cancellable)
+{
+	const gchar *runtime_id, *runtime_branch, *services_branch, *arch;
```
joaquimrocha: Usually we don't have declarations like this in g-s. Even if I prefer these actually.
I am also not sure if the C99 standard will complain or not, but I'd make these several lines. (Just a nitpick though, not a blocker for me, unless there's a compiler warning)

plugins/flatpak/gs-flatpak.c
```diff
@@ -2500,6 +2564,19 @@ gs_flatpak_get_list_for_remove (GsFlatpak *self, GsApp *app,
 		return FALSE;
 	}
 
+	/* If this is the SDK runtime, remove the associated EknServices.
+	 * Ignores EknServices associated with com.endlessm.Platform//eos3.1. */
+	if (g_strcmp0 (ref_name, "com.endlessm.apps.Platform") == 0) {
```
joaquimrocha: As you pointed out recently, since we never remove a runtime from g-s. Even if we have to do so in the future, it's probably a good idea not to carry this code for now (because e.g. if it gets outdated, then when we do implement the removal of runtimes we could forget to update or verify this as long as it compiles...).

plugins/flatpak/gs-flatpak.c
```diff
@@ -2647,19 +2742,18 @@ gs_flatpak_get_list_for_install (GsFlatpak *self, GsApp *app,
...
-				 gs_flatpak_app_get_ref_name (runtime),
-				 gs_flatpak_app_get_ref_branch (runtime));
-			gs_app_list_add (list, runtime);
-		}
+	if (runtime != NULL)
+		gs_flatpak_add_app_to_list_if_not_installed (runtime, list, hash_installed);
+
+	/* add services flatpak */
+	services = gs_flatpak_get_services_app_for_runtime (self, runtime, cancellable);
+	if (g_cancellable_is_cancelled (cancellable)) {
```
joaquimrocha: You can use `g_cancellable_set_error_if_cancelled` which sets the error and returns `TRUE` if it's been cancelled. So all you'd need to do is return FALSE in that case.

plugins/flatpak/gs-flatpak.c
ptomato: OMG, I stared at this for so long and didn't notice that I had failed to remove this line! Thanks!

plugins/flatpak/gs-flatpak.c
ptomato: Multiple declarations per line have been a thing in C since at least [1978](https://archive.org/details/TheCProgrammingLanguageFirstEdition) :smile: But I agree it's better to keep the style of the surrounding code.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2647,19 +2731,22 @@ gs_flatpak_get_list_for_install (GsFlatpak *self, GsApp *app,
...
+
+	/* add services flatpak */
+	services_app = gs_flatpak_get_services_app_for_runtime (self, runtime, cancellable);
+	if (g_cancellable_set_error_if_cancelled (cancellable, error))
+		return FALSE;
+	if (services_app != NULL)
+		gs_flatpak_add_app_to_list_if_not_installed (services_app, list, hash_installed);
+
+	/* add services flatpak's runtime, if different from app's runtime */
+	if (!gs_flatpak_refine_runtime_for_install (self, services_app, cancellable, error))
```
joaquimrocha: I think we will have problems here (and below) when `services_app` is NULL. This should be part of the `if` block above.

ptomato:
Fixed the last comment.

PR #282, opened 2017-11-03
eos: Block coding chatbox if the coding feature is disabled

rshuler:
This app is only intended for use with the coding feature
and is not yet ready for general use.

https://phabricator.endlessm.com/T19819

### DIFF
--- a/plugins/eos/gs-plugin-eos.c
+++ b/plugins/eos/gs-plugin-eos.c
@@ -64,6 +64,7 @@ struct GsPluginData
 	int applications_changed_id;
 	SoupSession *soup_session;
 	char *personality;
+	gboolean is_coding_enabled;
 	char *os_version_id;
 	gboolean eos_arch_is_arm;
 };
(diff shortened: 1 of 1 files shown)


plugins/eos/gs-plugin-eos.c
```diff
@@ -744,6 +766,19 @@ app_is_banned_for_personality (GsPlugin *plugin, GsApp *app)
...
+{
+	GsPluginData *priv = gs_plugin_get_data (plugin);
+	const char *id = gs_app_get_id (app);
+
+	/* the coding chatbox is in the core ostree, so unlike the
+	   personality-based blocking, we block even if it is installed */
+
+	return ((g_strcmp0 (priv->product, "coding") != 0) &&
+	        (g_strcmp0 (id, "com.endlessm.Coding.Chatbox.desktop") == 0));
+}
```
smspillaz: Interestingly enough, in the shell we blacklist based on the `enable-coding-game` key in `org.gnome.Shell`. Would it make more sense to blacklist based on that?

rshuler: Yes, agreed, and I have updated the PR.

PR #284, opened 2017-11-06
T19329

joaquimrocha:
https://phabricator.endlessm.com/T19329


plugins/flatpak/gs-self-test.c
```diff
@@ -1483,6 +1483,113 @@ gs_plugins_flatpak_app_update_func (GsPluginLoader *plugin_loader)
...
+					 NULL);
+	ret = gs_plugin_loader_job_action (plugin_loader, plugin_job, NULL, &error);
+	gs_test_flush_main_context ();
+	g_assert_no_error (error);
+	g_assert (ret);
+	g_assert_cmpint (gs_app_get_state (app), ==, AS_APP_STATE_INSTALLED);
+	g_assert_cmpstr (gs_app_get_version (app), ==, "1.2.3");
+
+	/* check if the extension was installed */
+        global_cache = gs_plugin_loader_get_global_cache (plugin_loader);
```
mariospr: Nit: wrong indentation

plugins/flatpak/gs-flatpak.c
```diff
@@ -2499,10 +2499,23 @@ gs_flatpak_get_list_for_remove (GsFlatpak *self, GsApp *app,
...
+			if (g_cancellable_is_cancelled (cancellable)) {
+				g_propagate_error (error, g_steal_pointer (&error_local));
+				return NULL;
+			}
+			g_warning ("Failed to refine %s when getting the list of apps to "
+				   "remove: %s; not adding the app...",
+				   gs_app_get_unique_id (app_tmp),
+				   error_local->message);
+			continue;
+		}
```
mariospr: It feels like this chunk of code could be refactored into a separate private function, and reuse it from `_get_list_for_install()`

mariospr:
Typo in the git commit message of `flatpak: Set apps as updatable instead of their extensions`: "...their ap should be shown..." -> "...they should be shown..."

plugins/flatpak/gs-flatpak-app.h
```diff
@@ -72,6 +72,9 @@ void			 gs_flatpak_app_set_file_kind		(GsApp		*app,
 								 GsFlatpakAppFileKind	file_kind);
 void			 gs_flatpak_app_set_runtime_repo	(GsApp		*app,
 								 GsApp		*runtime_repo);
+void			 gs_flatpak_app_set_main_app		(GsApp		*app,
+								 const gchar	*main_app_ref);
+const gchar		*gs_flatpak_app_get_main_app		(GsApp		*app);
```
mariospr: I find it a bit strange that this method is called `_get_main_app()` and it returns a `const char *` instead of a GsApp. IMHO, it should either return a `GsApp` or be renamed to something like `_get_main_app_name()`.

plugins/flatpak/gs-flatpak.c
```diff
@@ -1301,6 +1320,39 @@ gs_flatpak_app_install_source (GsFlatpak *self, GsApp *app,
...
+			 GsApp *related_app,
+			 GCancellable *cancellable,
+			 GError **error)
+{
+	g_autofree gchar *main_app_name = g_strdup (gs_flatpak_app_get_ref_name (related_app));
+	g_autoptr(FlatpakInstalledRef) ref = NULL;
+	const gchar *app_name;
+	g_auto(GStrv) app_tokens = NULL;
+
+        app_name = gs_flatpak_app_get_main_app (related_app);
```
mariospr: Wrong indentation here

plugins/flatpak/gs-flatpak.c
```diff
@@ -1301,6 +1320,39 @@ gs_flatpak_app_install_source (GsFlatpak *self, GsApp *app,
...
+	if (app_name == NULL)
+		return NULL;
+
+	app_tokens = g_strsplit (app_name, "/", -1);
+	if (g_strv_length (app_tokens) != 4)
```
mariospr: It looks like you should set the `GError` here, if this happens for any reason (e.g. maybe with `G_IO_ERROR_INVALID_DATA`)

plugins/flatpak/gs-flatpak.c
```diff
@@ -1301,6 +1320,39 @@ gs_flatpak_app_install_source (GsFlatpak *self, GsApp *app,
...
+			 GCancellable *cancellable,
+			 GError **error)
+{
+	g_autofree gchar *main_app_name = g_strdup (gs_flatpak_app_get_ref_name (related_app));
+	g_autoptr(FlatpakInstalledRef) ref = NULL;
+	const gchar *app_name;
+	g_auto(GStrv) app_tokens = NULL;
+
+        app_name = gs_flatpak_app_get_main_app (related_app);
+	if (app_name == NULL)
```
mariospr: It looks like you should set the GError here, if this happens for any reason (e.g. maybe with `G_IO_ERROR_NOT_FOUND`)

plugins/flatpak/gs-flatpak.c
```diff
@@ -1301,6 +1320,39 @@ gs_flatpak_app_install_source (GsFlatpak *self, GsApp *app,
...
+
+	ref = flatpak_installation_get_installed_ref (self->installation,
+						      FLATPAK_REF_KIND_APP,
+						      app_tokens[1],
+						      app_tokens[2],
+						      app_tokens[3],
+						      cancellable,
+						      error);
+
+	if (ref == NULL)
```
mariospr: Same comment, need to set GError

plugins/flatpak/gs-flatpak.c
```diff
@@ -1324,6 +1376,7 @@ gs_flatpak_add_updates (GsFlatpak *self, GsAppList *list,
 		const gchar *latest_commit;
 		g_autoptr(GsApp) app = NULL;
 		g_autoptr(GError) error_local = NULL;
+		GsApp *main_app = NULL;
```
mariospr: You're leaking `main_app` in this function

plugins/flatpak/gs-flatpak.c
```diff
@@ -1380,6 +1445,7 @@ gs_flatpak_add_updates_pending (GsFlatpak *self, GsAppList *list,
 		guint64 download_size = 0;
 		g_autoptr(GsApp) app = NULL;
 		g_autoptr(GError) error_local = NULL;
+		GsApp *main_app = NULL;
```
mariospr: You're leaking `main_app` in this function too

plugins/flatpak/gs-flatpak.c
```diff
@@ -1348,12 +1401,24 @@ gs_flatpak_add_updates (GsFlatpak *self, GsAppList *list,
...
-		gs_app_set_update_urgency (app, AS_URGENCY_KIND_UNKNOWN);
-		gs_app_set_size_download (app, 0);
-		gs_app_list_add (list, app);
+
+		if (gs_app_get_kind (app) == AS_APP_KIND_RUNTIME)
+			main_app = get_main_app_of_related (self, app, cancellable,
+							    &error_local);
+
+		if (main_app == NULL)
+			main_app = app;
```
mariospr: Maybe print `error_local->message` with a `g_warning()` here would be good.

plugins/flatpak/gs-flatpak.c
```diff
@@ -1388,10 +1454,22 @@ gs_flatpak_add_updates_pending (GsFlatpak *self, GsAppList *list,
...
 			continue;
 		}
-		gs_app_set_state (app, AS_APP_STATE_UPDATABLE_LIVE);
+
+		if (gs_app_get_kind (app) == AS_APP_KIND_RUNTIME)
+			main_app = get_main_app_of_related (self, app, cancellable,
+							    &error_local);
+
+		if (main_app == NULL)
+			main_app = app;
```
mariospr: Maybe print `error_local->message` with a `g_warning()` here would be good as well

plugins/flatpak/gs-self-test.c
```diff
@@ -1483,6 +1483,113 @@ gs_plugins_flatpak_app_update_func (GsPluginLoader *plugin_loader)
...
+	g_autoptr(GsAppList) list_updates = NULL;
+	g_autoptr(GsPluginJob) plugin_job = NULL;
+	g_autoptr(GMainLoop) loop = g_main_loop_new (NULL, FALSE);
+	GsAppList *global_cache = NULL;
+
+	/* drop all caches */
+	gs_plugin_loader_setup_again (plugin_loader);
+
+	/* no flatpak, abort */
+	if (!gs_plugin_loader_get_enabled (plugin_loader, "flatpak"))
```
cosimoc: You could call `g_assert_true()` here instead

plugins/flatpak/gs-self-test.c
```diff
@@ -1483,6 +1483,113 @@ gs_plugins_flatpak_app_update_func (GsPluginLoader *plugin_loader)
...
+	/* no files to use */
+	repodir1_fn = gs_test_get_filename (TESTDATADIR, "app-extension/repo");
+	if (repodir1_fn == NULL ||
+	    !g_file_test (repodir1_fn, G_FILE_TEST_EXISTS)) {
+		g_test_skip ("no flatpak test repo");
+		return;
+	}
+
+	/* add indirection so we can switch this after install */
+	g_assert (symlink (repodir1_fn, "/var/tmp/self-test/repo") == 0);
```
cosimoc: Use `g_assert_cmpint()`

plugins/flatpak/gs-self-test.c
```diff
@@ -1483,6 +1483,113 @@ gs_plugins_flatpak_app_update_func (GsPluginLoader *plugin_loader)
...
+	gs_app_set_management_plugin (app_source, "flatpak");
+	gs_app_set_state (app_source, AS_APP_STATE_AVAILABLE);
+	gs_flatpak_app_set_repo_url (app_source, "file:///var/tmp/self-test/repo");
+	plugin_job = gs_plugin_job_newv (GS_PLUGIN_ACTION_INSTALL,
+					 "app", app_source,
+					 NULL);
+	ret = gs_plugin_loader_job_action (plugin_loader, plugin_job, NULL, &error);
+	gs_test_flush_main_context ();
+	g_assert_no_error (error);
+	g_assert (ret);
```
cosimoc: Use `g_assert_true()` (many occurrences)

plugins/flatpak/gs-self-test.c
```diff
@@ -1483,6 +1483,113 @@ gs_plugins_flatpak_app_update_func (GsPluginLoader *plugin_loader)
...
+	g_assert_cmpint (gs_app_get_state (app), ==, AS_APP_STATE_INSTALLED);
+	g_assert_cmpstr (gs_app_get_version (app), ==, "1.2.3");
+
+	/* check if the extension was installed */
+        global_cache = gs_plugin_loader_get_global_cache (plugin_loader);
+	extension = gs_app_list_lookup (global_cache, "user/flatpak/*/runtime/org.test.Chiron.Extension/master");
+	g_assert (extension != NULL);
```
cosimoc: `g_assert_nonnull()`

plugins/flatpak/gs-self-test.c
```diff
@@ -1483,6 +1483,113 @@ gs_plugins_flatpak_app_update_func (GsPluginLoader *plugin_loader)
...
+	g_object_unref (plugin_job);
+	plugin_job = gs_plugin_job_newv (GS_PLUGIN_ACTION_REMOVE,
+					 "app", app,
+					 NULL);
+	ret = gs_plugin_loader_job_action (plugin_loader, plugin_job, NULL, &error);
+	g_assert_no_error (error);
+	g_assert (ret);
+
+	/* verify that the extension has been removed by the app's removal */
+	g_assert (!gs_app_is_installed (extension));
```
cosimoc: `g_assert_false()`

plugins/flatpak/gs-flatpak.c
```diff
@@ -2499,10 +2499,23 @@ gs_flatpak_get_list_for_remove (GsFlatpak *self, GsApp *app,
...
 		FlatpakRelatedRef *xref_related = g_ptr_array_index (related, i);
 		g_autoptr(GsApp) app_tmp = NULL;
+		g_autoptr(GError) error_local = NULL;
+
 		if (!flatpak_related_ref_should_delete (xref_related))
 			continue;
 		app_tmp = gs_flatpak_create_app (self, FLATPAK_REF (xref_related));
 		gs_app_set_origin (app_tmp, gs_app_get_origin (app));
+		if (!gs_plugin_refine_item_state (self, app_tmp, cancellable, &error_local)) {
+			if (g_cancellable_is_cancelled (cancellable)) {
```
cosimoc: Any reason not to use `g_error_matches()` instead? Sounds more robust.

plugins/flatpak/gs-self-test.c
```diff
@@ -1577,6 +1593,92 @@ gs_plugins_flatpak_runtime_extension_func (GsPluginLoader *plugin_loader)
 	g_assert (extension != NULL);
 	g_assert_cmpint (gs_app_get_state (extension), ==, AS_APP_STATE_INSTALLED);
 
+	/* switch to the new repo (to get the update) */
+	g_assert (unlink ("/var/tmp/self-test/repo") == 0);
```
cosimoc: `g_assert_cmpint()`

plugins/flatpak/gs-self-test.c
joaquimrocha: This was copied from previous code by Richard, but of course I can change it.

plugins/flatpak/gs-self-test.c
joaquimrocha: Ditto.

plugins/flatpak/gs-self-test.c
joaquimrocha: All these were copied too :)

plugins/flatpak/gs-flatpak.c
joaquimrocha: Well, I only really want to care about the cancellation. I could check the error for `G_IO_ERROR_CANCELLED` but then we have to assume that this is the one that gets assigned on cancellation. I will do it.

plugins/flatpak/gs-flatpak.c
joaquimrocha: I get your point but without the warning (it's slightly different in each version) then we'd not be saving a lot (and I am not a fan of printing debug/warning messages from shared helper functions).

plugins/flatpak/gs-self-test.c
joaquimrocha: Ah.. the tabs curse thingy... Thanks.

plugins/flatpak/gs-flatpak-app.h
joaquimrocha: I agree. I will rename it as `gs_flatpak_app_get_main_app_ref`.

plugins/flatpak/gs-flatpak.c
```diff
@@ -1301,6 +1320,39 @@ gs_flatpak_app_install_source (GsFlatpak *self, GsApp *app,
...
+			 GsApp *related_app,
+			 GCancellable *cancellable,
+			 GError **error)
+{
+	g_autofree gchar *main_app_name = g_strdup (gs_flatpak_app_get_ref_name (related_app));
+	g_autoptr(FlatpakInstalledRef) ref = NULL;
+	const gchar *app_name;
+	g_auto(GStrv) app_tokens = NULL;
+
+        app_name = gs_flatpak_app_get_main_app_ref (related_app);
```
cosimoc: Indentation is off here

plugins/flatpak/gs-flatpak.c
cosimoc: Still applies

plugins/flatpak/gs-flatpak.c
cosimoc: Also need to set the error when returning `NULL`

plugins/flatpak/gs-flatpak.c
cosimoc: The `GError` is actually already set in this case.

plugins/flatpak/gs-flatpak.c
cosimoc: Still applies

plugins/flatpak/gs-flatpak.c
cosimoc: If `main_app` gains e.g. autoptr, you'll need to increase the ref here.

plugins/flatpak/gs-flatpak.c
cosimoc: Still applies

plugins/flatpak/gs-self-test.c
```diff
@@ -1483,6 +1483,214 @@ gs_plugins_flatpak_app_update_func (GsPluginLoader *plugin_loader)
...
+	g_assert_cmpstr (gs_app_get_version (app), ==, "1.2.3");
+
+	/* check if the extension was installed */
+	global_cache = gs_plugin_loader_get_global_cache (plugin_loader);
+	extension = gs_app_list_lookup (global_cache, "user/flatpak/*/runtime/org.test.Chiron.Extension/master");
+	g_assert_nonnull (extension);
+	g_assert_cmpint (gs_app_get_state (extension), ==, AS_APP_STATE_INSTALLED);
+
+	/* switch to the new repo (to get the update) */
+	g_assert (unlink ("/var/tmp/self-test/repo") == 0);
```
cosimoc: `g_assert_cmpint()`

plugins/flatpak/gs-self-test.c
```diff
@@ -1483,6 +1483,214 @@ gs_plugins_flatpak_app_update_func (GsPluginLoader *plugin_loader)
...
+	/* refresh the appstream metadata */
+	g_object_unref (plugin_job);
+	plugin_job = gs_plugin_job_newv (GS_PLUGIN_ACTION_REFRESH,
+					 "age", (guint64) 0, /* force now */
+					 "refresh-flags", GS_PLUGIN_REFRESH_FLAGS_METADATA |
+							  GS_PLUGIN_REFRESH_FLAGS_PAYLOAD,
+					 NULL);
+	ret = gs_plugin_loader_job_action (plugin_loader, plugin_job, NULL, &error);
+	g_assert_no_error (error);
+	g_assert (ret);
```
cosimoc: `g_assert_true()`

plugins/flatpak/gs-self-test.c
```diff
@@ -1483,6 +1483,214 @@ gs_plugins_flatpak_app_update_func (GsPluginLoader *plugin_loader)
...
+	/* get the updates list */
+	g_object_unref (plugin_job);
+	plugin_job = gs_plugin_job_newv (GS_PLUGIN_ACTION_GET_UPDATES,
+					 "refine-flags", GS_PLUGIN_REFINE_FLAGS_REQUIRE_ICON |
+							 GS_PLUGIN_REFINE_FLAGS_REQUIRE_UPDATE_DETAILS,
+					 NULL);
+	list_updates = gs_plugin_loader_job_process (plugin_loader, plugin_job, NULL, &error);
+	gs_test_flush_main_context ();
+	g_assert_no_error (error);
+	g_assert (list_updates != NULL);
```
cosimoc: `g_assert_nonnull()`

plugins/flatpak/gs-self-test.c
```diff
@@ -1483,6 +1483,214 @@ gs_plugins_flatpak_app_update_func (GsPluginLoader *plugin_loader)
...
+
+	g_assert_cmpint (gs_app_list_length (list_updates), ==, 1);
+	for (guint i = 0; i < gs_app_list_length (list_updates); i++) {
+		app_tmp = gs_app_list_index (list_updates, i);
+		g_debug ("got update %s", gs_app_get_unique_id (app_tmp));
+	}
+
+	/* check that the extension has no update */
+	app_tmp = gs_app_list_lookup (list_updates, "*/flatpak/test/*/org.test.Chiron.Extension/*");
+	g_assert (app_tmp == NULL);
```
cosimoc: `g_assert_null()`

plugins/flatpak/gs-self-test.c
```diff
@@ -1483,6 +1483,214 @@ gs_plugins_flatpak_app_update_func (GsPluginLoader *plugin_loader)
...
+					 NULL);
+	gs_plugin_loader_job_process_async (plugin_loader, plugin_job,
+					    NULL,
+					    update_app_action_finish_sync,
+					    loop);
+	g_main_loop_run (loop);
+	gs_test_flush_main_context ();
+	g_assert_cmpint (gs_app_get_state (app), ==, AS_APP_STATE_INSTALLED);
+	g_assert_cmpstr (gs_app_get_version (app), ==, "1.2.3");
+	g_assert (got_progress_installing);
```
cosimoc: `g_assert_true()`

plugins/flatpak/gs-flatpak.c
joaquimrocha: I will do it but only when the error is not `...NO_DATA` because not all runtimes are extensions.

plugins/flatpak/gs-flatpak.c
mariospr: Ok. No big deal anyway

plugins/flatpak/gs-flatpak.c
mariospr: Oh, right! Sorry for the noise

plugins/flatpak/gs-flatpak.c
mariospr: Ok, but then a comment in the code explaining why that exception would be useful

plugins/flatpak/gs-flatpak.c
```diff
@@ -1388,10 +1469,29 @@ gs_flatpak_add_updates_pending (GsFlatpak *self, GsAppList *list,
...
 		}
-		gs_app_set_state (app, AS_APP_STATE_UPDATABLE_LIVE);
+
+		if (gs_app_get_kind (app) == AS_APP_KIND_RUNTIME)
+			main_app = get_main_app_of_related (self, app, cancellable,
+							    &error_local);
+
+		if (main_app == NULL) {
+			if (error_local != NULL &&
+			    !g_error_matches (error_local, G_IO_ERROR, G_IO_ERROR_NOT_FOUND)) {
```
mariospr: I think a comment here explaining why you make an exception for `_NOT_FOUND` (i.e. not all runtimes are extensions) would be helpful to understand this line.

Also, if `error_local != NOT_FOUND` is considered a legit problem, I think that should be printed with a `g_warning()`, not `g_debug()` (which you could use it if you want for an `else` branch.

plugins/flatpak/gs-flatpak.c
```diff
@@ -1388,10 +1469,29 @@ gs_flatpak_add_updates_pending (GsFlatpak *self, GsAppList *list,
...
+				g_debug ("Couldn't get the main app for related app %s: %s; continuing",
```
mariospr: The text in this `g_debug()` suggests that a `continue` is missing.

joaquimrocha: Hmmm, not really. Why would it have to be tied to a `continue` statement? What it means is that it's not aborting the main op. This is only because in the past we've seen some scary warnings only to find out they don't skip any execution.

mariospr: Ok, in that case the text in there is a bit confusing. I'd suggest replacing "continuing" for something that better reflects what's happening right afterwards (`main_app = app`)

joaquimrocha: That block of code was getting complex with all the error handling and comments so I split it out into a different function. Not a big fan of that because of the warning/debug messages but I think overall it's better this way. Because of that I have removed the continuing (as in the context of a helper function is not so useful anymore).

plugins/flatpak/gs-flatpak.c
```diff
@@ -1301,6 +1320,77 @@ gs_flatpak_app_install_source (GsFlatpak *self, GsApp *app,
...
+
+	if (ref == NULL)
+		return NULL;
+
+	return gs_flatpak_create_installed (self, ref, error);
+}
+
+static GsApp *
+get_real_app_for_update (GsFlatpak *self, GsApp *app, GCancellable *cancellable,
+			GError **error)
```
mariospr: Wrong indentation

mariospr: Also, seeing other functions (e.g. `get_main_app_of_related()`)  it feels like you want to put each param in a single line

plugins/flatpak/gs-flatpak.c
```diff
@@ -1348,12 +1439,15 @@ gs_flatpak_add_updates (GsFlatpak *self, GsAppList *list,
...
 			continue;
 		}
-		gs_app_set_state (app, AS_APP_STATE_UPDATABLE_LIVE);
-		gs_app_set_update_details (app, NULL);
-		gs_app_set_update_version (app, NULL);
-		gs_app_set_update_urgency (app, AS_URGENCY_KIND_UNKNOWN);
-		gs_app_set_size_download (app, 0);
-		gs_app_list_add (list, app);
+
+		main_app = get_real_app_for_update (self, app, cancellable, error);
```
mariospr: It looks like `get_real_app_for_update()` would always return a valid `GsApp*`, and never `NULL`, but that's not clear when reading the code here unless you check that function's implementation. However, I'm not entirely convinced that this function can never fail because the operation could get cancelled (see my other comment).

So, please add an assertion if this is actually guaranteed to never return NULL, or a NULL-check plus an early return otherwise.

plugins/flatpak/gs-flatpak.c
```diff
@@ -1301,6 +1320,77 @@ gs_flatpak_app_install_source (GsFlatpak *self, GsApp *app,
...
+
+	return gs_flatpak_create_installed (self, ref, error);
+}
+
+static GsApp *
+get_real_app_for_update (GsFlatpak *self, GsApp *app, GCancellable *cancellable,
+			GError **error)
+{
+	GsApp *main_app = NULL;
+	g_autoptr(GError) error_local = NULL;
```
mariospr: It looks like you're using this `error_local` so that you can simply report with a warning when `error != NOT_FOUND`, but then that means you don't need to pass an `GError **error` parameter to this `get_real_app_for_update()` function, as it would be guaranteed to always return a valid `GsApp*`, and never `NULL`.

plugins/flatpak/gs-flatpak.c
mariospr: Hmm... after reading the rest of the code and realizing that `get_real_app_for_update()` is not expected to ever return NULL (thus it's not expected to fail), it looks like you should not pass a `GCancellable *cancellable` here, and simply use `NULL` below when calling `get_main_app_of_related()`, since there's nothing you would do in case the operation gets cancelled.

But that depends on whether it's realistically possible to get this action cancelled or not:
  * If it's possible to cancel this function, I think you should keep both the `GCancellable` and the `GError **` parameters, handle the cancellable error as a real error below and therefore accepting there's at least a case where this function can return NULL, thus checking for that option in the caller.
  * If it's NOT possible to cancel this function, then you should simply drop both the `GCancellable` and the `GError **` parameters from this function, and simply use `NULL` below when calling `get_main_app_of_related()`, as mentioned above (and add an assertion in the caller, after using this function).

joaquimrocha: About adding a parameter per line, it's not really something enforced or recommended in g-s, but I will change it.

About the function never returning NULL, you're right. It's a bit messy and I think that we should instead abort when the error is other than `NOT_FOUND`.

mariospr:
Typo in the commit message of the last patch: `instaled` -> `installed`

plugins/flatpak/gs-flatpak.c
```diff
@@ -2808,24 +2811,24 @@ gs_flatpak_get_list_for_install (GsFlatpak *self, GsApp *app,
 		/* already installed? */
 		app_tmp = gs_flatpak_create_app (self, FLATPAK_REF (xref_related));
 		ref_display = gs_flatpak_app_get_ref_display (app_tmp);
-		if (g_hash_table_contains (hash_installed, ref_display)) {
+		if (!is_update && g_hash_table_contains (hash_installed, ref_display)) {
 			g_debug ("not adding related %s as already installed", ref_display);
```
mariospr: Could you add a `continue` after this line and then un-indent the code inside the `else { }` branch one level to the left? I think the code would be clearer that way.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2736,8 +2736,11 @@ gs_flatpak_refine_runtime_for_install (GsFlatpak *self,
 }
 
 static GsAppList *
-gs_flatpak_get_list_for_install (GsFlatpak *self, GsApp *app,
-				 GCancellable *cancellable, GError **error)
+gs_flatpak_get_list_for_install_or_update (GsFlatpak *self,
+					   GsApp *app,
+					   gboolean is_update,
```
mariospr: I'd normally suggest using an `enum` here instead of a gboolean as that's usually a better thing to do, but since this is a private static function only used from 2 places, I think it's fine. Your call.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2808,24 +2811,24 @@ gs_flatpak_get_list_for_install (GsFlatpak *self, GsApp *app,
...
 				g_warning ("Failed to refine %s when getting the list of apps to "
-					   "install: %s; not adding the app...",
+					   "install/update: %s; not adding the app...",
 					   gs_app_get_unique_id (app_tmp),
 					   error_local->message);
 				continue;
 			}
-
+			if (is_update && !gs_app_is_updatable (app_tmp))
+				continue;
```
mariospr: A g_debug() here might be useful too

plugins/flatpak/gs-flatpak.c
joaquimrocha: Good idea.

plugins/flatpak/gs-flatpak.c
joaquimrocha: I prefer to keep the boolean for now.

plugins/flatpak/gs-flatpak.c
```diff
@@ -1301,6 +1320,85 @@ gs_flatpak_app_install_source (GsFlatpak *self, GsApp *app,
 	return TRUE;
 }
 
+static GsApp *
+get_main_app_of_related (GsFlatpak *self,
+			 GsApp *related_app,
+			 GCancellable *cancellable,
+			 GError **error)
+{
+	g_autofree gchar *main_app_name = g_strdup (gs_flatpak_app_get_ref_name (related_app));
```
cosimoc: This variable is unused

joaquimrocha: Sorry. That was for a previous version where I had to put a null char in the string directly, but I think the compiler for some reason doesn't warn me about unused variables when they have `g_auto...`.

PR #290, opened 2017-12-04
Remove the "generic-updates" plugin

joaquimrocha:
This plugin is not only not needed by us but is also creating a problem
by converting any packages that are of the kind "generic" into
OS updates, and this is happening with some of our apps.

https://phabricator.endlessm.com/T20135


rshuler:
Please reword the commit message: "not only needed by us" implies that it is needed by us.

Possible better wording:
"not only not needed by us"
"not only unneeded by us"
"This plugin is not needed by us, and it creates a problem..." (most direct and my preference)

joaquimrocha:
Yeah. I missed the extra 'not' in my original message. Should be fine now.

PR #303, opened 2018-01-10
Install the mime type related files

joaquimrocha:
https://phabricator.endlessm.com/T16949

### DIFF
--- a/debian/gnome-software.install
+++ b/debian/gnome-software.install
@@ -13,4 +13,5 @@ usr/share/eos-discovery-feed/
 usr/share/glib-2.0/schemas/
 usr/share/gnome-shell/
 usr/share/man/
+usr/share/mime/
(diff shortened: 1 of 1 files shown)


debian/gnome-software.install
```diff
@@ -13,4 +13,5 @@ usr/share/eos-discovery-feed/
 usr/share/glib-2.0/schemas/
 usr/share/gnome-shell/
 usr/share/man/
+usr/share/mime
```
pwithnall: Nitpick: I think it’s conventional to end the line with `/` if you intend to install the whole directory, otherwise this could be mistaken for a file called `mime` in the `/usr/share` directory.

joaquimrocha: Good point. Updated. Thanks.

PR #305, opened 2018-01-16
T19569

joaquimrocha:
https://phabricator.endlessm.com/T19569

### DIFF
--- a/src/gs-application.c
+++ b/src/gs-application.c
@@ -895,6 +895,7 @@ gs_application_startup (GApplication *application)
 {
 	GSettings *settings;
 	GsApplication *app = GS_APPLICATION (application);
+	g_autoptr(GAction) action = NULL;
 	G_APPLICATION_CLASS (gs_application_parent_class)->startup (application);
 
 	gs_application_add_wrapper_actions (application);
@@ -913,6 +914,9 @@ gs_application_startup (GApplication *application)
 	g_signal_connect_swapped (settings, "changed",
 				  G_CALLBACK (gs_application_settings_changed_cb),
 				  application);
+
+	action = g_settings_create_action (app->settings, "download-updates");
+	g_action_map_add_action (G_ACTION_MAP (app), action);
 
 	gs_application_initialize_ui (app);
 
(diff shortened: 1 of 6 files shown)


src/gs-update-monitor.c
```diff
@@ -788,17 +812,36 @@ gs_update_monitor_init (GsUpdateMonitor *monitor)
...
 static void
 gs_update_monitor_dispose (GObject *object)
 {
 	GsUpdateMonitor *monitor = GS_UPDATE_MONITOR (object);
 
+	if (monitor->network_changed_handler != 0) {
+		g_signal_handler_disconnect (monitor->network_monitor,
+					     monitor->network_changed_handler);
+		monitor->network_changed_handler = 0;
+	}
```
GeorgesStavracas: I think you should also drop the `GNetworkMonitor` reference that was taken in `init()`.

src/gs-update-monitor.c
```diff
@@ -788,17 +812,36 @@ gs_update_monitor_init (GsUpdateMonitor *monitor)
...
+
 	if (monitor->cancellable) {
 		g_cancellable_cancel (monitor->cancellable);
 		g_clear_object (&monitor->cancellable);
 	}
+	if (monitor->network_changed_handler) {
```
GeorgesStavracas: I guess it should be `monitor->network_cancellable` here?

src/gs-application.c
```diff
@@ -894,6 +894,7 @@ static void
 gs_application_startup (GApplication *application)
 {
 	GSettings *settings;
+	GAction *action;
```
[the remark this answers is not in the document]
joaquimrocha: Good catch!

src/gs-update-monitor.c
joaquimrocha: Thanks. Failed to spot the mistake in my last check.

src/gs-update-monitor.c
GeorgesStavracas: Actually, my comment is plain wrong

PR #307, opened 2018-01-22
flatpak: Fix EknServices-picking logic

ptomato:
The previous code would start looking for com.endlessm.EknServices3 when
installing an SDK3 app. That is incorrect, it should keep using
com.endlessm.EknServices2 until we bump the version number of that
service due to some backwards-incompatibility.

https://phabricator.endlessm.com/T20846


joaquimrocha:
I think we should have a more future-proof solution here (i.e. looking for the latest EknServices available in the remote) so we don't have to always come back to this solution which will only work for users who upgrade the OS (and it's likely Flatpaks get updated before the OS does).

I am still merging this since it's an easy change that may be needed for now.

PR #314, opened 2018-01-27
flatpak: add update_remote_sync call to appstream refresh

ramcq:
This applies changes to the flatpak remote which come from the summary file,
such as updated keys, titles or the URL. Adding it into the appstream refresh
function means it's behind the timestamp checks and skipping broken remotes,
which avoids adding network roundtrips to the refresh.  libflatpak caches the
summary file in the common case that the remote details have not changed.

https://phabricator.endlessm.com/T20328

### DIFF
--- a/plugins/flatpak/gs-flatpak.c
+++ b/plugins/flatpak/gs-flatpak.c
@@ -559,18 +559,6 @@ gs_flatpak_add_apps_from_xremote (GsFlatpak *self,
 	if (g_settings_get_boolean (settings, "filter-default-branch"))
 		default_branch = flatpak_remote_get_default_branch (xremote);
 
-	/* on Endless we rely heavily on always having the default branch
-	 * defined for the 'eos-apps' repository, so we need a fallback value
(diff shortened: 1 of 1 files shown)


plugins/flatpak/gs-flatpak.c
```diff
@@ -857,6 +858,18 @@ gs_flatpak_refresh_appstream_remote (GsFlatpak *self,
...
 	gs_plugin_status_update (self->plugin, app_dl, GS_PLUGIN_STATUS_DOWNLOADING);
+
+	if (!flatpak_installation_update_remote_sync (self->installation,
+						      remote_name,
+						      cancellable,
+						      &local_error)) {
+		g_debug ("Failed to update metadata for remote %s: %s\n",
+			 remote_name, local_error->message);
+		gs_flatpak_error_convert (&local_error);
+		g_propagate_error (error, g_pointer_steal (local_error));
```
joaquimrocha: I think you meant `g_pointer_steal`. And it should be used with `&local_error` in this case.

PR #326, opened 2018-02-19
T21194

joaquimrocha:
https://phabricator.endlessm.com/T21194

### DIFF
--- a/lib/gs-plugin-loader.c
+++ b/lib/gs-plugin-loader.c
@@ -53,7 +53,7 @@ typedef struct
 	GsPluginStatus		 global_status_last;
 
 	GMutex			 pending_apps_mutex;
-	GPtrArray		*pending_apps;
+	GsAppList		*pending_apps;
 
 	GThreadPool		*queued_ops_pool;
 
@@ -1695,7 +1695,7 @@ gs_plugin_loader_pending_apps_add (GsPluginLoader *plugin_loader,
 	g_assert (gs_app_list_length (list) > 0);
 	for (guint i = 0; i < gs_app_list_length (list); i++) {
 		GsApp *app = gs_app_list_index (list, i);
-		g_ptr_array_add (priv->pending_apps, g_object_ref (app));
+		gs_app_list_add (priv->pending_apps, app);
 		/* make sure the progress is properly initialized */
 		gs_app_set_progress (app, 0);
 	}
@@ -1713,7 +1713,7 @@ gs_plugin_loader_pending_apps_remove (GsPluginLoader *plugin_loader,
 	g_assert (gs_app_list_length (list) > 0);
 	for (guint i = 0; i < gs_app_list_length (list); i++) {
 		GsApp *app = gs_app_list_index (list, i);
-		g_ptr_array_remove (priv->pending_apps, app);
+		gs_app_list_remove (priv->pending_apps, app);
 
 		/* check the app is not still in an action helper */
 		switch (gs_app_get_state (app)) {
@@ -1740,7 +1740,6 @@ load_install_queue (GsPluginLoader *plugin_loader, GError **error)
 	g_autofree gchar *contents = NULL;
 	g_autofree gchar *file = NULL;
 	g_auto(GStrv) names = NULL;
-	g_autoptr(GsAppList) list = NULL;
 	g_autoptr(GMutexLocker) locker = NULL;
 
 	/* load from file */
@@ -1754,8 +1753,8 @@ load_install_queue (GsPluginLoader *plugin_loader, GError **error)
 	if (!g_file_get_contents (file, &contents, NULL, error))
 		return FALSE;
 
+	locker = g_mutex_locker_new (&priv->pending_apps_mutex);
 	/* add each app-id */
-	list = gs_app_list_new ();
 	names = g_strsplit (contents, "\n", 0);
 	for (i = 0; names[i]; i++) {
 		g_autoptr(GsApp) app = NULL;
@@ -1766,30 +1765,10 @@ load_install_queue (GsPluginLoader *plugin_loader, GError **error)
 		if (!as_utils_unique_id_valid (id))
 			continue;
 
-		app = gs_app_new (NULL);
-		gs_app_set_from_unique_id (app, id);
+		app = gs_plugin_loader_app_create (plugin_loader, id);
 
-		gs_app_list_add (list, app);
-	}
-
-	/* refine */
-	if (gs_app_list_length (list) > 0) {
-		g_autoptr(GsPluginLoaderHelper) helper = NULL;
-		g_autoptr(GsPluginJob) plugin_job = NULL;
-		plugin_job = gs_plugin_job_newv (GS_PLUGIN_ACTION_REFINE, NULL);
-		helper = gs_plugin_loader_helper_new (plugin_loader, plugin_job);
-		gs_plugin_job_set_failure_flags (helper->plugin_job,
-						 GS_PLUGIN_FAILURE_FLAGS_USE_EVENTS);
-		if (!gs_plugin_loader_run_refine (helper, list, NULL, error))
-			return FALSE;
-	}
-
-	/* add the pending apps now */
-	locker = g_mutex_locker_new (&priv->pending_apps_mutex);
-	for (i = 0; i < gs_app_list_length (list); ++i) {
-		GsApp *app = gs_app_list_index (list, i);
-		g_ptr_array_add (priv->pending_apps, g_object_ref (app));
 		g_debug ("adding pending app %s", gs_app_get_unique_id (app));
+		gs_app_list_add (priv->pending_apps, app);
 	}
 
 	return TRUE;
@@ -1799,7 +1778,6 @@ static void
 save_install_queue (GsPluginLoader *plugin_loader)
 {
 	GsPluginLoaderPrivate *priv = gs_plugin_loader_get_instance_private (plugin_loader);
-	GPtrArray *pending_apps;
 	gboolean ret;
 	gint i;
 	g_autoptr(GError) error = NULL;
@@ -1807,15 +1785,11 @@ save_install_queue (GsPluginLoader *plugin_loader)
 	g_autofree gchar *file = NULL;
 
 	s = g_string_new ("");
-	pending_apps = priv->pending_apps;
 	g_mutex_lock (&priv->pending_apps_mutex);
-	for (i = (gint) pending_apps->len - 1; i >= 0; i--) {
-		GsApp *app;
-		app = g_ptr_array_index (pending_apps, i);
-		if (gs_app_get_state (app) == AS_APP_STATE_QUEUED_FOR_INSTALL) {
-			g_string_append (s, gs_app_get_unique_id (app));
-			g_string_append_c (s, '\n');
-		}
+	for (i = (gint) gs_app_list_length (priv->pending_apps) - 1; i >= 0; i--) {
+		GsApp *app = gs_app_list_index (priv->pending_apps, i);
+		g_string_append (s, gs_app_get_unique_id (app));
+		g_string_append_c (s, '\n');
 	}
 	g_mutex_unlock (&priv->pending_apps_mutex);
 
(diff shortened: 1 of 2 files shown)


src/gs-installed-page.c
```diff
@@ -546,44 +553,74 @@ gs_installed_page_has_app (GsInstalledPage *self,
...
 
-		/* do not to add pending apps more than once. */
-		if (gs_installed_page_has_app (self, app) == FALSE)
-			gs_installed_page_add_app (self, pending, app);
+		if (gs_app_get_state (app) == AS_APP_STATE_AVAILABLE)
+			gs_app_set_state (app, AS_APP_STATE_QUEUED_FOR_INSTALL);
 
-		/* incremement the label */
-		cnt++;
+		++pending_apps_count;
```
rshuler: Again, for consistency, I prefer `pending_apps_count++;`, but I can live with it as is if you feel strongly.

src/gs-installed-page.c
```diff
@@ -546,44 +553,74 @@ gs_installed_page_has_app (GsInstalledPage *self,
...
+		if (!g_error_matches (error, GS_PLUGIN_ERROR, GS_PLUGIN_ERROR_CANCELLED))
+			g_warning ("failed to refine pending apps: %s", error->message);
+		goto out;
+	}
 
-	/* add new apps to the list */
-	pending = gs_plugin_loader_get_pending (plugin_loader);
-	for (i = 0; i < gs_app_list_length (pending); i++) {
-		app = gs_app_list_index (pending, i);
+	for (guint i = 0; i < gs_app_list_length (list); ++i) {
```
rshuler: Stylistically (and for consistency with the rest of gnome-software), I prefer `i++` (reserving `++i` for those cases where the order of operation actually matters), but I can live with it as is if you feel strongly (or has already been approved with `++i` upstream).

lib/gs-plugin-loader.c
```diff
@@ -1797,23 +1775,18 @@ static void
...
 	g_autoptr(GString) s = NULL;
 	g_autofree gchar *file = NULL;
 
 	s = g_string_new ("");
-	pending_apps = priv->pending_apps;
 	g_mutex_lock (&priv->pending_apps_mutex);
-	for (i = (gint) pending_apps->len - 1; i >= 0; i--) {
-		GsApp *app;
-		app = g_ptr_array_index (pending_apps, i);
-		if (gs_app_get_state (app) == AS_APP_STATE_QUEUED_FOR_INSTALL) {
```
rshuler: Is it correct that this check for AS_APP_STATE_QUEUED_FOR_INSTALL is no longer necessary?

lib/gs-plugin-loader.c
```diff
@@ -1916,16 +1890,10 @@ GsAppList *
 gs_plugin_loader_get_pending (GsPluginLoader *plugin_loader)
 {
 	GsPluginLoaderPrivate *priv = gs_plugin_loader_get_instance_private (plugin_loader);
-	GsAppList *array;
-	guint i;
+	g_autoptr(GMutexLocker) locker = g_mutex_locker_new (&priv->pending_apps_mutex);
```
rshuler: Again, why the change to g_mutex_locker_new rather than just keep the g_mutex_lock / g_mutex_unlock?  And, I suppose it doesn't really matter, but do you intentionally include the gs_app_list_new() within the mutex where it used to be outside?

lib/gs-plugin-loader.c
```diff
@@ -2978,10 +2946,9 @@ gs_plugin_loader_network_changed_cb (GNetworkMonitor *monitor,
 		g_autoptr(GsAppList) queue = NULL;
 		g_mutex_lock (&priv->pending_apps_mutex);
 		queue = gs_app_list_new ();
-		for (guint i = 0; i < priv->pending_apps->len; i++) {
-			GsApp *app = g_ptr_array_index (priv->pending_apps, i);
-			if (gs_app_get_state (app) == AS_APP_STATE_QUEUED_FOR_INSTALL)
```
rshuler: Again, is it safe to skip the check for AS_APP_STATE_QUEUED_FOR_INSTALL?

lib/gs-plugin-loader.c
```diff
@@ -1776,14 +1776,13 @@ save_install_queue (GsPluginLoader *plugin_loader)
...
 	gboolean ret;
-	gint i;
 	g_autoptr(GError) error = NULL;
 	g_autoptr(GString) s = NULL;
 	g_autofree gchar *file = NULL;
 
 	s = g_string_new ("");
 	g_mutex_lock (&priv->pending_apps_mutex);
-	for (i = (gint) gs_app_list_length (priv->pending_apps) - 1; i >= 0; i--) {
+	for (guint i = 0; i < gs_app_list_length (priv->pending_apps); ++i) {
```
rshuler: Again, might want `i++` for consistency.

lib/gs-plugin-loader.c
```diff
@@ -1700,18 +1700,56 @@ gs_plugin_loader_pending_apps_add (GsPluginLoader *plugin_loader,
...
 	g_autoptr(GMutexLocker) locker = g_mutex_locker_new (&priv->pending_apps_mutex);
+	gboolean save_queue = FALSE;
 
 	g_assert (gs_app_list_length (list) > 0);
 	for (guint i = 0; i < gs_app_list_length (list); i++) {
 		GsApp *app = gs_app_list_index (list, i);
-		gs_app_list_remove (priv->pending_apps, app);
+		if (gs_app_get_state (app) != AS_APP_STATE_QUEUED_FOR_INSTALL) {
+			save_queue = TRUE;
+			gs_app_list_remove (priv->pending_apps, app);
```
rshuler: Is it okay that the removal is within the check for AS_APP_STATE_QUEUED_FOR_INSTALL, where previously there was no such check?

lib/gs-plugin-loader.c
```diff
@@ -1727,6 +1765,10 @@ gs_plugin_loader_pending_apps_remove (GsPluginLoader *plugin_loader,
 		}
 
 	}
+	if (save_queue) {
+		g_clear_pointer (&locker, g_mutex_locker_free);
+		save_install_queue (plugin_loader);
+	}
 	g_idle_add (emit_pending_apps_idle, g_object_ref (plugin_loader));
```
rshuler: Is it okay that this line may be within or outside the mutex depending on the state of save_queue?

lib/gs-plugin-loader.c
```diff
@@ -1781,6 +1781,15 @@ load_install_queue (GsPluginLoader *plugin_loader, GError **error)
 		if (!gs_plugin_loader_run_refine (helper, list, NULL, error))
 			return FALSE;
 	}
+
+	/* add the pending apps now */
+	locker = g_mutex_locker_new (&priv->pending_apps_mutex);
+	for (i = 0; i < gs_app_list_length (list); ++i) {
+		GsApp *app = gs_app_list_index (list, i);
+		g_ptr_array_add (priv->pending_apps, g_object_ref (app));
+		g_debug ("adding pending app %s", gs_app_get_unique_id (app));
```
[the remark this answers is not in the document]
joaquimrocha: Isn't this what's happening?

lib/gs-plugin-loader.c
```diff
@@ -1781,6 +1781,15 @@ load_install_queue (GsPluginLoader *plugin_loader, GError **error)
 		if (!gs_plugin_loader_run_refine (helper, list, NULL, error))
 			return FALSE;
 	}
+
+	/* add the pending apps now */
+	locker = g_mutex_locker_new (&priv->pending_apps_mutex);
```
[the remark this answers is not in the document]
joaquimrocha: I wanted to lock the whole array while adding apps. And I used the locker because it's safer like an autoptr.

src/gs-installed-page.c
joaquimrocha: I got used to the `++i` with the excuse that it's more efficient if the compiler didn't optimize it, so I like to use it. Let's hear from upstream, I have rebased + pushed the PR again upstream today.

rshuler: As I understand it, all modern compilers will optimize it.  In any case, whatever upstream likes is fine with me.  I just noticed that the style appeared to post-increment elsewhere in gnome-software, so I thought it would be nice to remain consistent, but in the end I really don't care.

lib/gs-plugin-loader.c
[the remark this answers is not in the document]
rshuler: Yes, that does seem to be the case here.  I don't see why I commented here (especially on a commit revert).

lib/gs-plugin-loader.c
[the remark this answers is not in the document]
rshuler: Okay.  Since there is not much going on in this loop, it does make sense to have the mutex around the entire block.

lib/gs-plugin-loader.c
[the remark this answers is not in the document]
rshuler: What!!!  This is crazy...  This conversation shows up on two commits.  My comment is with regard to https://github.com/endlessm/gnome-software/pull/326/commits/416ad38c61158e3448db8e7adf6f4bb50f1970a3, where the g_debug statement was added after the g_ptr_array_add.  (So, I don't think I had tried to comment on the revert commit https://github.com/endlessm/gnome-software/pull/326/commits/22562db749693570232863a8685e65c116e0ab95 after all.)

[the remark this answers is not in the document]
rshuler: My comment also applies to https://github.com/endlessm/gnome-software/pull/326/commits/27a5e769746c7ac2dd641f3b01e55d5708dadcb0.  I don't understand how github is associating the comments with the individual commits.

lib/gs-plugin-loader.c
rshuler: This question still stands, at least with regard to commit https://github.com/endlessm/gnome-software/pull/326/commits/27a5e769746c7ac2dd641f3b01e55d5708dadcb0

lib/gs-plugin-loader.c
rshuler: And the question still applies here, too.

lib/gs-plugin-loader.c
rshuler: And this question still applies, too.

lib/gs-plugin-loader.c
rshuler: And, just to be sure, can you answer this question, too.

lib/gs-plugin-loader.c
```diff
@@ -1729,9 +1767,54 @@ gs_plugin_loader_pending_apps_remove (GsPluginLoader *plugin_loader,
...
+		const gchar *id = names[i];
+
+		if (strlen (id) == 0)
+			continue;
+		if (!as_utils_unique_id_valid (id))
+			continue;
+
+		app = gs_plugin_loader_app_create (plugin_loader, id);
+		gs_app_list_add (priv->pending_apps, app);
+		g_debug ("adding pending app %s", gs_app_get_unique_id (app));
```
rshuler: On the final file comparison, here is the debug statement after the action.

lib/gs-plugin-loader.c
joaquimrocha: As far as I know, this just adds the callback + args to be called with idle priority, so it should return before the function is called, which would unlock the mutex. There are other places where this is called while locked, but I can look into how glib manages all this if needed (not today though).

lib/gs-plugin-loader.c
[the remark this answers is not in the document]
joaquimrocha: In https://github.com/endlessm/gnome-software/commit/27a5e769746c7ac2dd641f3b01e55d5708dadcb0 I see only one `g_debug` which didn't touch. GitHub (but also GitLab) also seem a bit messy in the reviews system for me, I always get a bit lost and honestly didn't see the other questions you reminded me I had to answer too...

lib/gs-plugin-loader.c
joaquimrocha: Ah, I see it now. I can change that.

lib/gs-plugin-loader.c
joaquimrocha: Honestly, since the commit is months old already, I don't remember exactly. I think there were several problems related to the apps not being in this state while dealing with them, and decoupling the state from the queue means we don't have to rely on the state (one less way to get the typical issues we have with g-s).

I'd have to thoroughly retest all this again to give you a better answer (let me know if I should do it).

lib/gs-plugin-loader.c
joaquimrocha: It's because it's better to use the RAII, and I modified most of the function so I thought it'd be nice to sneak in this change. I will change the order though as you pointed out.

lib/gs-plugin-loader.c
joaquimrocha: See the lousy answer to the other question.

lib/gs-plugin-loader.c
rshuler: Okay, as long as you think it was intentional, I'm fine.

PR #328, opened 2018-02-21
T19479: Search by Endless::HasDiscoveryFeedContent tag

smspillaz:
Adjust the search handler in `gs-appstream` to allow searching by metadata keys and do the entire search by `Endless::HasDiscoveryFeedContent`. This allows us to remove a previous hack which assumed that all discovery-feed capable apps started with the ID `com.endlessm` and then had to filter for the right ones later.

It might be the case that we are doing the search of the metadata in the wrong place. I took a look at appstream-glib and there doesn't seem to be a particularly obivous process for adding new search fields without hardcoding new ones.

https://phabricator.endlessm.com/T19479

### DIFF
--- a/plugins/core/gs-appstream.c
+++ b/plugins/core/gs-appstream.c
@@ -829,6 +829,34 @@ gs_appstream_refine_app (GsPlugin *plugin,
 }
 
 static gboolean
+matches_all_on_metadata (AsApp  *item,
+			 gchar **values)
+{
+	GHashTable *metadata = as_app_get_metadata (item);
+	gchar **value_iter = values;
+
+	for (; *value_iter != NULL; ++value_iter) {
+		GHashTableIter iter;
+		gboolean matched_value = FALSE;
+		gpointer key, value;
+
+		g_hash_table_iter_init (&iter, metadata);
+
+		while (g_hash_table_iter_next (&iter, &key, &value)) {
+			if (g_ascii_strcasecmp ((gchar *) key, *value_iter) == 0) {
+				matched_value = TRUE;
+				break;
+			}
+		}
+
+		if (!matched_value)
+			return FALSE;
+	}
+
+	return TRUE;
+}
+
+static gboolean
 gs_appstream_store_search_item (GsPlugin *plugin,
 				AsApp *item,
 				gchar **values,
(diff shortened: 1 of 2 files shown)


src/gs-discovery-feed-content-provider.c
```diff
@@ -162,13 +162,13 @@ search_done_cb (GObject *source,
...
 	g_autoptr(GsAppList) list = NULL;
 
-	list = gs_plugin_loader_job_process_finish (self->plugin_loader, res, NULL);
+	list = gs_plugin_loader_job_process_finish (self->plugin_loader, res, &error);
 	if (list == NULL) {
-		gs_discovery_feed_installable_apps_complete_get_installable_apps (self->skeleton,
-										  search->invocation,
-										  NULL);
+		g_warning ("Error searching for Discovery Feed Apps: %s", error->message);
+		g_dbus_method_invocation_return_gerror (search->invocation, g_steal_pointer (&error));
```
cosimoc: If you're passing ownership of the error to the method, you should use `g_dbus_method_invocation_take_error()`, since `_return_gerror()` makes a copy.

src/gs-discovery-feed-content-provider.c
```diff
@@ -182,7 +182,6 @@ search_done_cb (GObject *source,
 	 * internally randomized order. */
 	gs_app_list_filter (list, filter_for_discovery_feed_apps, NULL);
 	gs_app_list_randomize (list);
-	gs_app_list_sort (list, search_sort_by_kudo_cb, NULL);
```
cosimoc: The function is unused now, no?

plugins/core/gs-appstream.c
```diff
@@ -828,6 +828,34 @@ gs_appstream_refine_app (GsPlugin *plugin,
...
+
+	for (; *value_iter != NULL; ++value_iter) {
+		GHashTableIter iter;
+		gboolean matched_value = FALSE;
+		gpointer key, value;
+
+		g_hash_table_iter_init (&iter, metadata);
+
+		while (g_hash_table_iter_next (&iter, &key, &value)) {
+			if (g_ascii_strcasecmp ((gchar *) key, *value_iter) == 0) {
```
cosimoc: Why can't you lookup the key directly? Do we really need to use case insensitive match?

src/gs-discovery-feed-content-provider.c
smspillaz: It is! I must have forgotten to remove it.

plugins/core/gs-appstream.c
smspillaz: We do - G-S lowercases all the search terms

src/gs-discovery-feed-content-provider.c
smspillaz: Makes sense

PR #330, opened 2018-02-26
flatpak: Fix setting/removing USB category in cached apps

joaquimrocha:
If an app existed in a USB remote but had been cached (loaded) in GNOME
Software before, then it would never get the USB category set to it.
This resulted in apps not being installable from USB because, by not
having the USB category, and being offline, they would get to the
pending state instead.

The reason for the USB category not being set on previously cached apps
was that we were setting it only to the AsApp objects of the Flatpak
plugin's internal store as the GsApps that are created from these
objects will also get the same categories. However, if the apps had
already been created before, no categories would be set.

This patch fixes that by also setting/removing the USB category to the
apps' corresponding GsApp objects.

https://phabricator.endlessm.com/T17097


mwleeds:
Seems reasonable

PR #333, opened 2018-03-10
T21493

joaquimrocha:
https://phabricator.endlessm.com/T21493

### DIFF
--- a/plugins/eos/gs-plugin-eos.c
+++ b/plugins/eos/gs-plugin-eos.c
@@ -525,7 +525,8 @@ gs_plugin_eos_blacklist_kapp_if_needed (GsPlugin *plugin, GsApp *app)
 }
 
 static gboolean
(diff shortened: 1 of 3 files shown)


plugins/eos/gs-plugin-eos.c
```diff
@@ -773,6 +781,14 @@ gs_plugin_eos_blacklist_upstream_app_if_needed (GsPlugin *plugin, GsApp *app)
 
 	/* We need to check for the app's origin, otherwise we'd be
 	 * blacklisting matching apps coming from any repo */
+	if (g_str_has_suffix (hostname, ".endlessm.com") &&
+	    g_strv_contains (legacy_apps, gs_app_get_id (app))) {
+		g_debug ("Blacklisting '%s': it's a legacy app",
+			 gs_app_get_unique_id (app));
+		do_blacklist = TRUE;
+		goto check_blacklist;
```
rshuler: I'm not a fan of a goto, when I think you could simply convert the following two "if" statements to "else if".

PR #336, opened 2018-03-27
T21133

joaquimrocha:
https://phabricator.endlessm.com/T21133


plugins/eos/gs-plugin-eos.c
```diff
@@ -267,6 +302,243 @@ read_icon_replacement_overrides (GHashTable *replacement_app_lookup)
...
+	case EOS_UPDATER_STATE_NONE:
+	case EOS_UPDATER_STATE_READY:
+		gs_app_set_state (app, AS_APP_STATE_UNKNOWN);
+		break;
+	case EOS_UPDATER_STATE_POLLING:
+		break;
+	case EOS_UPDATER_STATE_UPDATE_AVAILABLE: {
+		guint64 total_size = 0;
+		if (gs_app_get_state (app) == AS_APP_STATE_INSTALLING) {
+			/* when the OS upgrade is already being installed and the
```
cosimoc: How can it happen that the `GsApp` state is "ahead" of the updater state? The comment here mentions polling, but we don't seem to change the `GsApp` state while the updater is polling.

plugins/eos/gs-plugin-eos.c
```diff
@@ -1553,3 +1850,191 @@ gs_plugin_add_popular (GsPlugin *plugin,
...
+	gboolean needs_polling = FALSE;
+	EosUpdaterState eos_updater_state;
+
+	/* only process this app if was created by this plugin */
+	if (g_strcmp0 (gs_app_get_management_plugin (app),
+		       gs_plugin_get_name (plugin)) != 0)
+		return TRUE;
+
+	os_upgrade = get_os_upgrade (plugin);
+	if (app != os_upgrade) {
```
cosimoc: Should we instead assert on this?

plugins/eos/gs-plugin-eos.c
```diff
@@ -1553,3 +1850,191 @@ gs_plugin_add_popular (GsPlugin *plugin,
...
+	}
+
+	os_upgrade_set_download_by_user (app, TRUE);
+
+	/* if it's already updatable (needs reboot) then there's nothing to be
+	 * done here */
+	if (gs_app_get_state (app) == AS_APP_STATE_UPDATABLE)
+		return TRUE;
+
+	gs_app_set_state (app, AS_APP_STATE_INSTALLING);
```
cosimoc: Why do you call this here? I would prefer if all the state management on the `GsApp` instance was done in `sync_state_from_updater()`. Why is changing the updater state not enough?

plugins/eos/gs-plugin-eos.c
```diff
@@ -1553,3 +1850,191 @@ gs_plugin_add_popular (GsPlugin *plugin,
...
+		needs_polling = TRUE;
+	default:
+		break;
+	}
+
+	os_upgrade_set_download_by_user (app, TRUE);
+
+	/* if it's already updatable (needs reboot) then there's nothing to be
+	 * done here */
+	if (gs_app_get_state (app) == AS_APP_STATE_UPDATABLE)
```
cosimoc: I'd prefer if you looked at the updater state here instead (i.e. used the the `UPDATE_APPLIED` state)

plugins/eos/gs-plugin-eos.c
```diff
@@ -267,6 +302,243 @@ read_icon_replacement_overrides (GHashTable *replacement_app_lookup)
...
+		guint64 total_size = 0;
+		if (gs_app_get_state (app) == AS_APP_STATE_INSTALLING) {
+			/* when the OS upgrade is already being installed and the
+			 * updater reports an available update, (meaning we were
+			 * polling before), we should readily call fetch */
+			os_upgrade_force_fetch (priv->updater_proxy);
+			break;
+		}
+
+		total_size = eos_updater_get_download_size (priv->updater_proxy);
```
cosimoc: I would prefer if this code was in an `else` block and you had a single `break` at the end of the `case`

plugins/eos/gs-plugin-eos.c
```diff
@@ -267,6 +302,243 @@ read_icon_replacement_overrides (GHashTable *replacement_app_lookup)
...
+	}
+	case EOS_UPDATER_STATE_FETCHING: {
+		guint64 total_size = 0;
+		guint64 downloaded = 0;
+		gfloat progress = 0;
+
+		/* ensure the state transition to 'installing' is allowed */
+		if (gs_app_get_state (app) != AS_APP_STATE_INSTALLING)
+			gs_app_set_state (app, AS_APP_STATE_AVAILABLE);
+		gs_app_set_state (app, AS_APP_STATE_INSTALLING);
```
cosimoc: Is it because `INSTALLING` can only be reached from `AVAILABLE`? Can you refactor these into an `ensure_installing_state()` function?

plugins/eos/gs-plugin-eos.c
```diff
@@ -267,6 +302,243 @@ read_icon_replacement_overrides (GHashTable *replacement_app_lookup)
...
+		/* if there's an update ready to be deployed and we are not yet
+		 * showing the OS upgrade as downloading (state 'installing'), then
+		 * just show it as available so the user has a change to click 'download'
+		 * which will end up deploying the update */
+		if (gs_app_get_state (app) != AS_APP_STATE_INSTALLING) {
+			gs_app_set_state (app, AS_APP_STATE_AVAILABLE);
+			break;
+		}
+
+		gs_app_set_progress (app, 95);
```
cosimoc: I'd also prefer if this was in an `else` block and there was just a single `break`

cosimoc: But do we also need to set the progress here? Is it not enough to do it later in the `APPLYING_UPDATE` transition?

plugins/eos/gs-plugin-eos.c
```diff
@@ -267,6 +302,243 @@ read_icon_replacement_overrides (GHashTable *replacement_app_lookup)
...
+	default:
+		break;
+	}
+
+	/* ensure that a version (and URL) are set */
+	version = gs_app_get_version (app);
+	if (version == NULL || *version == '\0') {
+		version = eos_updater_get_version (priv->updater_proxy);
+		gs_app_set_version (app, version);
+	}
```
cosimoc: Don't you also need to connect to `notify::version` then?

plugins/eos/gs-plugin-eos.c
```diff
@@ -267,6 +302,243 @@ read_icon_replacement_overrides (GHashTable *replacement_app_lookup)
...
+
+	/* reset the 'download-by-user' state if the the app is no longer
+	 * shown as downloading */
+	if (previous_app_state != AS_APP_STATE_INSTALLING)
+		os_upgrade_set_download_by_user (app, FALSE);
+
+	if (gs_app_get_state (app) == AS_APP_STATE_INSTALLING)
+		setup_os_upgrade_cancellable (plugin);
```
cosimoc: You can fetch `current_app_state` before this block and use that to check against `INSTALLING`

plugins/eos/gs-plugin-eos.c
```diff
@@ -267,6 +302,243 @@ read_icon_replacement_overrides (GHashTable *replacement_app_lookup)
...
+	version = gs_app_get_version (app);
+	if (version == NULL || *version == '\0') {
+		version = eos_updater_get_version (priv->updater_proxy);
+		gs_app_set_version (app, version);
+	}
+
+	/* reset the 'download-by-user' state if the the app is no longer
+	 * shown as downloading */
+	if (previous_app_state != AS_APP_STATE_INSTALLING)
+		os_upgrade_set_download_by_user (app, FALSE);
```
cosimoc: I am not sure the comment and the code agree here; `previous_app_state != AS_APP_STATE_INSTALLING` means that prior to this state change we were not polling, fetching, or installing the update (i.e. we were not showing the app as downloading). Now we may be in one of those states, and we may be showing the app as downloading.

The comment instead suggests that the condition is there to handle the case where we used to show the app as downloading and now we're not anymore, which is the opposite.

(I think that the code is correct and we only want to clear the flag after we stopped downloading/installing.)

plugins/eos/gs-plugin-eos.c
```diff
@@ -267,6 +302,243 @@ read_icon_replacement_overrides (GHashTable *replacement_app_lookup)
...
+	if (value == NULL)
+		return FALSE;
+	return g_variant_get_boolean (value);
+}
+
+static void
+os_upgrade_force_fetch (EosUpdater *updater)
+{
+	g_auto(GVariantDict) options_dict = G_VARIANT_DICT_INIT (NULL);
+	g_variant_dict_insert (&options_dict, "force", "b", TRUE);
```
cosimoc: What does the `force` flag do here, and why do we need it?

plugins/eos/gs-plugin-eos.c
```diff
@@ -51,6 +52,37 @@
...
+					       "Fetching",
+					       "UpdateReady",
+					       "ApplyingUpdate",
+					       "UpdateApplied"};
+
+#define EOS_UPDATER_ERROR_LIVE_BOOT_STR "com.endlessm.Updater.Error.LiveBoot"
+#define EOS_UPDATER_ERROR_NON_OSTREE_STR "com.endlessm.Updater.Error.NotOstreeSystem"
+
+#define EOS_UPGRADE_ID "com.endlessm.EOS.upgrade"
+#define EOS_UPGRADE_BASE_URL "https://www.endlessos.com/release-notes/"
```
cosimoc: Remove this

plugins/eos/gs-plugin-eos.c
```diff
@@ -51,6 +52,37 @@
...
+					       "Error",
+					       "Polling",
+					       "UpdateAvailable",
+					       "Fetching",
+					       "UpdateReady",
+					       "ApplyingUpdate",
+					       "UpdateApplied"};
+
+#define EOS_UPDATER_ERROR_LIVE_BOOT_STR "com.endlessm.Updater.Error.LiveBoot"
+#define EOS_UPDATER_ERROR_NON_OSTREE_STR "com.endlessm.Updater.Error.NotOstreeSystem"
```
cosimoc: These two are unused

plugins/eos/gs-plugin-eos.c
```diff
@@ -1953,6 +1990,10 @@ gs_plugin_add_distro_upgrades (GsPlugin *plugin,
 		return TRUE;
 	}
 
+	/* check if the OS upgrade has been disabled */
+	if (priv->updater_proxy == NULL)
+		return TRUE;
```
cosimoc: This should be moved to before the `get_os_upgrade()` call, or the debug code path will crash as it assumes `priv->updater_proxy != NULL`

plugins/eos/gs-plugin-eos.c
```diff
@@ -2005,6 +2046,13 @@ gs_plugin_app_upgrade_download (GsPlugin *plugin,
 		break;
 	}
 
+	/* if the OS upgrade has been disabled */
+	if (priv->updater_proxy == NULL) {
+		g_set_error (error, GS_PLUGIN_ERROR, GS_PLUGIN_ERROR_FAILED,
+			     "the OS upgrade has been disabled in the EOS plugin");
+		return FALSE;
+	}
```
cosimoc: Ditto, `get_os_upgrade()` will probably crash above when this is `NULL`

plugins/eos/gs-plugin-eos.c
```diff
@@ -369,6 +370,37 @@ os_upgrade_force_fetch_sync (EosUpdater *updater,
...
+
+static void
+disable_os_updater (GsPlugin *plugin)
+{
+	GsPluginData *priv = gs_plugin_get_data (plugin);
+
+	if (priv->upgrade_fake_progress_handler != 0) {
+		g_source_remove (priv->upgrade_fake_progress_handler);
+		priv->upgrade_fake_progress_handler = 0;
+	}
```
cosimoc: This should go in the next commit

plugins/eos/gs-plugin-eos.c
```diff
@@ -369,6 +370,37 @@ os_upgrade_force_fetch_sync (EosUpdater *updater,
...
+
+	if (priv->updater_proxy == NULL)
+		return;
+
+	g_signal_handlers_disconnect_by_func (priv->updater_proxy,
+					      G_CALLBACK (updater_state_changed),
+					      plugin);
+
+	if (priv->os_upgrade_cancellable != NULL) {
+		g_cancellable_cancel (priv->os_upgrade_cancellable);
```
cosimoc: Same comment applies about this function being `NULL-`safe

lib/gs-app-private.h
```diff
@@ -34,7 +34,6 @@ void		 gs_app_set_unique_id		(GsApp		*app,
 						 const gchar	*unique_id);
 void		 gs_app_remove_addon		(GsApp		*app,
 						 GsApp		*addon);
-GCancellable	*gs_app_get_cancellable		(GsApp		*app);
```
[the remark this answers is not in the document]
joaquimrocha: Hey! Sorry for the stray message (it used to be a different one).

I also started by implementing what you're saying: when the upgrade download happens, just keep a ref to the passed cancellable. This doesn't work for two reasons:
1) We plugin-loader passes its own cancellable object to its methods in order to manage a time-out (and it cancels the caller's cancellable when the former emits `cancelled`);
2) Even if we changed the way the plugin loader uses the cancellable, we change the OS upgrade GsApp's state whenever the `eos-updater` changes its state, so we could never enforce that the plugin method would occur before this state change in order to get a ref to the cancellable. i.e. We launch GNOME Software -> the OS upgrade is shown -> the user hasn't clicked `Download` but the auto-updater starts fetching the upgrade -> we set the OS upgrade's state as `installed` (g-s logic) which makes the `Cancel` button be shown; so at this point we wouldn't have a ref for the cancellable.

I will add this information to the commit message.

src/gs-updates-page.c
```diff
@@ -1516,7 +1516,8 @@ gs_updates_page_upgrade_download_cb (GsUpgradeBanner *upgrade_banner,
 
 	if (self->cancellable_upgrade_download != NULL)
 		g_object_unref (self->cancellable_upgrade_download);
-	self->cancellable_upgrade_download = g_cancellable_new ();
+	self->cancellable_upgrade_download = g_object_ref (gs_app_get_cancellable (app));
```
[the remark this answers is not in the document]
joaquimrocha: I didn't explain the reason a lot here because we already use a similar logic for all apps' operations. The upgrade is no exception and should use the GsApp object's cancellable too. I will send this patch upstream too.

plugins/eos/gs-plugin-eos.c
joaquimrocha: The normal `Fetch()` method will check whether `libmogwai` allows it to perform the download or not. Since we only call fetch ourselves when the user clicks `Download`, I think we should force the download (hence the flag). Any verification (like a dialog confirming that the user wants to proceed even on a metered connection) should be done outside of this plugin.

plugins/eos/gs-plugin-eos.c
joaquimrocha: In `gs_plugin_app_upgrade_download` we set the state to `installing` + initiate the `Poll()` when needed. I wrote this before I needed the `eos::DownloadByUser` metadata which can also be used for the same effect. Still, this was well tested and also means that if we drop or set that metadata key in a different way, then we don't need to change it here.

plugins/eos/gs-plugin-eos.c
joaquimrocha: I did it like this because we use early returns everywhere, and it's the same principle. But I can change it...

plugins/eos/gs-plugin-eos.c
joaquimrocha: Yes. There are "safeguard" checks in the state transactions.
This logic is in only 2 places, so I thought I'd keep it like this. A more efficient force function would assign the state to `NULL` and then to whatever state we want, as that transaction is always allowed. But we do get a log about changing the state to `NULL` which can be confusing in the logs, so I preferred not to do that this time (I think I had done that before in other parts of the code) just for these cases + the one that keeps the `available` state.

plugins/eos/gs-plugin-eos.c
joaquimrocha: This was more like ensuring we are in that progress than anything, but yeah, I think it's safe to remove it.

plugins/eos/gs-plugin-eos.c
joaquimrocha: Hmm, good point. it shouldn't matter most of the times but if we have the GsApp object for the updater and a new upgrade becomes available, this wouldn't update the version. I will change that.

plugins/eos/gs-plugin-eos.c
joaquimrocha: The `previous_app_state` is wrong here. The `os_upgrade_set_download_by_user` sets a metadata key that tells us that the download has been started by the user. When the app changes to not-downloading, then we need to reset that. Since this is used to decide whether to show warnings on errors, it was still working for that because we do set the installing state before polling/fetching, so the previous state will always be `installing` and the `download-by-user` will keep being `TRUE`.
So the correct state to check is the current (I don't think this is what you said) and I will change it.

plugins/eos/gs-plugin-eos.c
joaquimrocha: Thanks for spotting that!

plugins/eos/gs-plugin-eos.c
```diff
@@ -319,10 +591,25 @@ gs_plugin_setup (GsPlugin *plugin,
...
 
+	g_assert (G_N_ELEMENTS (eos_updater_state_str) == EOS_UPDATER_N_STATES);
+
 	/* Synchronous, but this guarantees that the lookup table will be
 	 * there when we call ReplaceApplication later on */
 	read_icon_replacement_overrides (priv->replacement_app_lookup);
 
+	{
+		g_autoptr(GError) local_error = NULL;
+		priv->updater_proxy = eos_updater_proxy_new_for_bus_sync (G_BUS_TYPE_SYSTEM,
```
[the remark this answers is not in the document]
joaquimrocha: I wanted to make sure we have the object before a `refresh` happens. Maybe this may still be possible by doing it asynchronously, but I didn't experience any noticeable delay, and prefer to make sure we have the proxy from the beginning.
Since GNOME Software is started in the background, even if this delays things a bit, it shouldn't be noticeable to the user. I should get a better insight when I work on improving the performance, but for now I feel safer by leaving this as is.

plugins/eos/gs-plugin-eos.c
```diff
@@ -351,10 +638,17 @@ gs_plugin_destroy (GsPlugin *plugin)
 		priv->applications_changed_id = 0;
 	}
 
+	if (priv->os_upgrade_cancellable != NULL) {
+		g_cancellable_cancel (priv->os_upgrade_cancellable);
```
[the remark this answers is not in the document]
joaquimrocha: I didn't remember that. Thanks!

plugins/eos/gs-plugin-eos.c
```diff
@@ -1553,3 +1873,191 @@ gs_plugin_add_popular (GsPlugin *plugin,
...
+
+	priv->os_upgrade = app;
+
+	return app;
+}
+
+static gboolean
+should_add_os_upgrade (GsApp *os_upgrade)
+{
+	switch (gs_app_get_state (os_upgrade)) {
```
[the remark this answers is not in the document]
joaquimrocha: Most code that deals with the state changes in g-s uses switch so I tried to make it closer to that. But I can change it in this case.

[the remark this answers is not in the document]
joaquimrocha: Actually I changed my mind and prefer to keep the switch as after changing it, I realize that for this logic it just seems easier to read and to extend if needed later...

plugins/eos/gs-plugin-eos.c
joaquimrocha: I had an assert but changed it before pushing the PR because when the code is released we should ensure that we get some log if this happens + not let the logic proceed.

plugins/eos/gs-plugin-eos.c
```diff
@@ -1553,3 +1873,191 @@ gs_plugin_add_popular (GsPlugin *plugin,
...
+		g_warning ("The OS upgrade to download (%s) differs from the "
+			   "one in the EOS plugin, yet it's managed by it!",
+			   gs_app_get_unique_id (app));
+		return TRUE;
+	}
+
+	eos_updater_state = sync_state_from_updater (plugin);
+	/* we need to poll again when there has been an error or it's one of the
+	 * initial states */
+	switch (eos_updater_state) {
```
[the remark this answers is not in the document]
joaquimrocha: Again, usually I don't love switch statements but maybe I got used to dealing with GsApp's states in switch statements

plugins/eos/gs-plugin-eos.c
joaquimrocha: Should be the same ATM. But if we change the states' logic to only download, without applying the update, then we'd need to change this logic again. Even though I needed to check the updater state above, I tried to keep those mostly in the `sync_state_from_updater`. Is there a problem with keeping it like this that I may be missing?

plugins/eos/gs-plugin-eos.c
```diff
@@ -1553,3 +1873,191 @@ gs_plugin_add_popular (GsPlugin *plugin,
...
+	if (gs_app_get_state (app) == AS_APP_STATE_UPDATABLE)
+		return TRUE;
+
+	gs_app_set_state (app, AS_APP_STATE_INSTALLING);
+
+	/* poll or fetch the update, and the state of the OS upgrade will be
+	 * dealt with from outside this function, according to the state changes
+	 * of the update itself */
+	if (needs_polling) {
+		if (!eos_updater_call_poll_sync (priv->updater_proxy, cancellable,
```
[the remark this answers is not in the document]
joaquimrocha: Because the `polling` state doesn't change the state, so we need to start the change here since we know that the user is explicitly downloading it.
I added the `download-by-user` metadata after most of this logic so I didn't change what I had. But I can use this metadata in the `sync_state_from_updater` if this is important.

plugins/eos/gs-plugin-eos.c
joaquimrocha: Because the `polling` state doesn't change the state, so we need to start the change here since we know that the user is explicitly downloading it.
I added the `download-by-user` metadata after most of this logic so I didn't change what I had. But I can use this metadata in the `sync_state_from_updater` if this is important.

plugins/eos/gs-plugin-eos.c
joaquimrocha: Oops! Thanks.

plugins/eos/gs-plugin-eos.c
joaquimrocha: The `get_os_upgrade()` runs after we create the proxy and only creates the GsApp for the upgrade once, so it won't crash as you say.
However, I was counting on the proxy only being `NULL` after we disable the upgrades. But if it fails to create the proxy for some reason, then yes, it would crash. I will fix that.

plugins/eos/gs-plugin-eos.c
joaquimrocha: See my reply above.

src/gs-updates-page.c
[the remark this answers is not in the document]
cosimoc: OK

plugins/eos/gs-plugin-eos.c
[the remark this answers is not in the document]
cosimoc: Same comment as the other switch below, but I won't insist if you want to keep it this way :)

plugins/eos/gs-plugin-eos.c
```diff
@@ -267,6 +310,330 @@ read_icon_replacement_overrides (GHashTable *replacement_app_lookup)
...
+	} else {
+		/* otherwise, ensure we have the right cancellable*/
+		setup_os_upgrade_cancellable (plugin);
+	}
+
+	/* if the state changed from or to 'unknown', we need to notify that a
+	 * new update should be shown */
+	if ((previous_app_state == AS_APP_STATE_UNKNOWN ||
+	     current_app_state == AS_APP_STATE_UNKNOWN) &&
+	    previous_app_state != current_app_state)
```
cosimoc: I wonder if the last line is really required; I don't think it's possible to get all the way here with the state unchanged and still UNKNOWN

plugins/eos/gs-plugin-eos.c
```diff
@@ -267,6 +310,330 @@ read_icon_replacement_overrides (GHashTable *replacement_app_lookup)
...
+					      plugin);
+
+	g_cancellable_cancel (priv->os_upgrade_cancellable);
+	g_clear_object (&priv->os_upgrade_cancellable);
+
+	g_clear_object (&priv->updater_proxy);
+}
+
+static gboolean
+fake_os_upgrade_progress (GsPlugin *plugin)
```
cosimoc: I am not a huge fan of faking progress during installation of files; ideally we could have a different metaphor to use, such as a spinner, in those cases. But I understand why you did it, so we can keep it like this for now; perhaps something good to bring up with upstream though.

plugins/eos/gs-plugin-eos.c
```diff
@@ -267,6 +310,330 @@ read_icon_replacement_overrides (GHashTable *replacement_app_lookup)
...
+	if (gs_app_get_state (app) != AS_APP_STATE_INSTALLING)
+		gs_app_set_state (app, AS_APP_STATE_AVAILABLE);
+
+	gs_app_set_state (app, AS_APP_STATE_INSTALLING);
+}
+
+static gboolean
+eos_updater_error_is_cancelled (const gchar *error_name)
+{
+	return (g_strcmp0 (error_name, "com.endlessm.Updater.Error.Cancelled") != 0);
```
cosimoc: This is check for the wrong condition; should be `== 0` to match the cancelled error.

plugins/eos/gs-plugin-eos.c
```diff
@@ -267,6 +310,330 @@ read_icon_replacement_overrides (GHashTable *replacement_app_lookup)
...
+	}
+
+	current_app_state = gs_app_get_state (app);
+
+	/* reset the 'download-by-user' state if the the app is no longer
+	 * shown as downloading */
+	if (current_app_state != AS_APP_STATE_INSTALLING) {
+		os_upgrade_set_download_by_user (app, FALSE);
+	} else {
+		/* otherwise, ensure we have the right cancellable*/
```
cosimoc: Missing space before the end of comment

plugins/eos/gs-plugin-eos.c
joaquimrocha: Arggh. I had if/else blocks checking for "not cancelled" first, but then changed this last minute to be a one-liner... and screwed up setting the condition.

plugins/eos/gs-plugin-eos.c
joaquimrocha: I agree with you. Let's leave this for now and follow a better solution with upstream as this may happen there too.

plugins/eos/gs-plugin-eos.c
joaquimrocha: If it's an initial state (app state = `UNKNOWN`) **and** the EOS updater reports state `NONE`, `READY`, or `POLLING` (without user intervention), the state of the app will not change, and we don't want the updates to be gathered again. Without it, since `sync_state_from_updater` is also called directly in `gs_plugin_add_distro_upgrades` (not only in the state-change callback), we could reach a state where it's calls the check for updates sequentially. It's not a likely thing to happen, but still it's good to have this safeguard IMO.

plugins/eos/gs-plugin-eos.c
joaquimrocha: I still think this should be fine as it's checking the app state, not changing it. But I will change it to include more states in the switch check above it.

plugins/eos/gs-plugin-eos.c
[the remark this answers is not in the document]
joaquimrocha: I ended up adding a few more conditions after your comment that you didn't like the check for the app's state.

plugins/eos/gs-plugin-eos.c
```diff
@@ -1553,3 +1934,177 @@ gs_plugin_add_popular (GsPlugin *plugin,
...
+	return TRUE;
+}
+
+gboolean
+gs_plugin_app_upgrade_download (GsPlugin *plugin,
+				GsApp *app,
+			        GCancellable *cancellable,
+				GError **error)
+{
+	g_autoptr(GError) local_error = NULL;
```
cosimoc: Unused now

PR #337, opened 2018-03-28
T21424

joaquimrocha:
https://phabricator.endlessm.com/T21424

### DIFF
--- a/src/gs-app-row.c
+++ b/src/gs-app-row.c
@@ -699,7 +699,6 @@ gs_app_row_init (GsAppRow *app_row)
 	gtk_widget_init_template (GTK_WIDGET (app_row));
 	gs_star_widget_set_icon_size (GS_STAR_WIDGET (priv->star), 12);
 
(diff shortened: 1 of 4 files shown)


src/gs-app-row.c
```diff
@@ -255,11 +255,11 @@ gs_app_row_refresh_button (GsAppRow *app_row, gboolean missing_search_result)
 		gtk_style_context_remove_class (context, "destructive-action");
 	} else {
 		switch (gs_app_get_state (priv->app)) {
-		case AS_APP_STATE_UPDATABLE:
 		case AS_APP_STATE_INSTALLED:
-		case AS_APP_STATE_UPDATABLE_LIVE:
 			gtk_style_context_add_class (context, "destructive-action");
 			break;
+		case AS_APP_STATE_UPDATABLE:
```
rshuler: These lines are not strictly needed since the `default` case will catch them anyhow, but I suppose it's okay if you want to be explicit.

PR #339, opened 2018-04-05
Remove the big shopping bag icon from the loading page

joaquimrocha:
The big shopping bag icon in the loading page looks very similar to a
lockpad and this can be confusing for some users. Since we will also
replace the shopping bag icons in the app tiles, the big icon in the
loading page should also be replaced or removed. So for now, we are
removing it.

https://phabricator.endlessm.com/T21925


joaquimrocha:
Merged the PR after @mariospr confirmed it looks okay to him.

PR #340, opened 2018-04-09
details-page: Handle Escape key

uajain:
https://phabricator.endlessm.com/T13085


src/gs-details-page.c
```diff
@@ -2265,6 +2265,16 @@ gs_details_page_network_available_notify_cb (GsPluginLoader *plugin_loader,
 	gs_details_page_refresh_reviews (self);
 }
 
+static void
+gs_details_page_key_pressed_cb (GtkWidget *widget, GdkEventKey *key, GsDetailsPage *self)
+{
+	g_return_if_fail (key != NULL);
+
+	if (key->keyval == GDK_KEY_Escape) {
```
cosimoc: I believe gnome-software uses no curly braces for single-line condition blocks.

src/gs-details-page.c
```diff
@@ -2265,6 +2265,16 @@ gs_details_page_network_available_notify_cb (GsPluginLoader *plugin_loader,
 	gs_details_page_refresh_reviews (self);
 }
 
+static void
```
cosimoc: These callbacks need to return `FALSE` (or `GDK_EVENT_PROPAGATE`/however it's called)

src/gs-details-page.c
```diff
@@ -2265,6 +2265,16 @@ gs_details_page_network_available_notify_cb (GsPluginLoader *plugin_loader,
...
+gs_details_page_key_pressed_cb (GtkWidget *widget, GdkEventKey *key, GsDetailsPage *self)
+{
+	g_return_if_fail (key != NULL);
```
cosimoc: I don't think this condition here is needed; `g_return_if_fail()` are used for programmer errors, but if we arrive here with `key == NULL` it's a bug in gnome-software and nothing that the caller of this function can do about.

src/gs-details-page.c
```diff
@@ -2343,6 +2353,9 @@ gs_details_page_setup (GsPage *page,
 	g_signal_connect (self->label_license_nonfree_details, "activate-link",
 			  G_CALLBACK (gs_details_page_activate_link_cb),
 			  self);
+	g_signal_connect (GTK_WIDGET (self), "key-press-event",
```
cosimoc: `g_signal_connect()` takes a `gpointer` so you don't need any cast here.

src/gs-details-page.c
joaquimrocha: No hard rule AFAIK. But yeah, let's try to keep it simpler without the brackets.

PR #341, opened 2018-04-12
T20936: simplify logic for EknServicesMultiplexer & auto-upgrade

ramcq:
We now have one EknServicesMultiplexer which is required to be installed
whenever we have a knowledge app installed, rather than the previous versioned
EknServices that were dependent on the SDK version being used. This multiplexer
should be installed by the OS, but we want to also pull it in for an upgrade
whenever we're installing a knowledge app as it could be using a new SDK that's
not supported by the old multiplexer.

https://phabricator.endlessm.com/T20936

### DIFF
--- a/plugins/flatpak/gs-flatpak.c
+++ b/plugins/flatpak/gs-flatpak.c
@@ -2923,18 +2923,31 @@ gs_flatpak_refine_runtime_for_install (GsFlatpak *self,
 
 static void
 gs_flatpak_add_app_to_list_if_not_installed (GsApp *app, GsAppList *list,
-					     const GHashTable *hash_installed)
+					     const GHashTable *hash_installed,
+					     gboolean include_updates)
 {
 	g_autofree gchar *ref_display = gs_flatpak_app_get_ref_display (app);
-	if (g_hash_table_contains (hash_installed, ref_display)) {
-		g_debug ("%s is already installed, so skipping",
-			 gs_app_get_unique_id (app));
+	gboolean is_installed = g_hash_table_contains (hash_installed, ref_display);
+	if (is_installed &&
+	    (!include_updates || !gs_app_is_updatable (app))) {
+		if (include_updates)
+			g_debug ("%s is already installed and up to date, so skipping",
+				 gs_app_get_unique_id (app));
+		else
+			g_debug ("%s is already installed, so skipping",
+				 gs_app_get_unique_id (app));
 		return;
 	}
 
-	g_debug ("%s/%s is not already installed, so installing",
-		 gs_flatpak_app_get_ref_name (app),
-		 gs_flatpak_app_get_ref_branch (app));
+	if (is_installed)
+		g_debug ("%s/%s needs update, so updating",
+			 gs_flatpak_app_get_ref_name (app),
+			 gs_flatpak_app_get_ref_branch (app));
+	else
+		g_debug ("%s/%s is not already installed, so installing",
+			 gs_flatpak_app_get_ref_name (app),
+			 gs_flatpak_app_get_ref_branch (app));
+
 	gs_app_list_add (list, app);
 }
 
@@ -2982,14 +2995,15 @@ gs_flatpak_get_list_for_install_or_update (GsFlatpak *self,
 		runtime = gs_app_get_update_runtime (app);
 	}
 	if (runtime != NULL)
-		gs_flatpak_add_app_to_list_if_not_installed (runtime, list, hash_installed);
+		gs_flatpak_add_app_to_list_if_not_installed (runtime, list, hash_installed, FALSE);
 
 	/* add services flatpak */
 	services_app = gs_flatpak_get_services_app_if_needed (self, app, runtime, cancellable);
 	if (g_cancellable_set_error_if_cancelled (cancellable, error))
 		return FALSE;
 	if (services_app != NULL) {
-		gs_flatpak_add_app_to_list_if_not_installed (services_app, list, hash_installed);
+		/* ensure the app is installed and up to date */
+		gs_flatpak_add_app_to_list_if_not_installed (services_app, list, hash_installed, TRUE);
 
 		/* add services flatpak's runtime, if different from app's runtime */
 		if (is_repair) {
@@ -3001,7 +3015,7 @@ gs_flatpak_get_list_for_install_or_update (GsFlatpak *self,
 			services_runtime = gs_app_get_update_runtime (services_app);
 		}
 		if (services_runtime != NULL)
-			gs_flatpak_add_app_to_list_if_not_installed (services_runtime, list, hash_installed);
+			gs_flatpak_add_app_to_list_if_not_installed (services_runtime, list, hash_installed, FALSE);
 	}
 
 	/* lookup any related refs for this ref */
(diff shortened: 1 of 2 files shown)


ramcq:
It builds! Ship it. (Actually: untested, someone please take this over.)

plugins/flatpak/gs-flatpak.c
```diff
@@ -2891,13 +2891,14 @@ gs_flatpak_add_app_to_list_if_not_installed (GsApp *app, GsAppList *list,
 					     const GHashTable *hash_installed)
 {
 	g_autofree gchar *ref_display = gs_flatpak_app_get_ref_display (app);
-	if (g_hash_table_contains (hash_installed, ref_display)) {
-		g_debug ("%s is already installed, so skipping",
+	if (g_hash_table_contains (hash_installed, ref_display) &&
+	    !gs_app_is_updatable (app)) {
```
joaquimrocha: I am not sure about this.
Does this mean that any installation/update may try to update the runtimes and services app too? Before we used to keep the update only for the proxy app, so maybe it's better to keep it that way until we revise this when we start adding the "size" as a criteria for Mogwai.

In any case, such a change should also change how the function is called as it's now misleading.

This code if kind of critical/delicate as it may stop installations and updates from working, so I prefer to be conservative here. Thus, if possible let's remove this part and open a task for the future if runtimes and services should be updated together with the apps.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2675,50 +2675,45 @@ gs_flatpak_app_remove_source (GsFlatpak *self,
...
-			services_branch = "master";
-		else
-			services_branch = "stable";
-	} else {
-		/* Runtime doesn't require an EknServices app */
-		return NULL;
+
+	if (g_strcmp0 (runtime_id, "com.endlessm.apps.Platform") == 0) {
+		needed = TRUE;
+	} else if (g_strcmp0 (runtime_id, "com.endlessm.Platform") == 0 &&
```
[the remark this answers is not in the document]
ramcq: There are certainly endlessm.com apps like photos and hatch previewer which are not knowledge apps.

plugins/flatpak/gs-flatpak.c
ramcq: At the moment, flatpak on the CLI always upgrades any related refs and the runtimes whenever you install or upgrade an app. Does GNOME software not do this? It could cause subtle bugs (for instance LibreOffice 6 needs some settings schemas for HiDPI in Wayland which were only added more recently to FD.o).

Anyway, this function is Endless specific, it's only called in the code that considers adding the SDK to the list of apps to install.

In any case, if we don't upgrade EKSM, this PR isn't worthwhile - the OS makes it nearly impossible for us to not have EKSM installed. The situation we want to avoid is you install your first SDK4 app but your EKSM is old and doesn't know how to access those files. That's the primary goal of keeping this code in GS - otherwise we could drop it all.

joaquimrocha: g-s does update related apps together with other apps. But it doesn't update runtimes when updating an app (if there's a new runtime/runtime-branch needed for the new app update, then it's installed though). Just to be clear, I am fine if we need to update the services app (which may be relatively small) together with knowledge apps, but your patch affects also runtimes, which I am not sure should be treated that way here.
We have some logic in the auto updater to consider app updates first, if now every app ends up updating the runtimes too, that logic doesn't make much sense (besides requiring more testing, etc.).
All this makes me feel safer if we leave the runtimes outside of this ATM (after the release we can discuss if this is a good thing with upstream as well).

plugins/flatpak/gs-flatpak.c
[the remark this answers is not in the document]
joaquimrocha: So those apps that are not knowledge apps do not use the Endless' runtime? I'm just trying to figure out if checking both runtime + app is necessary or not, as checking just the app makes it simpler.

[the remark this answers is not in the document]
ramcq: Checking the runtime is the correct thing to do for SDK1 or later. We have an OSS SDK precisely for other people making apps. Checking the app ID to disambiguate the old runtime is a heuristic but considering this is a fixed legacy set of apps, not a bad one. :)

plugins/flatpak/gs-flatpak.c
ramcq: Oh interesting. The dangers of mobile github. I saw in Philip's old commit that this function was new but didn't catch it was still used in the runtime case (which it was factored out of). So yeah, I hadn't intended to change the runtime behaviour indeed, that's a separate debate. Arguably not necessary for flatpak runtimes as a branch is meant to be an ABI... but... well. We should fix this so it only updates EKSM in the knowledge app case. For a quick fix we can just drop this code and it will be OK until SDK5 appears. :)

plugins/flatpak/gs-flatpak.c
```diff
@@ -2924,16 +2924,23 @@ gs_flatpak_refine_runtime_for_install (GsFlatpak *self,
...
+					     const GHashTable *hash_installed,
+					     gboolean include_updates)
 {
 	g_autofree gchar *ref_display = gs_flatpak_app_get_ref_display (app);
-	if (g_hash_table_contains (hash_installed, ref_display)) {
-		g_debug ("%s is already installed, so skipping",
-			 gs_app_get_unique_id (app));
+	if (g_hash_table_contains (hash_installed, ref_display) &&
+	    (!include_updates || !gs_app_is_updatable (app))) {
+		if (include_updates) {
```
joaquimrocha: No need for curly brackets.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2924,16 +2924,23 @@ gs_flatpak_refine_runtime_for_install (GsFlatpak *self,
...
+			g_debug ("%s is already installed and up to date, so skipping",
+				 gs_app_get_unique_id (app));
+		} else {
```
joaquimrocha: No need for curly brackets.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2924,16 +2924,23 @@ gs_flatpak_refine_runtime_for_install (GsFlatpak *self,
...
+			g_debug ("%s is already installed, so skipping",
+				 gs_app_get_unique_id (app));
+		}
 		return;
 	}
 
-	g_debug ("%s/%s is not already installed, so installing",
+	g_debug ("%s/%s is not already installed or needs update, so installing",
```
joaquimrocha: I think we should either divide this message, or change it further so it doesn't say `installing`.
Perhaps:
`is not already installed/up-to-date, so adding to the list to be installed/updated`

Yeah, maybe it's better to split it.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2710,51 +2710,47 @@ gs_flatpak_app_remove_source (GsFlatpak *self,
...
-		services_id = g_strdup ("com.endlessm.EknServices2");
-		if (g_strcmp0 (runtime_branch, "master") == 0)
-			services_branch = "master";
-		else
-			services_branch = "stable";
-	} else {
-		/* Runtime doesn't require an EknServices app */
-		return NULL;
+
+	if (g_strcmp0 (runtime_id, "com.endlessm.apps.Platform") == 0) {
```
rshuler: I suspect that @joaquimrocha would prefer not to have curly braces here, but I'm not going to hold up this ticket for it.  @joaquimrocha please go ahead and push a style cleanup commit if you like.

rshuler: Actually, I'm going to rebase the commits anyhow to fix the capitalization of the commit messages as requested by @joaquimrocha, so I'll clean up these curly braces, too.

PR #342, opened 2018-04-13
T20402

joaquimrocha:
https://phabricator.endlessm.com/T20402


mariospr:
FWIW, the build failure is strictly related to the lack of the mogwai build-dep at the time Jenkins tried to build it, which is fixed by https://github.com/endlessm/gnome-software/pull/343, so I will ignore it for review purposes.

src/gs-update-monitor.c
```diff
@@ -150,6 +190,269 @@ no_updates_for_a_week (GsUpdateMonitor *monitor)
...
+	parameters = g_variant_ref_sink (g_variant_dict_end (&parameters_dict));
+
+	if (!setup_scheduler (monitor, cancellable, &local_error)) {
+		g_warning ("Could not schedule update for app %s: %s",
+			   app_id, local_error->message);
+		return;
+	}
+
+	if (helper == NULL) {
+		helper = g_new0 (UpdateScheduleHelper, 1);
```
mariospr: I'd use `g_slice_new0` here, but maybe my knowledge is outdated and it's not recommended anymore?

src/gs-update-monitor.c
```diff
@@ -48,10 +52,46 @@ struct _GsUpdateMonitor {
...
+						     helper->download_now_handler_id);
+		if (helper->invalidate_handler_id > 0)
+			g_signal_handler_disconnect (helper->entry,
+						     helper->invalidate_handler_id);
+		mwsc_schedule_entry_remove_async (helper->entry, NULL, NULL, NULL);
+		g_clear_object (&helper->entry);
+	}
+
+	g_object_unref (helper->app);
+	g_free (helper);
```
mariospr: Remember to use `g_slice_free` if you have moved to `g_slice_new0` below.

src/gs-update-monitor.c
```diff
@@ -150,6 +190,269 @@ no_updates_for_a_week (GsUpdateMonitor *monitor)
...
+		helper->app = g_object_ref (app);
+	}
+
+	mwsc_scheduler_schedule_async (monitor->scheduler, parameters, cancellable,
+				       (GAsyncReadyCallback) async_result_cb,
+				       &new_result);
+
+	context = g_main_context_default ();
+	while (new_result == NULL)
+		g_main_context_iteration (context, TRUE);
```
mariospr: This doesn't feel right: you're manually spinning the main loop of the global main context which will potentially block things while `mwsc_scheduler_schedule_async` is doing its work, right?

I think it would be better to either move the logic below into `async_result_cb` or, if you really want to do this thing of manually spinning the main loop, to make sure this is executed from a separate thread (and then probably use `g_main_context_get_thread_default` instead).

src/gs-update-monitor.c
```diff
@@ -200,6 +512,12 @@ get_updates_finished_cb (GObject *object,
 	    no_updates_for_a_week (monitor)) {
 		notify_offline_update_available (monitor);
 	}
+
+	/* try to update any apps that should be updated already, and connect to
+	 * MwscScheduleEntry signals; this is done here because now all the updates
+	 * have been scheduled, otherwise we would risk starting an update only
+	 * for it to be canceled if a higher priority app was added */
+	finish_scheduling_updates (monitor);
```
mariospr: You'll have to move this somewhere else if you implement `schedule_update` in an asynchronous way (instead of manually spinning the main loop, as you do now)

src/gs-update-monitor.c
```diff
@@ -150,6 +190,269 @@ no_updates_for_a_week (GsUpdateMonitor *monitor)
...
+	if (helper->entry == NULL) {
+		g_warning ("Failed to get schedule entry for updating app %s: %s",
+			   app_id, local_error->message);
+		return;
+	}
+
+	g_debug ("Scheduling new update for app %s with entry id %s", app_id,
+		 mwsc_schedule_entry_get_id (helper->entry));
+	g_hash_table_insert (monitor->scheduled_updates, g_strdup (app_id),
+			     g_steal_pointer (&helper));
```
mariospr: `helper` is a local variable and not a auto pointer to `UpdateScheduleHelper`, so you don't need to use `g_steal_pointer` here as it won't make any difference to set it to `NULL` before exiting the function's scope.

src/gs-update-monitor.c
```diff
@@ -150,6 +190,269 @@ no_updates_for_a_week (GsUpdateMonitor *monitor)
...
+	if (monitor->scheduler != NULL)
+		return TRUE;
+
+	context = g_main_context_default ();
+
+	mwsc_scheduler_new_async (cancellable,
+				  (GAsyncReadyCallback) async_result_cb,
+				  &new_result);
+	while (new_result == NULL)
+		g_main_context_iteration (context, TRUE);
```
mariospr: Same comment to the one I made on `schedule_update` here: I think this is wrong since you're manually spinning the main loop of the global main context manually. Either properly deal with asynchrony here or move it into its own thread (and probably use `g_main_context_get_thread_default`)

src/gs-update-monitor.c
```diff
@@ -768,32 +770,26 @@ check_updates (GsUpdateMonitor *monitor)
...
 	if (last_refreshed != NULL) {
-		gint now_year, now_month, now_day, now_hour;
-		gint year, month, day;
-		g_autoptr(GDateTime) now = NULL;
+		GTimeSpan time_passed;
+		guint time_passed_secs;
 
 		now = g_date_time_new_now_local ();
+		time_passed = g_date_time_difference (now, last_refreshed);
+		time_passed_secs = (guint) (time_passed / 1000000); /* usec to sec */
```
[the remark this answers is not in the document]
pwithnall: `time_passed` and `time_passed_secs` need to be signed here, as a DST transition or system clock change could cause `time_passed` to be negative.

src/gs-update-monitor.c
```diff
@@ -450,32 +770,26 @@ check_updates (GsUpdateMonitor *monitor)
...
+				 time_passed_secs);
 			return;
+		}
 	}
 
-	g_debug ("Daily update check due");
-	now_refreshed = g_date_time_new_now_local ();
+	g_debug ("Updates check due");
+	if (now == NULL)
+		now = g_date_time_new_now_local ();
```
pwithnall: Seems like you should just unconditionally assign `now` when it’s declared, because you assign it in all branches afterwards.

src/gs-update-monitor.c
```diff
@@ -876,6 +1199,11 @@ gs_update_monitor_dispose (GObject *object)
 						      monitor);
 		monitor->plugin_loader = NULL;
 	}
+
+	g_cancellable_cancel (monitor->scheduled_updates_cancellable);
+	g_clear_object (&monitor->scheduled_updates_cancellable);
```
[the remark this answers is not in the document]
pwithnall: Not necessary: `g_cancellable_cancel (NULL)` is defined as a safe no-op, and `g_clear_object()` checks for `NULL` before unreffing anything.

src/gs-update-monitor.c
```diff
@@ -150,6 +190,269 @@ no_updates_for_a_week (GsUpdateMonitor *monitor)
...
+
+static void
+finish_scheduling_updates (GsUpdateMonitor *monitor)
+{
+	GHashTableIter iter;
+	gpointer value;
+
+	g_hash_table_iter_init (&iter, monitor->scheduled_updates);
+	while (g_hash_table_iter_next (&iter, NULL, &value)) {
+		UpdateScheduleHelper *helper = value;
```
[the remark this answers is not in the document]
pwithnall: Not needed: `value` is a `gpointer`.

src/gs-update-monitor.c
```diff
@@ -150,6 +192,269 @@ no_updates_for_a_week (GsUpdateMonitor *monitor)
...
+	const gchar *app_id = gs_app_get_unique_id (helper->app);
+
+	g_debug ("Removing scheduled update of app %s", app_id);
+	g_hash_table_remove (helper->monitor->scheduled_updates, app_id);
+}
+
+static void
+schedule_update (GsUpdateMonitor *monitor,
+		 GsApp *app,
+		 GCancellable *cancellable)
```
pwithnall: Nitpick: Alignment problems with the arguments here?

src/gs-update-monitor.c
pwithnall: There’s not much of a performance difference any more. I use `g_new0()` because the API is a little nicer.

src/gs-update-monitor.c
pwithnall: If you’re manually iterating a main context, make sure it’s a new `GMainContext` instance specifically for this operation (and temporarily pushed onto the thread default stack). No need to run it from a separate thread.

src/gs-update-monitor.c
```diff
@@ -150,6 +192,269 @@ no_updates_for_a_week (GsUpdateMonitor *monitor)
...
+	GVariant *tmp = gs_app_get_metadata_variant (app, APP_METADATA_AUTO_UPDATING);
+	if (tmp == NULL)
+		return FALSE;
+	return g_variant_get_boolean (tmp);
+}
+
+static void
+app_update_finished_cb (GObject *source,
+			GAsyncResult *res,
+			UpdateScheduleHelper *helper)
```
pwithnall: Nitpick: Misalignment on the arguments?

pwithnall: Another nitpick: I find it slightly more readable to put the `GAsyncReadyCallback` function below the `_async` call — so I’d put `app_update_finished_cb()` immediately below the end of `update_app()`. Then you can read through the code for `update_app()`, the `_async()` call`, and then immediately read through the code in the async callback which is next in the control flow.

The downside to that is that it does require you to declare `app_update_finished_cb()` in advance.

src/gs-update-monitor.c
```diff
@@ -150,6 +192,269 @@ no_updates_for_a_week (GsUpdateMonitor *monitor)
 	return FALSE;
 }
 
+static void
+app_set_auto_updating (GsApp *app,
+		       gboolean auto_updating)
+{
+	/* we always have to set it to NULL as otherwise GsApp doesn't allow
+	 * the metadata to be overridden */
```
pwithnall: This need to always reset to `NULL` before setting anything seems like a flaw in `GsApp`. Worth fixing?

src/gs-update-monitor.c
```diff
@@ -48,10 +54,46 @@ struct _GsUpdateMonitor {
...
+			 gs_app_get_unique_id (helper->app),
+			 mwsc_schedule_entry_get_id (helper->entry));
+
+		if (helper->download_now_handler_id > 0)
+			g_signal_handler_disconnect (helper->entry,
+						     helper->download_now_handler_id);
+		if (helper->invalidate_handler_id > 0)
+			g_signal_handler_disconnect (helper->entry,
+						     helper->invalidate_handler_id);
+		mwsc_schedule_entry_remove_async (helper->entry, NULL, NULL, NULL);
```
pwithnall: I’m a little uneasy about firing off an async function like this without tracking whether it completes. Does the relevant `GMainContext` ever get iterated? Mogwai will automatically remove its schedule entries if gnome-software drops off D-Bus (i.e. when it exits), but not before then.

joaquimrocha: I don't understand. I am not interested in the result of this function as, if it fails, there's little value in printing a warning. But if there's something I am missing, I can add the callback...

src/gs-update-monitor.c
joaquimrocha: @mariospr I think it's no longer an advantage to use the slice API, that's why I've used the normal `g_new0`.

src/gs-update-monitor.c
joaquimrocha: @mariospr See my other reply to the `g_slice_new0`.

src/gs-update-monitor.c
joaquimrocha: I've tried to change these little safeguard upstream but Richard thinks they're a good thing... So this is me complying with that.

src/gs-update-monitor.c
pwithnall: You might not be interested in the result, but you should be interested in whether the call has completed.

src/gs-update-monitor.c
joaquimrocha: About the misalignments. This code uses tabs (I know...) so it shows funny depending on the editor. In emacs it looks aligned so I guess I'll believe that.

joaquimrocha: About defining the callback after its use, I agree with you generally, and it's even a recommendation from the [Clean Code book](https://www.goodreads.com/book/show/3735293-clean-code) IIRC. However, for C code and GNOME in particular, the opposite is more common as it skips having to pre-declare the function. So I prefer to leave it like this.

src/gs-update-monitor.c
joaquimrocha: Why is this a problem? The `GsUpdateMonitor` runs in the main thread. The first version of this code was using the real asynchronicity as you say, and I usually prefer that (iterating in loops is ugly), but the code did get more complicated, so added it like this.

src/gs-update-monitor.c
joaquimrocha: Tabs... :disappointed:

src/gs-update-monitor.c
joaquimrocha: @mariospr See my response to the other similar comment.
@pwithnall I thought that if the code was running in the main thread it'd be fine like this, but if not, then I can do what you're saying. Maybe we could have this cumbersome synchronicity in a `..._sync` method in the Mogwai schedule client API.

src/gs-update-monitor.c
joaquimrocha: It used to be an autoptr elsewhere... copy&paste mistake I think. Thanks for spotting.

src/gs-update-monitor.c
```diff
@@ -179,11 +482,20 @@ get_updates_finished_cb (GObject *object,
 		return;
 	}
 
+	if (monitor->scheduled_updates_cancellable == NULL ||
+	    g_cancellable_is_cancelled (monitor->scheduled_updates_cancellable))
+		monitor->scheduled_updates_cancellable = g_cancellable_new ();
```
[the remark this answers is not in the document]
joaquimrocha: I will use `g_set_object`. Thanks.

src/gs-update-monitor.c
mariospr: The problem is that you're iterating the main loop of the global context, which might be processing other sources than the ones you're interested in for the purpose of this functionality. If you want to go the "manually spinning the loop" route, I think a better way would be to create a new context and push it as the thread default before starting the asynchronous operation, and then spinning it from there, then pop that context out.

src/gs-update-monitor.c
```diff
@@ -33,6 +33,8 @@
 
 #define APP_METADATA_AUTO_UPDATING "GnomeSoftware::auto-updating"
 
+#define UPDATE_CHECK_INTERVAL (2 * 3600) /* secs */
```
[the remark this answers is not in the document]
joaquimrocha: I don't like to add comments like `2 hours` in this case because if we change the value we must change the comment, and often we forget to do that.

src/gs-update-monitor.c
joaquimrocha: I did it like this so I don't call a function that may not be needed (notice the early return). But I realize now I did these changes based on what was there before. We can just assign the `now` var before the previous if (and after the early returns).

src/gs-update-monitor.c
pwithnall: What Mario said. Additionally, how can you be sure the `GsUpdateMonitor` will always run in the main thread? If the code gets refactored in the future, this will silently start to cause a lot of hard to debug problems.

src/gs-update-monitor.c
```diff
@@ -68,6 +69,20 @@ typedef struct {
...
 
+static void
+scheduled_entry_removed_cb (GObject *source_object,
+			    GAsyncResult *result,
+			    GAsyncResult **out_result)
+{
+	MwscScheduleEntry *entry = (MwscScheduleEntry *) source_object;
+	g_autoptr(GError) local_error = NULL;
+
+        if(!mwsc_schedule_entry_remove_finish (entry, result, &local_error))
```
pwithnall: Nitpick: Missing space before `(`

src/gs-update-monitor.c
```diff
@@ -500,6 +459,72 @@ schedule_update (GsUpdateMonitor *monitor,
...
+static void
+schedule_updates (GsUpdateMonitor *monitor,
+		  GsAppList *apps_to_update)
+{
+	/* if we don't have a valid scheduler yet, create it asynchronously and
+	 * delegate the updates scheduling to its callback */
+	if (monitor->scheduler == NULL) {
+		SchedulerHelper *helper = g_new0 (SchedulerHelper, 1);
+		helper->monitor = monitor;
+		helper->apps_to_update = apps_to_update;
```
pwithnall: Nitpick: Use `g_steal_pointer (&apps_to_update)` here to make it obvious you’re transferring ownership to the `SchedulerHelper`?

src/gs-update-monitor.c
```diff
@@ -500,6 +459,72 @@ schedule_update (GsUpdateMonitor *monitor,
...
+
+		mwsc_scheduler_new_async (monitor->scheduled_updates_cancellable,
+					  (GAsyncReadyCallback) scheduler_ready_cb,
+					  helper);
+		return;
+	}
+
+	schedule_updates_real (monitor, apps_to_update);
```
pwithnall: I think `apps_to_update` is leaked on this path.

I think it probably makes more sense to not transfer ownership of it in to this function, and remove the `g_steal_pointer()` call from line 591 below.

src/gs-update-monitor.c
```diff
@@ -500,6 +459,72 @@ schedule_update (GsUpdateMonitor *monitor,
...
+		GsApp *app = gs_app_list_index (apps_to_update, i);
+		schedule_update (monitor, app,
+				 monitor->scheduled_updates_cancellable);
+	}
+}
+
+static void
+monitor_clear_scheduler (GsUpdateMonitor *monitor)
+{
+	g_clear_object (&monitor->scheduler);
```
pwithnall: You should disconnect from the `invalidated` signal here, in case the `MwscScheduler` object hangs around after you drop your reference. If it does, you might end up getting a subsequent `monitor_clear_scheduler()` callback and end up doing a double-unref.

src/gs-update-monitor.c
joaquimrocha: And there was spaces instead of tabs too...

src/gs-update-monitor.c
joaquimrocha: I pondered that and decided not to do it but yeah, maybe it's more future proof.

src/gs-update-monitor.c
joaquimrocha: I need to transfer ownership because this will be used by the scheduler async cb. Thanks for spotting the leak.

src/gs-update-monitor.c
joaquimrocha: (meaning I will do it :smile: )

src/gs-update-monitor.c
joaquimrocha: I connected to this using `g_signal_connect_object`, and AFAIU if we destroy GsUpdateMonitor, then the callback for this signal won't be called again. Or maybe I am wrong?

src/gs-update-monitor.c
joaquimrocha: I thought you were actually referring to the `helper`, sorry. I will add something to both vars in any case. Stay tuned :)

src/gs-update-monitor.c
mariospr: Ok

src/gs-update-monitor.c
mariospr: Ok

src/gs-update-monitor.c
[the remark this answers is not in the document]
mariospr: I don't see the problem with updating the comment and I think it increases readability, but not strong opìnion.

src/gs-update-monitor.c
pwithnall: Yes, but the code doesn’t destroy the `GsUpdateMonitor` when the `invalidated` signal is received (as far as I can tell); it only clears `monitor->scheduler`.

I realise this is a bit immaterial because you should expect the `invalidated` signal to only ever be emitted once. Fixing this would make the `GsUpdateMonitor` code more resilient to bugs in `libmogwai-schedule-client`.

src/gs-update-monitor.c
pwithnall: :+1:

src/gs-update-monitor.c
pwithnall: It does use `g_clear_object()`, though, so you won’t end up unreffing the scheduler more than once. My mistake.

joaquimrocha:
Jenkins, test this please.

PR #344, opened 2018-04-13
T21327

joaquimrocha:
**IMPORTANT:** This is using #342 (T20402) as a base, since the latter is not yet merged. So please use this PR only for reviewing purposes. I can point it to master once the mentioned PR is merged.

https://phabricator.endlessm.com/T21327


plugins/flatpak/gs-flatpak.c
```diff
@@ -1524,6 +1533,48 @@ gs_flatpak_add_updates (GsFlatpak *self, GsAppList *list,
...
+}
+
+static UpdatePriority
+get_app_update_priority (GsApp *app)
+{
+	if (gs_app_get_kind (app) == AS_APP_KIND_RUNTIME) {
+		if (is_endless_content_runtime (app))
+			return UPDATE_PRIORITY_HIGHEST;
+		return UPDATE_PRIORITY_LOW;
+	}
```
pwithnall: Nitpick: You could reduce the nesting here, since `is_endless_content_runtime()` checks whether the `GsApp` is a runtime. So you could do:
```
if (is_endless_content_runtime (app)
  return UPDATE_PRIORITY_HIGHEST;
else if (gs_app_get_kind (app) == AS_APP_KIND_RUNTIME)
  return UPDATE_PRIORITY_LOW;
else if (is_endless_app (app))
  /* we prioritise Endless apps … */
  return UPDATE_PRIORITY_HIGH;
else
  return UPDATE_PRIORITY_NORMAL;
```

I think that would make the ordering of priorities a little more obvious.

plugins/flatpak/gs-flatpak.c
```diff
@@ -1524,6 +1533,48 @@ gs_flatpak_add_updates (GsFlatpak *self, GsAppList *list,
 	return TRUE;
 }
 
+static gboolean
+is_endless_app (GsApp *app)
+{
+	return g_str_has_prefix (gs_app_get_id (app), "com.endlessm.");
```
pwithnall: Nitpick: Might want to add `gs_app_get_kind (app) == AS_APP_KIND_APP`. `get_app_update_priority()` currently calls this method in a way that this doesn’t matter, but it might matter in future.

joaquimrocha: That check would have to be done outside, as this function is also used inside the `is_endless_content_runtime`. `app` here means GsApp, not Flatpak app.

PR #345, opened 2018-04-16
eos: remove LibreOffice from Flatpak blacklist

ramcq:
https://phabricator.endlessm.com/T22113


ramcq:
@nedrichards Change to featured apps OK?

nedrichards:
Yep. I'd prefer to retain the individual tiles all pushing to the same landing page, but I don't believe that's how GNOME Software does its matching currently and it's not worth delaying the change for that.

ramcq:
> Yep. I'd prefer to retain the individual tiles all pushing to the same landing page, but I don't believe that's how GNOME Software does its matching currently and it's not worth delaying the change for that.

@joaquimrocha How hard/easy is that to achieve without horrible hardcoding?

PR #346, opened 2018-04-19
flatpak: Fix adding apps from special desktop files

joaquimrocha:
Some apps (like LibreOffice) ship several desktop files corresponding
to sub-apps or different ways of launching the same binary. These were
not getting shown in GNOME Software because the Flatpak plugin was
trying to match a Flatpak app for those desktop files from their file
name. Instead of matching the filename in these cases, we should
instead look for the "X-Flatpak" key in the desktop file, which will
directly indicate what's the app corresponding to it.

This patch implements the solution mentioned above, which allows e.g.
for the Writer, Calc, Impress, etc.

https://phabricator.endlessm.com/T22154


mwleeds:
The linked ticket has been closed as WONTFIX.

mwleeds:
For posterity, the reason was:

> In 3.9, all LibreOffice components' icons are shown on the desktop. We have removed the ability to add and remove things from the desktop using the app center, so the second AC is wontfix, and since they cannot be installed and uninstalled independently, they should not be shown independently there.

PR #352, opened 2018-04-28
Add new strings

GeorgesStavracas:
This is temporary until we have the notifications, as asked by Roddy.


rshuler:
Thanks @GeorgesStavracas.  I've updated the strings in Transifex accordingly, so this is no longer needed.

PR #353, opened 2018-05-02
eos: Blacklist gnome-system-monitor-kde

joaquimrocha:
There is a desktop file for providing details about the GNOME System
Monitor app in KDE environments, which was showing up in GNOME Software.

This app should eventually be blocked by checking its project group (set
by the OnlyShowIn key in the desktop file), but this value is being
incorrectly set to GNOME (see [1]) and thus there's no generic way of
hiding the mentioned app. So this patch just blacklists it instead of
setting it as installed, as it does to the rest of the core apps.

[1] https://github.com/hughsie/appstream-glib/issues/236

https://phabricator.endlessm.com/T21451

### DIFF
--- a/plugins/eos/gs-plugin-eos.c
+++ b/plugins/eos/gs-plugin-eos.c
@@ -1454,6 +1454,16 @@ gs_plugin_eos_refine_core_app (GsApp *app)
 	if (gs_app_get_kind (app) == AS_APP_KIND_OS_UPGRADE)
 		return;
 
+	/* blacklist the KDE desktop file of the GNOME System Monitor since
+	 * it's a core app but should not be shown */
+	if (g_strcmp0 (gs_app_get_id (app), "gnome-system-monitor-kde.desktop") == 0) {
+		g_debug ("Blacklisting %s because it will show as a duplicate "
+			 "of the real gnome-system-monitor one.",
+			 gs_app_get_unique_id (app));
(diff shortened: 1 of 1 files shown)


ximion:
@joaquimrocha Why don't you just add `X-AppStream-Ignore=true` to the respective .desktop file? That will make the component invisible in all AppStream implementations, not just this one, and is also more maintainable than hardcoding bits in C code.

joaquimrocha:
@ximion , we still want that desktop file to be considered in KDE. Won't that entry break that?

joaquimrocha:
@ximion If I understand this correctly, we shouldn't ignore this desktop file as it's supposedly useful for KDE. So I was just trying to make sure that we don't show it in GNOME Software, and due to this bug we get its project name as GNOME instead of KDE. I have no idea whether this should be shown or not in application menus, but it doesn't seem sensible to add the `X-Appstream-Ignore` if the issue is that it gets a wrong project name due to a bug.

ximion:
@joaquimrocha Heh, sorry for the confusion - this is the gnome-software fork of Endless, and I for some reason thought it was a different project. I should have looked more carefully :P
I think this patch is perfectly fine, although hardcoding names is still not great (but at the same time, I don't want to introduce any conditionals to when software is shown to AppStream, so at the moment this might be the best solution).

PR #355, opened 2018-05-03
Remove the refresh-when-metered setting

joaquimrocha:
This gsetting was used to control whether a refresh should happen on
metered connection, but we're now always allowing metadata refreshes
even on metered connections, since they are important and we're limiting
them to once every 24 hours. For that reason, this patch removes that
setting and its use altogether.

https://phabricator.endlessm.com/T22389

### DIFF
--- a/src/gs-update-monitor.c
+++ b/src/gs-update-monitor.c
@@ -806,8 +806,6 @@ static void
 check_updates (GsUpdateMonitor *monitor)
 {
 	gint64 tmp;
-	gboolean refresh_on_metered;
-	gboolean connection_is_metered;
 	g_autoptr(GDateTime) last_refreshed = NULL;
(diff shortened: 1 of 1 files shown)


data/org.gnome.software.gschema.xml
```diff
@@ -15,11 +15,6 @@
       <summary>Whether to automatically perform updates</summary>
       <description>If enabled, GNOME Software automatically downloads and installs updates in the background (except for some app types like packages, where it prompts the user to install them when ready).</description>
     </key>
-    <key name="refresh-when-metered" type="b">
```
cosimoc: Like I commented in #356, it feels wrong to have a downstream patch that removes the setting instead of ignoring the value that we don't want to happen in the code.

PR #356, opened 2018-05-03
T22404

joaquimrocha:
https://phabricator.endlessm.com/T22404

### DIFF
--- a/plugins/flatpak/gs-flatpak.c
+++ b/plugins/flatpak/gs-flatpak.c
@@ -1658,7 +1658,6 @@ gs_flatpak_refresh (GsFlatpak *self,
 		    GCancellable *cancellable,
 		    GError **error)
 {
-	guint i;
 	g_autoptr(GPtrArray) xrefs = NULL;
 
 	/* give all the repos a second chance */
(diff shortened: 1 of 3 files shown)


data/org.gnome.software.gschema.xml
```diff
@@ -10,11 +10,6 @@
       <summary>Whether to manage updates in GNOME Software</summary>
       <description>If disabled, GNOME Software will hide the updates panel and not perform any automatic updates actions.</description>
     </key>
-    <key name="download-updates" type="b">
```
cosimoc: Unless this is either also being removed upstream, or it's a downstream-only setting (which does not seem to be the case: https://gitlab.gnome.org/GNOME/gnome-software/blob/master/data/org.gnome.software.gschema.xml#L13), we should not remove it, but just make it always false and/or ignore its value when deciding what to do in the code.

rshuler: My 2 cents...  I don't think it makes sense to keep a gsetting, even if it is upstream, if the code does not actually implement what the gsetting says it should.

And, part of the complication here, is that while normally we could just keep the code and specify the default value that makes sense for Endless, in this case we previously exposed this setting via the UI, so some users may get into 3.4 with the gsetting true, and some with the gsetting false, and if the user had made any changes those changes would override whatever we set as the default, and the user will no longer have any UI to further adjust this value.

cosimoc: @rshuler I agree with you. What I am suggesting is that we simply keep the setting in the schema, and patch the code to ignore its value.

joaquimrocha: @cosimoc I had thought about keeping the setting in the schema but then thought that even if upstream starts using the setting elsewhere, we'd realize about it better if we don't have the setting at all. That said, I don't mind keeping it if for some reason it's better to keep it around.

plugins/flatpak/gs-flatpak.c
```diff
@@ -1679,45 +1678,6 @@ gs_flatpak_refresh (GsFlatpak *self,
 			return FALSE;
 	}
 
-	/* no longer interesting */
-	if ((flags & GS_PLUGIN_REFRESH_FLAGS_PAYLOAD) == 0)
-		return TRUE;
```
cosimoc: Is this patch also intended for upstream adoption? If not, is there anything that would actually call into this method with the refresh-payload flag? In other words I'd like to avoid carrying a downstream-only patch to remove this code if it would never be hit in our configuration.

joaquimrocha: The patch will not yet be added to upstream ATM because the strategy for auto updates there is still under discussion.
I would still like to keep this change because, even if it's not hit by our configuration, I think that downloading updates during refresh is not the correct approach and (equally important) I'd rather make sure this is never possible (otherwise we could hit this again on a future rebase, so it'd be yet another thing to care about for no good reason).
Also, I am trying to push upstream a division between `refreshing` (for metadata/catalogs/etc.) and auto downloads/updates, which should be performed on a different function.

PR #360, opened 2018-05-04
Add the top bar about auto updates to the Updates page

joaquimrocha:
The updates page should inform the user about whether the auto updates
are turned on or not, and offer a link to the Settings so users can
change it.

These changes implement a top bar in the Updates page to show the
mentioned information to users, and change it according to the
Scheduler's "allow-downloads" property.

https://phabricator.endlessm.com/T22436


src/gs-updates-page.ui
```diff
@@ -11,6 +11,39 @@
       <object class="GtkBox">
         <property name="visible">True</property>
         <property name="orientation">vertical</property>
+	<style>
```
joaquimrocha: I know this looks misaligned but it's because I have made sure to use space indentation as that's what should be used in UI files (even if **some** pre-existing bits are using tabs).

joaquimrocha:
I have changed the link in the PR because I have opened T22436 instead for this. But I am keeping this branch since the PR is already opened.

src/gs-updates-page.c
```diff
@@ -1786,6 +1819,75 @@ gs_updates_page_upgrade_cancel_cb (GsUpgradeBanner *upgrade_banner,
 	g_cancellable_cancel (self->cancellable_upgrade_download);
 }
 
+static void
+scheduler_invalidated_cb (GsUpdatesPage *self)
+{
+	g_warning ("The Mogway Scheduler has just been invalidated when "
```
rshuler: Spelling of Mogwai

PR #365, opened 2018-05-08
Only show an app tile as having a scheduled update if there's one

joaquimrocha:
https://phabricator.endlessm.com/T22492


src/gs-update-monitor.c
```diff
@@ -59,6 +59,7 @@ struct _GsUpdateMonitor {
 	GHashTable	*scheduled_updates; /* (element-type utf8 UpdateScheduleHelper) */
 	guint		 num_scheduled_updates;
 	GCancellable	*scheduled_updates_cancellable;
+	gulong		 allow_downloads_hander;
```
rshuler: s/hander/handler/g

PR #371, opened 2018-05-18
T22617

joaquimrocha:
https://phabricator.endlessm.com/T22617

### DIFF
--- a/plugins/flatpak/gs-plugin-flatpak.c
+++ b/plugins/flatpak/gs-plugin-flatpak.c
@@ -130,6 +130,18 @@ gs_plugin_flatpak_add_installation (GsPlugin *plugin,
 	return TRUE;
 }
 
+static void
+gs_plugin_flatpak_report_warning (GsPlugin *plugin,
+				  GError **local_error)
+{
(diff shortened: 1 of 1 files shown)


plugins/flatpak/gs-plugin-flatpak.c
```diff
@@ -130,27 +130,50 @@ gs_plugin_flatpak_add_installation (GsPlugin *plugin,
...
+	gs_plugin_event_add_flag (event,
+				  GS_PLUGIN_EVENT_FLAG_WARNING);
+	gs_plugin_report_event (plugin, event);
+}
+
 gboolean
 gs_plugin_setup (GsPlugin *plugin, GCancellable *cancellable, GError **error)
 {
 	GsPluginData *priv = gs_plugin_get_data (plugin);
+	guint num_installations_added = 0;
```
cosimoc: Why do you need to count this separately instead of just looking at the length of `priv->flatpaks`?

joaquimrocha: You're right. I will change that.

PR #372, opened 2018-05-21
Refactor screenshots and hide those unavailable when offline

joaquimrocha:
The details page displays a number of screenshots for regular apps and
it works offline as long as the screenshots have been previously cached.
However, if the screenshots have not been cached, an error screenshot
placeholder is shown instead. This is a problem for many users who are
constantly offline, as most of their apps will display the error
screenshot and thus affect the UX negatively as every app appears to
have an issue.

To help provide a cleaner UX, this patch hides screenshots if they
cannot be loaded when the user is offline. In the process these changes
also refactor the logic around showing the screenshots.

https://phabricator.endlessm.com/T19887

### DIFF
--- a/src/gs-details-page.c
+++ b/src/gs-details-page.c
@@ -602,11 +602,24 @@ gs_details_page_notify_state_changed_cb (GsApp *app,
 }
 
 static void
+gs_details_page_load_main_screenshot (GsDetailsPage *self,
+				      AsScreenshot *screenshot)
+{
+	GsScreenshotImage *ssmain;
+	g_autoptr(GList) children = NULL;
+
+	children = gtk_container_get_children (GTK_CONTAINER (self->box_details_screenshot_main));
+	ssmain = GS_SCREENSHOT_IMAGE (children->data);
+
+	gs_screenshot_image_set_screenshot (ssmain, screenshot);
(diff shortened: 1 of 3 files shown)


src/gs-details-page.c
```diff
@@ -687,62 +703,89 @@ gs_details_page_refresh_screenshots (GsDetailsPage *self)
...
+						      AS_IMAGE_NORMAL_WIDTH,
+						      AS_IMAGE_NORMAL_HEIGHT);
+			gs_screenshot_image_load_async (GS_SCREENSHOT_IMAGE (ssmain), NULL);
+
+			/* when we're offline, the load will be immediate, so we
+			 * can check if it succeeded, and just skip it and its
+			 * thumbnails otherwise */
+			if (is_offline &&
+			    !gs_screenshot_image_is_showing (GS_SCREENSHOT_IMAGE (ssmain)))
+				continue;
```
rshuler: Is it okay that we continue here without incrementing num_screenshots_loaded?  I'm guessing not.

joaquimrocha: The screenshot has not been loaded (even if we called it asynchronously here, we're offline, so it should load any cached screenshots immediately), so it's correct to continue without incrementing the value.

PR #373, opened 2018-05-23
flatpak: Include related refs to the app's runtime on install/update

joaquimrocha:
When installing/updating an app, if its related refs and its runtime
need to be installed as well, then they do. However, refs that are
related to the runtime itself do not get installed, and this is a
problem because some apps may need them (besides being a discrepancy
between using the Flatpak CLI and GNOME Software for installing Flatpak
apps).

Thus, these changes add any refs related to an app's runtime to the
list of refs that should be installed when installing or updating an
app. Note that it deliberately doesn't update refs related to the
runtime though, just as it doesn't update the runtime when updating an
app. Thus, refs that are related to the runtime get updated together
with it.

https://phabricator.endlessm.com/T22716

### DIFF
--- a/plugins/flatpak/gs-flatpak.c
+++ b/plugins/flatpak/gs-flatpak.c
@@ -2905,6 +2905,72 @@ gs_flatpak_add_app_to_list_if_not_installed (GsApp *app, GsAppList *list,
 	gs_app_list_add (list, app);
 }
 
+static GPtrArray *
+gs_flatpak_get_app_remote_related_refs (GsFlatpak *self,
+					GsApp *app,
+					GCancellable *cancellable,
+					GError **error)
+{
+	g_autofree gchar *ref = gs_flatpak_app_get_ref_display (app);
+	/* lookup any related refs for this ref */
+	return flatpak_installation_list_remote_related_refs_sync (self->installation,
+								   gs_app_get_origin (app),
+								   ref, cancellable, error);
+}
+
+static gboolean
+gs_flatpak_add_related_refs_to_list (GsFlatpak *self,
+				     GsApp *app,
+				     GPtrArray *related_refs,
+				     GHashTable *hash_installed,
+				     gboolean is_update,
+				     GsAppList *list,
+				     GCancellable *cancellable,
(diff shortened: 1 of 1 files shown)


plugins/flatpak/gs-flatpak.c
```diff
@@ -2905,6 +2905,78 @@ gs_flatpak_add_app_to_list_if_not_installed (GsApp *app, GsAppList *list,
...
+		/* already installed? */
+		app_tmp = gs_flatpak_create_app (self, FLATPAK_REF (xref_related));
+		ref_display = gs_flatpak_app_get_ref_display (app_tmp);
+		if (!is_update && g_hash_table_contains (hash_installed, ref_display)) {
+			g_debug ("not adding related %s as already installed", ref_display);
+			continue;
+		}
+
+		if (gs_app_get_origin (app_tmp) == NULL)
+			gs_app_set_origin (app_tmp, gs_app_get_origin (app));
```
cosimoc: The old code was doing this unconditionally; any reason you changed it?

joaquimrocha: Yes. I have noticed that it gave a warning when the `app_tmp` already has an origin set. So I sneaked it in since the whole block was changed.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2905,6 +2905,78 @@ gs_flatpak_add_app_to_list_if_not_installed (GsApp *app, GsAppList *list,
...
+					GError **error)
+{
+	g_autofree gchar *ref = NULL;
+
+	/* lookup any related refs for this ref */
+	ref = g_strdup_printf ("%s/%s/%s/%s",
+			       gs_flatpak_app_get_ref_kind_as_str (app),
+			       gs_flatpak_app_get_ref_name (app),
+			       gs_flatpak_app_get_ref_arch (app),
+			       gs_flatpak_app_get_ref_branch (app));
```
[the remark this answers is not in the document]
joaquimrocha: I had this code from months ago, so I just copied it. But I will check it.

PR #374, opened 2018-06-01
T22749

joaquimrocha:
https://phabricator.endlessm.com/T22749

### DIFF
--- a/plugins/flatpak/gs-flatpak.c
+++ b/plugins/flatpak/gs-flatpak.c
@@ -69,6 +69,10 @@ static gboolean
 gs_flatpak_refresh_appstream (GsFlatpak *self, guint cache_age,
 			      GsPluginRefreshFlags flags,
 			      GCancellable *cancellable, GError **error);
+static gboolean
+gs_plugin_refine_item_metadata (GsFlatpak *self, GsApp *app,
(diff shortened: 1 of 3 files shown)


plugins/flatpak/gs-flatpak.c
```diff
@@ -2048,11 +2049,28 @@ gs_plugin_refine_item_state (GsFlatpak *self,
...
 		gs_app_set_state (app, AS_APP_STATE_UPDATABLE_LIVE);
+
+		/* show event if needed */
+		if ((flags & GS_PLUGIN_REFINE_FLAGS_INTERACTIVE) != 0) {
+			event = gs_plugin_event_new ();
+			g_set_error (&local_error, GS_PLUGIN_ERROR, GS_PLUGIN_ERROR_FAILED,
+				     _("The app %s is missing its runtime. "
+				       "Update the app to repair this problem."),
+				     gs_app_get_id (app));
+			gs_flatpak_error_convert (&local_error);
```
ramcq: Is this call necessary when the error is already in the GS error domain?

joaquimrocha: Actually no. C&P I guess. I'll change that.
:+1:

PR #375, opened 2018-06-05
T22807

joaquimrocha:
https://phabricator.endlessm.com/T22807

### DIFF
--- a/src/gs-category-page.c
+++ b/src/gs-category-page.c
@@ -28,6 +28,8 @@
 #include "gs-background-tile.h"
 #include "gs-category-page.h"
 
+#define MAX_PLACEHOLDER_TILES 30
+
 struct _GsCategoryPage
 {
 	GsPage		 parent_instance;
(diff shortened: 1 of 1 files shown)


uajain:
Overall LGTM except that I have few questions.

I don't quite understand your e2358cf commit. Isn't the `gs_category_filter_apps_func` responsible to filter out apps which are not in category? As per my thinking, even if the app-tile was previously generated and now present in the cache, the view should be filtered by `gs_category_filter_apps_func`
Why there is a need to remove it from the cache ? Am I missing something?

src/gs-category-page.c
```diff
@@ -266,6 +338,11 @@ gs_category_page_dispose (GObject *object)
 	g_clear_object (&self->subcategory);
 	g_clear_object (&self->plugin_loader);
 
+	if (self->category_apps != NULL) {
+		g_hash_table_unref (self->category_apps);
+		self->category_apps = NULL;
+	}
```
uajain: Maybe just use `g_clear_pointer (&self->category_apps, (GDestroyNotify) g_hash_table_unref)` and remove the `if` condition ?

joaquimrocha: :+1:

joaquimrocha:
@uajain Thanks for reviewing this PR. The https://github.com/endlessm/gnome-software/commit/e2358cf5a360daafa0b57fc5b81160c5137233b0 commit is needed because e.g. if an app is removed from a remote, or its categories are changed and it no longer has a category it used to, then it'd still show up in the category view (because even though the plugins would no longer return it, it'd still be in the cache).
The filter function only works to filter the existing tiles for each app, but if the tiles are from an app in the conditions I mentioned above, then it's still shown. Does this make sense?

uajain:
Ah, yes. Thanks for the explaination.

PR #377, opened 2018-06-11
Lower the minimum number of popular apps required to be shown

joaquimrocha:
The minimum number of popular apps to be shown was being set to the
maximum allowed, as a result, if an image overrode the popular apps
(using the popular-overrides gsetting) then those weren't being shown.

This patch separates the maximum number of popular apps allowed from
the minimum required.

https://phabricator.endlessm.com/T22925

### DIFF
--- a/src/gs-overview-page.c
+++ b/src/gs-overview-page.c
@@ -35,6 +35,7 @@
 #include "gs-common.h"
 
 #define N_TILES 35
+#define MIN_POPULAR_TILES 2
 
 typedef struct
 {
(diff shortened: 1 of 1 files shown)


src/gs-overview-page.c
```diff
@@ -35,6 +35,7 @@
 #include "gs-common.h"
 
 #define N_TILES 35
+#define MIN_POPULAR_TILES 2
```
rshuler: I'm not sure why we bother setting a minimum of 2 -- if someone sets up an image to have just one popular title, it seems like we could respect it.  In practice, I doubt we will ever set just one, so I'm going to accept as is, but in the future we might want to reduce the minimum to 1.

PR #378, opened 2018-06-19
Don't show the fallback screenshot placeholder when offline

joaquimrocha:
The fallback screenshot placeholder is used when apps don't provide any
screenshots, and should not be shown if the user is offline (because
they will never get an updated AppStream data and thus would always see
the placeholder).
    
This is in line with the rest of the changes for T19887 that just hide the
screenshots when they fail to load and the user is offline.
    
https://phabricator.endlessm.com/T19887


uajain:
Reviewed +1

PR #380, opened 2018-06-20
Display ongoing app-update at top of the list for easy-visibility

uajain:
This is very specific to auto-updating which have long list
of pending updates. Mogwai is the actor choosing which update
to schedule next hence, the order cannot be guaranteed.

https://phabricator.endlessm.com/T22556

### DIFF
--- a/src/gs-updates-page.c
+++ b/src/gs-updates-page.c
@@ -629,6 +629,10 @@ _app_state_notify_cb (GsApp *app, GParamSpec *pspec, gpointer user_data)
 	if (gs_app_get_state (app) == AS_APP_STATE_INSTALLED) {
 		gs_app_row_unreveal (app_row);
 	}
+
+       if (page->scheduler != NULL && mwsc_scheduler_get_allow_downloads (page->scheduler))
+		gtk_list_box_invalidate_sort (page->listboxes[GS_UPDATE_PAGE_SECTION_ONLINE]);
+
 	gs_updates_page_update_ui_state (page);
 }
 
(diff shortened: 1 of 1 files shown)


uajain:
Hey. Thanks for review. I have updated the PR as per your review. I just have one comment of which I am not very sure. See below.

src/gs-updates-page.c
```diff
@@ -646,6 +650,21 @@ _get_app_sort_key (GsApp *app)
...
 	g_string_append_printf (key, "%u:", _get_app_section (app));
 
+	switch (gs_app_get_state (app)) {
+	case AS_APP_STATE_INSTALLING:
+		g_string_append (key, "1:");
+		break;
+	case AS_APP_STATE_QUEUED_FOR_INSTALL:
+		g_string_append (key, "2:");
+		break;
+	case AS_APP_STATE_REMOVING:
```
uajain: Not sure if this case is relevant.

uajain:
>One thing that'd be nice though would be to prevent the sorting if the app has been manually updated, but this may be very complicated, so if you still want to do it, make sure it's in a different commit.

This hold true with the current PR. Re-sorting occurs only if auto-update/scheduled update situation is detected. Otherwise, it won't (hence, manually updating will keep the app-row in the current view). Does that address you suggestion here?

src/gs-updates-page.c
joaquimrocha: If it's being removed, then maybe it shouldn't be shown in the updates page?
In any case, ignore the removing and queued for install, since apps in these states should not be shown in the updates page.

uajain:
Ah Ok, that's sounds tricky to solve. Maybe I need to look that specific use-case (allow-downloads is on and manual updating of the app by user). Definitely needs a separate ticket for that.

Meanwhile updated this PR. Thank you for the comments.

src/gs-updates-page.c
```diff
@@ -646,6 +650,15 @@ _get_app_sort_key (GsApp *app)
 	 * 4. online device firmware */
 	g_string_append_printf (key, "%u:", _get_app_section (app));
 
+	switch (gs_app_get_state (app)) {
+	case AS_APP_STATE_INSTALLING:
+		g_string_append (key, "1:");
+		break;
+	default:
+		g_string_append (key, "2:");
+		break;
```
joaquimrocha: Sorry to only notice it now, but if it's just one state, can you make this an if statement instead?

PR #384, opened 2018-07-02
appstream: Add "GnomeSoftware::popular" kudo to com.endless. apps

uajain:
"GnomeSoftware::popular" kudo has been removed from the build
process[1] due to recent version of appstream-validate started
rejecting that kudo. In order to still keep the logic for getting
"Featured" apps and search-provider relevant, we add this kudo at
runtime while reading the appstream.

[1] https://github.com/endlessm/eos-build/commit/05977432eb2a624a436caedbcb92f9acd41581de

https://phabricator.endlessm.com/T23152

### DIFF
--- a/plugins/core/gs-appstream.c
+++ b/plugins/core/gs-appstream.c
@@ -771,6 +771,11 @@ gs_appstream_refine_app (GsPlugin *plugin,
 	if (as_app_has_category (item, "featured"))
 		gs_app_add_kudo (app, GS_APP_KUDO_FEATURED_RECOMMENDED);
 
+	/* Mark com.endlessm. apps with featured kudo. See discussion on T23152 */
+	if (!gs_app_has_kudo (app, GS_APP_KUDO_FEATURED_RECOMMENDED) &&
+	     g_str_has_prefix (gs_app_get_id (app), "com.endlessm."))
+		gs_app_add_kudo (app, GS_APP_KUDO_FEATURED_RECOMMENDED);
+
 	/* add new-style kudos */
 	kudos = as_app_get_kudos (item);
 	for (i = 0; i < kudos->len; i++) {


plugins/core/gs-appstream.c
```diff
@@ -765,6 +765,10 @@ gs_appstream_refine_app (GsPlugin *plugin,
 	    as_app_get_language (item, tmp) > 50)
 		gs_app_add_kudo (app, GS_APP_KUDO_MY_LANGUAGE);
 
+	/* Add popular tag to com.endlessm. tags See discussion on T23152 */
+	if (g_str_has_prefix (gs_app_get_id (app), "com.endlessm."))
+		as_app_add_kudo (item, "GnomeSoftware::popular");
```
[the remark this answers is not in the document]
dbnicholson: Actually, I think this would be best if it was done in `gs_plugin_refine()` rather than `gs_plugin_add_popular()`, but either way I think you should just jump straight to `GS_APP_KUDO_FEATURED_RECOMMENDED` instead of trying to get it managed through appstream.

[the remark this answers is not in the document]
ramcq: Yeah, the add_popular is the wrong place for this if we want to do it in the eos plugin, because it's only passed the other apps in the list if they are already popular. If every plugin's refine is called on every app, that works.

uajain:
Addressed the comments above.

plugins/eos/gs-plugin-eos.c
```diff
@@ -1730,6 +1730,10 @@ gs_plugin_refine (GsPlugin		*plugin,
 		if (gs_plugin_eos_blacklist_app_for_remote_if_needed (plugin, app))
 			continue;
 
+		/* Mark com.endlessm. apps with featured kudo. See discussion on T23152 */
+		if (g_str_has_prefix (gs_app_get_id (app), "com.endlessm."))
```
joaquimrocha: I am not sure if we're going with this solution in the end, but if we are, perhaps verifying whether the kudo is already set makes for a good optimization (otherwise, in every operation we'd be comparing prefixes).

dbnicholson: That's a good idea. How often does `gs_plugin_refine` run? It seems like it's intended for batch operations that only run when refreshing.

plugins/core/gs-appstream.c
```diff
@@ -771,6 +771,11 @@ gs_appstream_refine_app (GsPlugin *plugin,
 	if (as_app_has_category (item, "featured"))
 		gs_app_add_kudo (app, GS_APP_KUDO_FEATURED_RECOMMENDED);
 
+	/* Mark com.endlessm. apps with featured kudo. See discussion on T23152 */
+	if (!gs_app_has_kudo (app, GS_APP_KUDO_FEATURED_RECOMMENDED) &&
+	     g_str_has_prefix (gs_app_get_id (app), "com.endlessm."))
+		gs_app_add_kudo (app, GS_APP_KUDO_FEATURED_RECOMMENDED);
```
dbnicholson: If it already has the `GS_APP_KUDO_FEATURED_RECOMMENDED`, then what's the point of adding it again? This is a no-op. What's setting `GS_APP_KUDO_FEATURED_RECOMMENDED` in the first place?

PR #386, opened 2018-07-10
Set install-timestamp gsetting when we update or upgrade

uajain:
In order to adjust the frequency of our update related app-center
notifications, we need to leverage "install-timestamp" first.
Tried to cover these cases (might be missing more of these!):

* App was set to auto-update and it got updated
* An OS upgrade has been done (from updates Tab)
* While clicking on "update all"
* App has been updated from App-details page or  from the Updates
  tab's app-row (i.e. it's a user-initiated single app update)

### DIFF
--- a/lib/gs-plugin-loader.c
+++ b/lib/gs-plugin-loader.c
@@ -3112,6 +3112,8 @@ gs_plugin_loader_generic_update (GsPluginLoader *plugin_loader,
 		helper->anything_ran = TRUE;
 		gs_plugin_status_update (plugin, NULL, GS_PLUGIN_STATUS_FINISHED);
 	}
+
+	gs_utils_update_install_timestamp (priv->settings);
 	return TRUE;
 }
 
@@ -3175,6 +3177,11 @@ gs_plugin_loader_process_thread_cb (GTask *task,
 			g_task_return_error (task, error);
 			return;
 		}
+	}
+
+	if (action == GS_PLUGIN_ACTION_UPGRADE_TRIGGER) {
+		GsPluginLoaderPrivate *priv = gs_plugin_loader_get_instance_private (plugin_loader);
+		gs_utils_update_install_timestamp (priv->settings);
 	}
 
 	/* remove from pending list */
--- a/lib/gs-utils.c
+++ b/lib/gs-utils.c
@@ -1041,4 +1041,22 @@ gs_utils_get_memory_total (void)
 	return si.totalram / MB_IN_BYTES / si.mem_unit;
 }
 
+/**
+ * gs_utils_update_install_timestamp:
+ *
+ * Sets the value of install-timestamp to current epoch. "install-timestamp" represents
+ * the last time we had an update or upgrade.
+ *
+ **/
+void
+gs_utils_update_install_timestamp (GSettings *settings)
+{
+	g_autoptr(GDateTime) now = NULL;
+
+	g_return_if_fail (settings != NULL);
+
+	now = g_date_time_new_now_local ();
+	g_settings_set (settings, "install-timestamp", "x", g_date_time_to_unix (now));
+}
+
 /* vim: set noexpandtab: */
--- a/lib/gs-utils.h
+++ b/lib/gs-utils.h
@@ -91,6 +91,7 @@ void		 gs_utils_append_key_value	(GString	*str,
 						 const gchar	*key,
 						 const gchar	*value);
 guint		 gs_utils_get_memory_total	(void);
+void		 gs_utils_update_install_timestamp (GSettings *settings);
 
 G_END_DECLS
 


src/gs-page.c
```diff
@@ -180,6 +184,9 @@ gs_page_app_installed_cb (GObject *source,
 		return;
 	}
 
+        now = g_date_time_new_now_local ();
+        g_settings_set (helper->settings, "install-timestamp", "x", g_date_time_to_unix (now));
+
 	/* only show this if the window is not active */
```
uajain: This might look spurious, but we are setting the timestamp here because `gs_page_app_installed_cb` is also the kinda callback for `gs_page_update_app`.

But we might to take in consideration that this will **also** set the timestamp for something like `gs_page_install_app`

uajain:
On a general note, the value of timestamp is `g_date_time_new_now_local ()` (which should do the job for now I think).
At two places, **maybe**  we can set the timestamp to be `gs_app_install_date` (not sure)
Needs testing for that probably.

src/gs-page.c
```diff
@@ -180,6 +184,9 @@ gs_page_app_installed_cb (GObject *source,
 		return;
 	}
 
+        now = g_date_time_new_now_local ();
+        g_settings_set (helper->settings, "install-timestamp", "x", g_date_time_to_unix (now));
+
```
joaquimrocha: Why do you need to update the timestamp in the installed page? Is it just for consistency with the setting's name?

src/gs-page.c
```diff
@@ -67,6 +68,8 @@ gs_page_helper_free (GsPageHelper *helper)
 		g_object_unref (helper->soup_session);
 	if (helper->price != NULL)
 		g_object_unref (helper->price);
+	if (helper->settings != NULL)
+		g_object_unref (helper->settings);
```
joaquimrocha: You should use `g_clear_object`.

src/gs-page.c
joaquimrocha: Also notice that this line + their copies are not correctly indented (they're using spaces instead of tabs).

src/gs-updates-page.c
```diff
@@ -783,6 +784,9 @@ _perform_update_cb (GsPluginLoader *plugin_loader, GAsyncResult *res, gpointer u
 		return;
 	}
 
+	now = g_date_time_new_now_local ();
+	g_settings_set (helper->self->settings, "install-timestamp", "x", g_date_time_to_unix (now));
+
```
joaquimrocha: If you are going to call this from many places, then I think it's worth that you abstract it in a function (probably in `gs-utils`).

src/gs-page.c
```diff
@@ -581,6 +588,7 @@ gs_page_update_app (GsPage *page, GsApp *app, GCancellable *cancellable)
 	helper->app = g_object_ref (app);
 	helper->page = g_object_ref (page);
 	helper->cancellable = g_object_ref (cancellable);
+	helper->settings = g_object_ref (g_settings_new ("org.gnome.software"));
```
joaquimrocha: The `g_settings_new` call already gives you a new object. If you call `g_object_ref` here, they you will be leaking memory.

uajain: Ah, I looked for a `utils` file but couldn't find it. I realized it I was looking in `src/` only but there is one in `lib/`

src/gs-updates-page.c
uajain: Ah, I looked for a utils file but couldn't find it. I realized it I was looking in src/ only but there is one in lib/

src/gs-page.c
uajain: Hi @joaquimrocha 
Maybe you didn't read my comment : https://github.com/endlessm/gnome-software/pull/386/files#r201407598 ?

Yes, this is not ideal but I am looking into if I can determine if that callback was in context of an update (and not in context of a installlation).

joaquimrocha: Ah. Sorry, it wasn't clear to be. In any case, I'd rather we only add it here once we really find out if it's needed.

lib/gs-utils.c
```diff
@@ -1041,4 +1041,20 @@ gs_utils_get_memory_total (void)
...
 
+/**
+ * gs_utils_set_install_timestamp:
+ *
+ * Sets the value of install-timestamp which represents the last time
+ * we had an update or upgrade.
+ *
+ **/
+void
+gs_utils_set_install_timestamp (GSettings *settings)
```
joaquimrocha: Nitpick: I'd prefer `.._update_install_timestamp` because set usually takes an argument, and I don't know if we will need to add that argument in the future.

src/gs-page.c
```diff
@@ -49,6 +50,7 @@ typedef struct {
 	GtkWidget	*button_install;
 	GsPluginAction	 action;
 	GsShellInteraction interaction;
+	GSettings	*settings;
```
joaquimrocha: The settings doesn't really belong to the operation, and thus the helper. So please add it rather to the private struct of the `GsPage`.

lib/gs-utils.c
uajain: hmm, you make a good point. How about we make this change while we update `install-timestamp` to `update-timestamp`. Does it co-relate? 

I should probably update the title of PR too. :laughing:

lib/gs-utils.c
```diff
@@ -1041,4 +1041,20 @@ gs_utils_get_memory_total (void)
...
+{
```
joaquimrocha: Maybe you should also check if the settings are valid here (`g_return_if_fail`). Because you don't check in all places you called the function.

lib/gs-utils.c
joaquimrocha: I don't see how it correlates. In your PR you're also updating the timestamp on every update, not on every install, thus the incoherence between the name and what it does is independently from how we set it on updates.

src/gs-updates-page.c
```diff
@@ -1627,6 +1628,8 @@ upgrade_trigger_finished_cb (GObject *source,
 		return;
 	}
 
+	gs_utils_update_install_timestamp ();
```
joaquimrocha: Since this is not dependent on whether the user actually reboots the machine, maybe we should add this to the `gs_plugin_loader_process_thread_cb` too? (you can check if the function is `GS_PLUGIN_ACTION_UPGRADE_TRIGGER`)

lib/gs-plugin-loader.c
```diff
@@ -3177,6 +3178,11 @@ gs_plugin_loader_process_thread_cb (GTask *task,
 		}
 	}
 
+	if (action == GS_PLUGIN_ACTION_UPGRADE_TRIGGER) {
+		GsPluginLoaderPrivate *priv = gs_plugin_loader_get_instance_private (plugin_loader);
```
joaquimrocha: Nitpick: Usually you'd want to put the `priv` in the outter scope because it'll often end up being used in other places of the function, but I guess that for now we can keep it close to where it's used.

lib/gs-plugin-loader.c
```diff
@@ -3112,6 +3112,7 @@ gs_plugin_loader_generic_update (GsPluginLoader *plugin_loader,
 		helper->anything_ran = TRUE;
 		gs_plugin_status_update (plugin, NULL, GS_PLUGIN_STATUS_FINISHED);
 	}
+	gs_utils_update_install_timestamp (priv->settings);
```
joaquimrocha: Nitpick: I'd have added a newline before the function call, otherwise it looks a bit cramped.

PR #388, opened 2018-07-25
Restore window activation while showing a local file

uajain:
Fallout from commit 54863ceaf54cf5128f19b7f6e4338905a1b155d0

https://phabricator.endlessm.com/T23365


ramcq:
Good catch!

PR #390, opened 2018-08-07
T23419 Potential fix for GError crash

pwithnall:
https://phabricator.endlessm.com/T23419

Untested, since I haven’t tried to reproduce the bug. This is essentially a drive-by fix.


joaquimrocha:
LGTM. But you should also send them upstream if you have the time.

joaquimrocha:
Nevermind, I see you already did it.

PR #391, opened 2018-08-09
T20146 copy usb apps

treitter:
This implements copying Flatpaks to USB through a new button on the details page.

### DIFF
--- a/lib/gs-plugin-job.c
+++ b/lib/gs-plugin-job.c
@@ -48,6 +48,7 @@ struct _GsPluginJob
 	AsReview		*review;
 	GsPrice			*price;
 	gint64			 time_created;
+	gchar			*copy_dest;
 };
 
 enum {
@@ -67,6 +68,7 @@ enum {
 	PROP_MAX_RESULTS,
 	PROP_PRICE,
 	PROP_TIMEOUT,
+	PROP_COPY_DEST,
 	PROP_LAST
 };
 
@@ -102,6 +104,10 @@ gs_plugin_job_to_string (GsPluginJob *self)
 	if (self->search != NULL) {
 		g_string_append_printf (str, " with search=%s",
 					self->search);
+	}
+	if (self->copy_dest != NULL) {
+		g_string_append_printf (str, " with copy destination=%s",
+					self->copy_dest);
 	}
 	if (self->category != NULL) {
 		GsCategory *parent = gs_category_get_parent (self->category);
@@ -432,6 +438,21 @@ gs_plugin_job_get_price (GsPluginJob *self)
 	return self->price;
 }
 
+void
+gs_plugin_job_set_copy_dest (GsPluginJob *self, const gchar *copy_dest)
+{
+	g_return_if_fail (GS_IS_PLUGIN_JOB (self));
+	g_free (self->copy_dest);
+	self->copy_dest = g_strdup (copy_dest);
+}
+
+const gchar *
+gs_plugin_job_get_copy_dest (GsPluginJob *self)
+{
+	g_return_val_if_fail (GS_IS_PLUGIN_JOB (self), NULL);
+	return self->copy_dest;
+}
+
 static void
 gs_plugin_job_get_property (GObject *obj, guint prop_id, GValue *value, GParamSpec *pspec)
 {
@@ -455,6 +476,9 @@ gs_plugin_job_get_property (GObject *obj, guint prop_id, GValue *value, GParamSp
 		break;
 	case PROP_SEARCH:
 		g_value_set_string (value, self->search);
+		break;
+	case PROP_COPY_DEST:
+		g_value_set_string (value, self->copy_dest);
 		break;
 	case PROP_AUTH:
 		g_value_set_object (value, self->auth);
@@ -513,6 +537,9 @@ gs_plugin_job_set_property (GObject *obj, guint prop_id, const GValue *value, GP
 	case PROP_SEARCH:
 		gs_plugin_job_set_search (self, g_value_get_string (value));
 		break;
+	case PROP_COPY_DEST:
+		gs_plugin_job_set_copy_dest (self, g_value_get_string (value));
+		break;
 	case PROP_AUTH:
 		gs_plugin_job_set_auth (self, g_value_get_object (value));
 		break;
@@ -551,6 +578,7 @@ gs_plugin_job_finalize (GObject *obj)
 {
 	GsPluginJob *self = GS_PLUGIN_JOB (obj);
 	g_free (self->search);
+	g_free (self->copy_dest);
 	g_clear_object (&self->auth);
 	g_clear_object (&self->app);
 	g_clear_object (&self->list);
@@ -602,6 +630,11 @@ gs_plugin_job_class_init (GsPluginJobClass *klass)
 				     NULL,
 				     G_PARAM_READWRITE);
 	g_object_class_install_property (object_class, PROP_SEARCH, pspec);
+
+	pspec = g_param_spec_string ("copy-dest", NULL, NULL,
+				     NULL,
+				     G_PARAM_READWRITE);
+	g_object_class_install_property (object_class, PROP_COPY_DEST, pspec);
 
 	pspec = g_param_spec_object ("auth", NULL, NULL,
 				     GS_TYPE_AUTH,
--- a/lib/gs-plugin-loader.c
+++ b/lib/gs-plugin-loader.c
@@ -71,6 +71,9 @@ typedef struct
 
 	GNetworkMonitor		*network_monitor;
 	gulong			 network_changed_handler;
+
+	GVolumeMonitor		*volume_monitor;
+	GList			*removable_mounts;
 } GsPluginLoaderPrivate;
 
 static void gs_plugin_loader_monitor_network (GsPluginLoader *plugin_loader);
@@ -93,6 +96,7 @@ enum {
 	PROP_EVENTS,
 	PROP_ALLOW_UPDATES,
 	PROP_NETWORK_AVAILABLE,
+	PROP_COPY_DESTS,
 	PROP_LAST
 };
 
@@ -180,6 +184,17 @@ typedef gboolean	 (*GsPluginUpdateFunc)		(GsPlugin	*plugin,
 							 GError		**error);
 typedef void		 (*GsPluginAdoptAppFunc)	(GsPlugin	*plugin,
 							 GsApp		*app);
+typedef gboolean	 (*GsPluginGetCopyableFunc)	(GsPlugin	*plugin,
+							 GsApp		*app,
+							 const gchar	*copy_dest,
+							 gboolean	*result,
+							 GCancellable	*cancellable,
+							 GError		**error);
+typedef gboolean	 (*GsPluginCopyFunc)		(GsPlugin	*plugin,
+							 GsApp		*app,
+							 const gchar	*copy_dest,
+							 GCancellable	*cancellable,
+							 GError		**error);
 
 /* async helper */
 typedef struct {
@@ -195,6 +210,7 @@ typedef struct {
 	guint				 timeout_id;
 	gboolean			 timeout_triggered;
 	gchar				**tokens;
+	gboolean			 boolean_retval;
 } GsPluginLoaderHelper;
 
 static GsPluginLoaderHelper *
@@ -734,6 +750,24 @@ gs_plugin_loader_call_vfunc (GsPluginLoaderHelper *helper,
 			GsPluginAuthFunc plugin_func = func;
 			ret = plugin_func (plugin,
 					   gs_plugin_job_get_auth (helper->plugin_job),
+					   cancellable, &error_local);
+		}
+		break;
+	case GS_PLUGIN_ACTION_GET_COPYABLE:
+		{
+			const gchar *copy_dest = gs_plugin_job_get_copy_dest (helper->plugin_job);
+			GsPluginGetCopyableFunc plugin_func = func;
+
+			ret = plugin_func (plugin, app, copy_dest,
+					   &helper->boolean_retval,
+					   cancellable, &error_local);
+		}
+		break;
+	case GS_PLUGIN_ACTION_COPY:
+		{
+			GsPluginCopyFunc plugin_func = func;
+			ret = plugin_func (plugin, app,
+					   gs_plugin_job_get_copy_dest (helper->plugin_job),
 					   cancellable, &error_local);
 		}
 		break;
@@ -1674,6 +1708,99 @@ gs_plugin_loader_job_get_categories_finish (GsPluginLoader *plugin_loader,
 	return g_task_propagate_pointer (G_TASK (res), error);
 }
 
+
+static gboolean
+gs_plugin_loader_app_get_copyable_run (GsPluginLoaderHelper *helper,
+				       GCancellable *cancellable,
+				       GError **error)
+{
+	GsPluginLoaderPrivate *priv = gs_plugin_loader_get_instance_private (helper->plugin_loader);
+	GsApp *app = gs_plugin_job_get_app (helper->plugin_job);
+
+	/* run each plugin */
+	for (guint i = 0; i < priv->plugins->len; i++) {
+		GsPlugin *plugin = g_ptr_array_index (priv->plugins, i);
+		if (g_cancellable_set_error_if_cancelled (cancellable, error)) {
+			gs_utils_error_convert_gio (error);
+			return FALSE;
+		}
+
+		helper->function_name = "gs_plugin_app_get_copyable";
+		if (!gs_plugin_loader_call_vfunc (helper, plugin, app, NULL,
+						  cancellable, error)) {
+			return FALSE;
+		}
+		gs_plugin_status_update (plugin, NULL, GS_PLUGIN_STATUS_FINISHED);
+	}
+
+	return TRUE;
+}
+
+static void
+gs_plugin_loader_job_app_get_copyable_thread_cb (GTask *task,
+						 gpointer object,
+						 gpointer task_data,
+						 GCancellable *cancellable)
+{
+	GError *error = NULL;
+	GsPluginLoaderHelper *helper = (GsPluginLoaderHelper *) task_data;
+
+	if (!gs_plugin_loader_app_get_copyable_run (helper, cancellable, &error)) {
+		g_task_return_error (task, error);
+		return;
+	}
+
+	g_task_return_boolean (task, helper->boolean_retval);
+}
+
+/**
+ * gs_plugin_loader_job_app_get_copyable_async:
+ *
+ * This method calls gs_plugin_app_get_copyable() for the @plugin belonging to
+ * @plugin_job if the plugin implements it.
+ **/
+void
+gs_plugin_loader_job_app_get_copyable_async (GsPluginLoader *plugin_loader,
+					     GsPluginJob *plugin_job,
+					     GCancellable *cancellable,
+					     GAsyncReadyCallback callback,
+					     gpointer user_data)
+{
+	GsPluginLoaderHelper *helper;
+	g_autoptr(GTask) task = NULL;
+
+	g_return_if_fail (GS_IS_PLUGIN_LOADER (plugin_loader));
+	g_return_if_fail (GS_IS_PLUGIN_JOB (plugin_job));
+	g_return_if_fail (cancellable == NULL || G_IS_CANCELLABLE (cancellable));
+
+	/* save helper */
+	helper = gs_plugin_loader_helper_new (plugin_loader, plugin_job);
+
+	/* run in a thread */
+	task = g_task_new (plugin_loader, cancellable, callback, user_data);
+	g_task_set_task_data (task, helper, (GDestroyNotify) gs_plugin_loader_helper_free);
+	g_task_run_in_thread (task, gs_plugin_loader_job_app_get_copyable_thread_cb);
+}
+
+/**
+ * gs_plugin_loader_job_app_get_copyable_finish:
+ *
+ * Return value: the value of the plugin's gs_plugin_app_get_copyable(), if it
+ * exists, or %FALSE if not.
+ **/
+gboolean
+gs_plugin_loader_job_app_get_copyable_finish (GsPluginLoader *plugin_loader,
+					      GAsyncResult *res,
+					      GError **error)
+{
+	g_return_val_if_fail (GS_IS_PLUGIN_LOADER (plugin_loader), FALSE);
+	g_return_val_if_fail (G_IS_TASK (res), FALSE);
+	g_return_val_if_fail (g_task_is_valid (res, plugin_loader), FALSE);
+	g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
+
+	return g_task_propagate_boolean (G_TASK (res), error);
+}
+
 /******************************************************************************/
 
 static gboolean
@@ -2336,6 +2463,96 @@ gs_plugin_loader_setup_again (GsPluginLoader *plugin_loader)
 	}
 }
 
+static GMount*
+find_mount_in_list (GList  *mounts, GMount *needle)
+{
+	g_autofree gchar* uuid_needle = g_mount_get_uuid (needle);
+
+	for (GList *l = mounts; l; l = l->next) {
+		g_autofree gchar *uuid_cur = g_mount_get_uuid (l->data);
+
+		if (g_strcmp0 (uuid_needle, uuid_cur) == 0)
+			return l->data;
+	}
+
+	return NULL;
+}
+
+static GList*
+mounts_dup_as_paths (GList *mounts)
+{
+	GList *paths = NULL;
+
+	for (GList *l = mounts; l; l = l->next) {
+		g_autoptr(GFile) root = g_mount_get_root (l->data);
+
+		if (root)
+			paths = g_list_append (paths, g_file_get_path (root));
+	}
+
+	return paths;
+}
+
+static void
+mount_added_cb (GVolumeMonitor *volume_monitor,
+		GMount *mount,
+		GsPluginLoader *plugin_loader)
+{
+	GsPluginLoaderPrivate *priv = gs_plugin_loader_get_instance_private (plugin_loader);
+	g_autoptr(GDrive) drive = g_mount_get_drive (mount);
+	g_autoptr(GMount) existing_mount = NULL;
+
+	if (!g_drive_is_removable (drive))
+		return;
+
+	existing_mount = find_mount_in_list (priv->removable_mounts, mount);
+
+	/* update the list to remove any equivalent mounts and move this to the
+	 * front so we prioritize mounts by recency
+	 */
+	if (existing_mount)
+		priv->removable_mounts = g_list_remove (priv->removable_mounts,
+							existing_mount);
+
+	priv->removable_mounts = g_list_prepend (priv->removable_mounts,
+						 g_object_ref (mount));
+
+	g_object_notify (G_OBJECT (plugin_loader), "copy-dests");
+}
+
+static void
+mount_removed_cb (GVolumeMonitor *volume_monitor,
+		  GMount *mount,
+		  GsPluginLoader *plugin_loader)
+{
+	GsPluginLoaderPrivate *priv = gs_plugin_loader_get_instance_private (plugin_loader);
+	GMount *existing_mount = find_mount_in_list (priv->removable_mounts,
+						     mount);
+	if (existing_mount) {
+		priv->removable_mounts = g_list_remove (priv->removable_mounts,
+							existing_mount);
+		g_object_notify (G_OBJECT (plugin_loader), "copy-dests");
+	}
+}
+
+/**
+ * gs_plugin_loader_get_copy_dests:
+ * @plugin_loader: a #GsPluginLoader
+ *
+ * Get the currently-valid list of copy destinations. Note this list is subject
+ * to change so you should also connect to the "copy-dests-changed" signal to be
+ * informed of updates.
+ *
+ * Returns: (transfer full): a #GList of strings of currently-valid copy
+ * destinations
+ */
+GList *
+gs_plugin_loader_dup_copy_dests (GsPluginLoader *plugin_loader)
+{
+	GsPluginLoaderPrivate *priv = gs_plugin_loader_get_instance_private (plugin_loader);
+	return mounts_dup_as_paths (priv->removable_mounts);
+}
+
 /**
  * gs_plugin_loader_setup:
  * @plugin_loader: a #GsPluginLoader
@@ -2369,6 +2586,7 @@ gs_plugin_loader_setup (GsPluginLoader *plugin_loader,
 	g_autoptr(AsProfileTask) ptask = NULL;
 	g_autoptr(GsPluginLoaderHelper) helper = NULL;
 	g_autoptr(GsPluginJob) plugin_job = NULL;
+	GList *mounts = NULL;
 
 	/* use the default, but this requires a 'make install' */
 	if (priv->locations->len == 0) {
@@ -2391,6 +2609,17 @@ gs_plugin_loader_setup (GsPluginLoader *plugin_loader,
 				  G_CALLBACK (gs_plugin_loader_plugin_dir_changed_cb), plugin_loader);
 		g_ptr_array_add (priv->file_monitors, monitor);
 	}
+
+	priv->volume_monitor = g_volume_monitor_get ();
+	priv->removable_mounts = NULL;
+	g_signal_connect (priv->volume_monitor, "mount-added",
+			  G_CALLBACK (mount_added_cb), plugin_loader);
+	g_signal_connect (priv->volume_monitor, "mount-removed",
+			  G_CALLBACK (mount_removed_cb), plugin_loader);
+	mounts = g_volume_monitor_get_mounts (priv->volume_monitor);
+	for (GList *l = mounts; l; l = l->next)
+		mount_added_cb (priv->volume_monitor, l->data, plugin_loader);
+	g_list_free_full (mounts, g_object_unref);
 
 	/* search for plugins */
 	ptask = as_profile_start_literal (priv->profile, "GsPlugin::setup");
@@ -2721,6 +2950,10 @@ gs_plugin_loader_dispose (GObject *object)
 							      plugin_loader);
 		}
 	}
+	g_signal_handlers_disconnect_by_func (priv->volume_monitor, mount_added_cb, plugin_loader);
+	g_signal_handlers_disconnect_by_func (priv->volume_monitor, mount_removed_cb, plugin_loader);
+	g_clear_object (&priv->volume_monitor);
+	g_list_free_full (priv->removable_mounts, g_object_unref);
 	g_clear_object (&priv->network_monitor);
 	g_clear_object (&priv->soup_session);
 	g_clear_object (&priv->profile);
@@ -2777,6 +3010,9 @@ gs_plugin_loader_class_init (GsPluginLoaderClass *klass)
 				      FALSE,
 				      G_PARAM_READABLE);
 	g_object_class_install_property (object_class, PROP_NETWORK_AVAILABLE, pspec);
+
+	pspec = g_param_spec_pointer ("copy-dests", NULL, NULL, G_PARAM_READABLE);
+	g_object_class_install_property (object_class, PROP_COPY_DESTS, pspec);
 
 	signals [SIGNAL_STATUS_CHANGED] =
 		g_signal_new ("status-changed",
--- a/plugins/flatpak/gs-flatpak.c
+++ b/plugins/flatpak/gs-flatpak.c
@@ -28,6 +28,7 @@
 
 #include <config.h>
 
+#include <sys/signal.h>
 #include <glib/gi18n.h>
 
 #include "gs-appstream.h"
@@ -384,6 +385,20 @@ gs_flatpak_remove_prefixed_names (AsApp *app)
 			continue;
 		as_app_set_name (app, locale, value + 10);
 	}
+}
+
+const gchar*
+gs_flatpak_get_collection_id (GsFlatpak *self,
+                              GsApp     *app)
+{
+	AsApp *as_app = as_store_get_app_by_unique_id (self->store,
+						       gs_app_get_unique_id (app),
+						       AS_STORE_SEARCH_FLAG_USE_WILDCARDS);
+
+	if (as_app == NULL)
+		return NULL;
+
+	return as_app_get_metadata_item (as_app, "flatpak::CollectionId");
 }
 
 static gboolean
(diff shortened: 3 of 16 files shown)


src/gs-details-page.c
```diff
@@ -2334,6 +2559,22 @@ gs_details_page_setup (GsPage *page,
...
+			  G_CALLBACK (gs_details_page_mount_added_cb),
+			  self);
+	g_signal_connect (self->volume_monitor, "mount-removed",
+			  G_CALLBACK (gs_details_page_mount_removed_cb),
+			  self);
+	mounts = g_volume_monitor_get_mounts (self->volume_monitor);
+	for (l = mounts; l; l = l->next) {
+		gs_details_page_mount_added_cb (self->volume_monitor, l->data,
+						self);
+	}
```
joaquimrocha: Nitpick: No need for curly brackets.

joaquimrocha: I also think this is too much responsibility for the details view. Usually views in g-s shouldn't have so much feature-related logic. Instead, this should be in a plugin, which would affect the apps' properties accordingly (like we do for the shortcuts). The EOS plugin also already tracks mountpoints, so this logic should be there, and maybe we just add a `COPIABLE` prop to the app (or do it using the app's metadata if easier), so the details views (and others maybe) will be notified if an app is copiable or not.

src/gs-page.c
```diff
@@ -246,6 +250,36 @@ gs_page_app_removed_cb (GObject *source,
...
+	GsPluginLoader *plugin_loader = GS_PLUGIN_LOADER (source);
+	GsPage *page = helper->page;
+	gboolean ret;
+	g_autoptr(GError) error = NULL;
+
+	ret = gs_plugin_loader_job_action_finish (plugin_loader, res, &error);
+	if (g_error_matches (error,
+			     GS_PLUGIN_ERROR,
+			     GS_PLUGIN_ERROR_CANCELLED)) {
+		g_debug ("%s", error->message);
```
joaquimrocha: Maybe this message needs a bit more context to be useful.

src/gs-page.c
```diff
@@ -246,6 +250,36 @@ gs_page_app_removed_cb (GObject *source,
...
+	}
+	if (!ret) {
+		g_warning ("failed to copy %s: %s", gs_app_get_id (helper->app),
+		           error->message);
+		return;
+	}
+
+	if (GS_PAGE_GET_CLASS (page)->app_copied != NULL) {
+		GS_PAGE_GET_CLASS (page)->app_copied (page, helper->app);
+	}
```
joaquimrocha: Nitpick: No need for curly brackets.

src/gs-page.c
```diff
@@ -481,6 +515,45 @@ gs_page_install_app (GsPage *page,
...
+gs_page_copy_app (GsPage *page,
+		  GsApp *app,
+		  const gchar *dest,
+		  GsShellInteraction interaction,
+		  GCancellable *cancellable)
+{
+	GsPagePrivate *priv = gs_page_get_instance_private (page);
+	GsPageHelper *helper;
+
+	helper = g_slice_new0 (GsPageHelper);
```
joaquimrocha: Nitpick: You can initialize it together with the declaration. And use `g_new0` is probably better too.

src/gs-page.c
```diff
@@ -481,6 +515,45 @@ gs_page_install_app (GsPage *page,
...
+	helper->action = GS_PLUGIN_ACTION_COPY;
+	helper->app = g_object_ref (app);
+	helper->page = g_object_ref (page);
+	helper->cancellable = g_object_ref (cancellable);
+	helper->interaction = interaction;
+
+	if (gs_app_get_state (app) == AS_APP_STATE_INSTALLED) {
```
joaquimrocha: Use `gs_app_is_installed`. Otherwise, if the app is e.g. updatable, then it won't be able to copy it to the USB.

src/gs-page.c
```diff
@@ -481,6 +515,45 @@ gs_page_install_app (GsPage *page,
...
+								  GS_PLUGIN_FAILURE_FLAGS_FATAL_ANY,
+						 NULL);
+
+		gs_plugin_loader_job_process_async (priv->plugin_loader,
+						    plugin_job,
+						    helper->cancellable,
+						    gs_page_app_copied_cb,
+						    helper);
+	} else {
+		// NOTE: it should be impossible to reach this state
```
joaquimrocha: I think it should be unlikely this is reached, but the states here are not atomic, so things like this are usually reachable in g-s :sweat_smile:

plugins/flatpak/gs-flatpak.c
```diff
@@ -3396,6 +3411,182 @@ gs_flatpak_create_app_from_repo_dir (GsFlatpak *self,
...
+gboolean
+gs_flatpak_app_get_copyable (GsFlatpak *self,
+			     GsApp *app,
+			     gboolean *copyable,
+			     GCancellable *cancellable,
+			     GError **error)
+{
+	if (copyable) {
+		*copyable = (gs_flatpak_get_collection_id (self, app) != NULL);
+	}
```
joaquimrocha: Nitpick: No need for curly brackets.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3396,6 +3411,182 @@ gs_flatpak_create_app_from_repo_dir (GsFlatpak *self,
...
+	gulong cancelled_id;
+	gboolean finished;
+	GError *error;
+} GsFlatpakCopyProcessHelper;
+
+static void
+gs_flatpak_copy_process_helper_free (GsFlatpakCopyProcessHelper *helper)
+{
+	g_object_unref (helper->app);
+	g_object_unref (helper->cancellable);
```
joaquimrocha: Nitpick: Maybe it's better/safer to clear it these objects instead?

plugins/flatpak/gs-flatpak.c
```diff
@@ -3396,6 +3411,182 @@ gs_flatpak_create_app_from_repo_dir (GsFlatpak *self,
...
+	return helper;
+}
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC(GsFlatpakCopyProcessHelper, gs_flatpak_copy_process_helper_free)
+
+static void
+copy_process_watch_cb (GPid pid, gint status, gpointer user_data)
+{
+	GsFlatpakCopyProcessHelper *helper =
+		(GsFlatpakCopyProcessHelper*) user_data;
```
joaquimrocha: No need for the cast.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3396,6 +3411,182 @@ gs_flatpak_create_app_from_repo_dir (GsFlatpak *self,
...
+	g_autoptr(GError) error = NULL;
+
+	if (!g_cancellable_is_cancelled (helper->cancellable) && status) {
```
joaquimrocha: Nitpick: I'd explicitly check if the status != 0.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3396,6 +3411,182 @@ gs_flatpak_create_app_from_repo_dir (GsFlatpak *self,
...
+		g_set_error (&helper->error,
+			     GS_PLUGIN_ERROR,
+			     GS_PLUGIN_ERROR_FAILED,
+			     "failed to copy flatpak app to removable "
+			     "media: `flatpak create-usb` failed with "
+			     "status %d", status);
+	}
```
joaquimrocha: Nitpick: No need for curly brackets.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3396,6 +3411,182 @@ gs_flatpak_create_app_from_repo_dir (GsFlatpak *self,
...
+
+	// update metadata to ensure the USB category includes this new app and
+	// we can immediately download an app, copy to USB, delete from
+	// computer, and install from USB without leaving the app page
+	gs_flatpak_refresh (helper->self, 0, GS_PLUGIN_REFRESH_FLAGS_METADATA,
+			    NULL, &error);
+	if (error != NULL) {
+		g_warning ("failed to refresh flatpak metadata after copying "
+			"app to USB: %s", error->message);
+	}
```
joaquimrocha: Nitpick: No need for curly brackets.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3396,6 +3411,182 @@ gs_flatpak_create_app_from_repo_dir (GsFlatpak *self,
...
+				 GS_PLUGIN_STATUS_FINISHED);
+
+	// update metadata to ensure the USB category includes this new app and
+	// we can immediately download an app, copy to USB, delete from
+	// computer, and install from USB without leaving the app page
+	gs_flatpak_refresh (helper->self, 0, GS_PLUGIN_REFRESH_FLAGS_METADATA,
+			    NULL, &error);
+	if (error != NULL) {
+		g_warning ("failed to refresh flatpak metadata after copying "
+			"app to USB: %s", error->message);
```
joaquimrocha: Alignment looks odd(er than usual), but maybe it's still Github's view.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3396,6 +3411,182 @@ gs_flatpak_create_app_from_repo_dir (GsFlatpak *self,
...
+
+	/* once the copy terminates (successfully or not), set plugin status to
+	 * update UI accordingly */
+	gs_plugin_status_update (helper->self->plugin, helper->app,
+				 GS_PLUGIN_STATUS_FINISHED);
+
+	// update metadata to ensure the USB category includes this new app and
+	// we can immediately download an app, copy to USB, delete from
+	// computer, and install from USB without leaving the app page
+	gs_flatpak_refresh (helper->self, 0, GS_PLUGIN_REFRESH_FLAGS_METADATA,
```
joaquimrocha: Are you sure this is the best way to do this? I think the refresh lists the apps in the repo which is quite slow, so maybe we should just modify the app as needed?

plugins/flatpak/gs-flatpak.c
```diff
@@ -3396,6 +3411,182 @@ gs_flatpak_create_app_from_repo_dir (GsFlatpak *self,
...
+	helper->finished = TRUE;
+
+	/* once the copy terminates (successfully or not), set plugin status to
+	 * update UI accordingly */
+	gs_plugin_status_update (helper->self->plugin, helper->app,
+				 GS_PLUGIN_STATUS_FINISHED);
+
+	// update metadata to ensure the USB category includes this new app and
+	// we can immediately download an app, copy to USB, delete from
+	// computer, and install from USB without leaving the app page
```
joaquimrocha: Use C style comments.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3396,6 +3411,182 @@ gs_flatpak_create_app_from_repo_dir (GsFlatpak *self,
...
+	}
+}
+
+static void
+app_copy_cancelled_cb (GCancellable *cancellable,
+		       gpointer user_data)
+{
+	GPid pid = GPOINTER_TO_INT (user_data);
+
+	// terminate the process which is copying the app
```
joaquimrocha: Use C style comments.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3396,6 +3411,182 @@ gs_flatpak_create_app_from_repo_dir (GsFlatpak *self,
...
+
+gboolean
+gs_flatpak_app_copy (GsFlatpak *self,
+		     GsApp *app,
+		     const gchar *dest,
+		     GCancellable *cancellable,
+		     GError **error)
+{
+	// this is used in an async function but we block here until that
+	// returns so we won't auto-free while other threads depend on this
```
joaquimrocha: Use C style comments.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3396,6 +3411,182 @@ gs_flatpak_create_app_from_repo_dir (GsFlatpak *self,
...
+	gboolean spawn_retval;
+	g_autofree GStrv argv;
+	GPid child_pid;
+	gulong cancelled_id;
+	const gchar* app_collection_id;
+
+	app_id = gs_flatpak_app_get_ref_name (app);
+	app_arch = gs_flatpak_app_get_ref_arch (app);
+	app_branch = gs_flatpak_app_get_ref_branch (app);
+	app_ref = g_strdup_printf ("%s/%s/%s", app_id,
```
joaquimrocha: Nitpick: Maybe set all these together with their declaration?

plugins/flatpak/gs-flatpak.c
```diff
@@ -3396,6 +3411,182 @@ gs_flatpak_create_app_from_repo_dir (GsFlatpak *self,
...
+						     app_arch, app_branch);
+	app_collection_id = gs_flatpak_get_collection_id (self, app);
+	g_debug ("Copying app %s with collection ID: %s", app_ref,
+		 (app_collection_id ? app_collection_id : "(null)"));
```
joaquimrocha: I think that in `g_debug` a NULL string arg will already show up as `(null)`.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3396,6 +3411,182 @@ gs_flatpak_create_app_from_repo_dir (GsFlatpak *self,
...
+
+	argv = g_malloc (sizeof (gchar*) * 5);
```
joaquimrocha: You should declare a constant `GStrv` here. See https://github.com/endlessm/gnome-software/blob/master/src/gs-category-page.c#L411 .

plugins/flatpak/gs-flatpak.c
```diff
@@ -3396,6 +3411,182 @@ gs_flatpak_create_app_from_repo_dir (GsFlatpak *self,
...
+						      G_CALLBACK (app_copy_cancelled_cb),
+						      GINT_TO_POINTER (child_pid),
+						      NULL);
+
+		helper = gs_flatpak_copy_process_helper_new (self, app,
+							     cancellable,
+							     cancelled_id);
+		g_child_watch_add (child_pid, copy_process_watch_cb, helper);
+	} else if (error && *error) {
+		return FALSE;
```
joaquimrocha: I think the return value should be dependent on the passed error arg.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3396,6 +3411,182 @@ gs_flatpak_create_app_from_repo_dir (GsFlatpak *self,
...
+	}
+
+	// Iterate the main loop until either the copy process completes or the
+	// user cancels the copy. Without this, it is impossible to cancel the
+	// copy because we reach the end of this function, its parent GTask
+	// returns and we disconnect the handler that would kill the copy
+	// process.
```
joaquimrocha: Use C style comments.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3396,6 +3411,182 @@ gs_flatpak_create_app_from_repo_dir (GsFlatpak *self,
...
+	while (!helper->finished) {
+		g_main_context_iteration (NULL, FALSE);
+	}
```
joaquimrocha: Nitpick: No need for curly brackets.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3396,6 +3411,182 @@ gs_flatpak_create_app_from_repo_dir (GsFlatpak *self,
...
+		return FALSE;
+	}
+
+	// Iterate the main loop until either the copy process completes or the
+	// user cancels the copy. Without this, it is impossible to cancel the
+	// copy because we reach the end of this function, its parent GTask
+	// returns and we disconnect the handler that would kill the copy
+	// process.
+	while (!helper->finished) {
+		g_main_context_iteration (NULL, FALSE);
```
joaquimrocha: This will be executed from a different thread. Is this fine?

src/gs-details-page.c
```diff
@@ -446,31 +446,28 @@ gs_details_page_refresh_progress (GsDetailsPage *self)
...
 				     _("Installing"));
-		break;
-	default:
+	} else {
 		gtk_widget_set_visible (self->label_progress_status, FALSE);
-		break;
 	}
 	if (app_has_pending_action (self->app)) {
 		GsPluginAction action = gs_app_get_pending_action (self->app);
+
```
joaquimrocha: Why the extra newline in this commit?

src/gs-details-page.c
```diff
@@ -446,31 +446,28 @@ gs_details_page_refresh_progress (GsDetailsPage *self)
 		gtk_widget_set_visible (self->button_cancel, FALSE);
 		break;
 	}
-	if (app_has_pending_action (self->app)) {
+	if (plugin_or_app_has_pending_action (self, self->app)) {
```
joaquimrocha: Commit titles in g-s (usually) only start by the module when it's a change in the plugins. Also seems like this commit should rather be squashed into the others you did before?

joaquimrocha:
Reviewed and wrote a few comments about the code and nitpicks. I think the approach here is putting too much feature-related code into the details view, instead of using our plugin.

lib/gs-plugin-job.c
```diff
@@ -67,6 +68,7 @@ enum {
 	PROP_MAX_RESULTS,
 	PROP_PRICE,
 	PROP_TIMEOUT,
+	PROP_DEST,
```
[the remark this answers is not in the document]
treitter: I've fixed the name locally.

I want the flexibility to change the destination on a per-job basis, I think, since the path can change as USB disks are added and removed and, in the future, I'll want to add the ability for the user to select the destination. Right now, it just picks the most-recently-mounted removable disk which is kind of crude.

Given those needs, would it still make sense to do this in the plugin loader?

lib/gs-plugin-job.c
```diff
@@ -103,6 +105,10 @@ gs_plugin_job_to_string (GsPluginJob *self)
 		g_string_append_printf (str, " with search=%s",
 					self->search);
 	}
+	if (self->dest != NULL) {
+		g_string_append_printf (str, " with destination=%s",
+					self->dest);
+	}
```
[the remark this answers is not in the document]
treitter: I changed this to "copy destination" locally.

lib/gs-plugin-loader.c
```diff
@@ -1674,6 +1704,100 @@ gs_plugin_loader_job_get_categories_finish (GsPluginLoader *plugin_loader,
...
+	GsPluginLoaderHelper *helper;
+	g_autoptr(GTask) task = NULL;
+
+	g_return_if_fail (GS_IS_PLUGIN_LOADER (plugin_loader));
+	g_return_if_fail (GS_IS_PLUGIN_JOB (plugin_job));
+	g_return_if_fail (cancellable == NULL || G_IS_CANCELLABLE (cancellable));
+
+	/* save helper */
+	helper = gs_plugin_loader_helper_new (plugin_loader, plugin_job);
+	helper->catlist = NULL;
```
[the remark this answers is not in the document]
treitter: I wanted to be sure we didn't try freeing uninitialized memory later but forgot I'm doing that in `gs_plugin_loader_helper_new()`, so I've removed this locally.

lib/gs-plugin-loader.c
```diff
@@ -3098,6 +3222,7 @@ gs_plugin_loader_generic_update (GsPluginLoader *plugin_loader,
 			gs_plugin_loader_action_start (plugin_loader, plugin, FALSE);
 			ret = plugin_app_func (plugin, app, app_cancellable, &error_local);
 			gs_plugin_loader_action_stop (plugin_loader, plugin);
+
```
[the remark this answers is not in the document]
treitter: Fixed locally

src/gs-details-page.c
```diff
@@ -238,6 +243,105 @@ gs_details_page_update_shortcut_button (GsDetailsPage *self)
...
+	GsDetailsPage *self = GS_DETAILS_PAGE (user_data);
+	g_autoptr(GError) error = NULL;
+	gboolean copyable;
+
+	copyable = gs_plugin_loader_job_app_get_copyable_finish (plugin_loader,
+								 res, &error);
+
+	if (!copyable) {
+		gtk_widget_set_sensitive (self->button_copy, FALSE);
+		gtk_widget_set_visible (self->button_copy, FALSE);
```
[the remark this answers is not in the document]
treitter: Fixed locally

src/gs-details-page.c
```diff
@@ -238,6 +243,105 @@ gs_details_page_update_shortcut_button (GsDetailsPage *self)
...
+	AsAppState state;
+	g_autofree gchar* mount_path = NULL;
+	g_autoptr(GsPluginJob) plugin_job = NULL;
+
+	gtk_widget_set_sensitive (self->button_copy, FALSE);
+	gtk_widget_set_visible (self->button_copy, FALSE);
+
+	if (self->plugin_status == GS_PLUGIN_STATUS_COPYING) {
+		return;
+	}
```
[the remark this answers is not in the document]
treitter: Fixed locally

src/gs-details-page.c
```diff
@@ -238,6 +243,105 @@ gs_details_page_update_shortcut_button (GsDetailsPage *self)
...
+	case AS_APP_STATE_UNKNOWN:
+	case AS_APP_STATE_REMOVING:
+	case AS_APP_STATE_QUEUED_FOR_INSTALL:
+		break;
+	default:
+		g_warning ("App unexpectedly in state %s",
+			   as_app_state_to_string (state));
+		g_assert_not_reached ();
+	}
+}
```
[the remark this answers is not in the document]
treitter: Fixed locally with a simple call to `gs_app_is_installed()`.

src/gs-details-page.c
```diff
@@ -253,6 +357,13 @@ app_has_pending_action (GsApp *app)
 	       (gs_app_get_state (app) == AS_APP_STATE_QUEUED_FOR_INSTALL);
 }
 
+static gboolean
+plugin_or_app_has_pending_action (GsDetailsPage *self)
```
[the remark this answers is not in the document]
treitter: Fixed locally

src/gs-details-page.c
```diff
@@ -532,8 +651,8 @@ gs_details_page_refresh_progress (GsDetailsPage *self)
 		gtk_widget_set_visible (self->box_progress, FALSE);
 		break;
 	}
-	if (plugin_or_app_has_pending_action (self, self->app)) {
-		gtk_widget_set_visible (self->box_progress, TRUE);
+	if (plugin_or_app_has_pending_action (self)) {
+	 	gtk_widget_set_visible (self->box_progress, TRUE);
```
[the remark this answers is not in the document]
treitter: The perils of mixing tabs and spaces :)

Fixed locally.

src/gs-details-page.c
```diff
@@ -2079,6 +2210,71 @@ gs_details_page_write_review_cb (GtkButton *button,
...
+{
+	GList *l;
+	g_autofree gchar* uuid_needle = g_mount_get_uuid (needle);
+
+	for (l = mounts; l; l = l->next) {
+		g_autofree gchar *uuid_cur = g_mount_get_uuid (l->data);
+
+		if (g_strcmp0 (uuid_needle, uuid_cur) == 0) {
+			return l->data;
+		}
```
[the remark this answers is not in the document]
treitter: Fixed locally

src/gs-details-page.c
```diff
@@ -2079,6 +2210,71 @@ gs_details_page_write_review_cb (GtkButton *button,
...
 }
 
+static gconstpointer
+find_mount_in_list (GList  *mounts,
+                    GMount *needle)
+{
+	GList *l;
+	g_autofree gchar* uuid_needle = g_mount_get_uuid (needle);
+
+	for (l = mounts; l; l = l->next) {
```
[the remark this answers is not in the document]
treitter: I did that originally but saw other code that did it this way. So, I was actually trying to follow upstream's style :) I guess I just didn't look for enough examples.

Fixed locally.

src/gs-details-page.c
```diff
@@ -2079,6 +2210,71 @@ gs_details_page_write_review_cb (GtkButton *button,
...
+				GsDetailsPage *self)
+{
+	g_autoptr(GDrive) drive = g_mount_get_drive (mount);
+	g_autoptr(GFile) root = g_mount_get_root (mount);
+	g_autofree gchar *path = g_file_get_path (root);
+	g_autoptr(GMount) mount_existing = NULL;
+
+	if (!g_drive_is_removable (drive)) {
+		return;
+	}
```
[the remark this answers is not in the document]
treitter: Fixed locally

src/gs-details-page.c
```diff
@@ -2079,6 +2210,71 @@ gs_details_page_write_review_cb (GtkButton *button,
 	gs_shell_modal_dialog_present (self->shell, GTK_DIALOG (dialog));
 }
 
+static gconstpointer
```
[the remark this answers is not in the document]
treitter: Fixed locally

src/gs-details-page.c
```diff
@@ -2079,6 +2210,71 @@ gs_details_page_write_review_cb (GtkButton *button,
...
+
+static void
+gs_details_page_mount_added_cb (GVolumeMonitor *volume_monitor,
+				GMount *mount,
+				GsDetailsPage *self)
+{
+	g_autoptr(GDrive) drive = g_mount_get_drive (mount);
+	g_autoptr(GFile) root = g_mount_get_root (mount);
+	g_autofree gchar *path = g_file_get_path (root);
+	g_autoptr(GMount) mount_existing = NULL;
```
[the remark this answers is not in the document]
treitter: I usually name variables `noun_adjective` since the noun is usually more important and it makes similar variables easier to visually group (eg, if we were re-mounting a device in a new location and had "mount_existing" and "mount_new").

But it's also not a big deal to me and it matters less when there's a single variable.

Fixed locally.

src/gs-details-page.c
```diff
@@ -2079,6 +2210,71 @@ gs_details_page_write_review_cb (GtkButton *button,
...
+
+	if (!g_drive_is_removable (drive)) {
+		return;
+	}
+
+	mount_existing = (GMount*) find_mount_in_list (self->removable_mounts,
+						       mount);
```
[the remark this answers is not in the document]
treitter: Fixed locally

src/gs-details-page.c
```diff
@@ -2079,6 +2210,71 @@ gs_details_page_write_review_cb (GtkButton *button,
...
+{
+	g_autoptr(GFile) root = g_mount_get_root (mount);
+	g_autofree gchar *path = g_file_get_path (root);
+	GMount *mount_existing =
+		(GMount*) find_mount_in_list (self->removable_mounts, mount);
+
+	if (mount_existing) {
+		self->removable_mounts = g_list_remove (self->removable_mounts,
+							mount_existing);
+	}
```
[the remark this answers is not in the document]
treitter: I get a little itchy when an unbraced body spans multiple lines in C but if that's the upstream style, I'll follow it.

src/gs-details-page.c
```diff
@@ -2091,6 +2287,16 @@ gs_details_page_app_removed (GsPage *page, GsApp *app)
 	gs_details_page_reload (page);
 }
 
+static void
+gs_details_page_app_copied (GsPage *page, GsApp *app)
+{
+	GsDetailsPage *self = GS_DETAILS_PAGE (page);
+
+	gtk_button_set_label (GTK_BUTTON (self->button_copy),
+			      _("Copied to USB"));
```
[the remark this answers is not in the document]
treitter: Not with 80-character lines. I guess upstream isn't very strict or does it just have a higher limit? 100 characters?

src/gs-details-page.c
```diff
@@ -2315,6 +2521,23 @@ gs_details_page_network_available_notify_cb (GsPluginLoader *plugin_loader,
...
+                                          GsApp *app,
+                                          GsPluginStatus status,
+                                          GsDetailsPage *self)
+{
+	if (status == GS_PLUGIN_STATUS_COPYING) {
+		/* prevent the app from being deleted, etc. while it's being copied */
+		gs_app_add_quirk (app, AS_APP_QUIRK_COMPULSORY);
+	} else {
+		/* allow removal once copying has completed */
+		gs_app_remove_quirk (app, AS_APP_QUIRK_COMPULSORY);
```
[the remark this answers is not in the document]
treitter: When I discussed making sure the app couldn't be deleted during a copy, Richard told me to be sure to set this.

The one concern I did have is that any "clean up" code that unsets this does so absolutely. So, it's like having two critical sections of code each guarded by a "mutex" that either can completely unlock on their own.

Maybe there's some reason why this isn't the case for quirks but it seems like an issue and there's no obvious way to solve it short of giving them ref counters.

src/gs-page.c
treitter: I prefixed with "app copy cancelled: "

src/gs-page.c
treitter: Fixed locally

src/gs-page.c
treitter: I've merged the lines locally.

But the free function uses `g_slice_free()` and all the other instances of `GsPageHelper` allocate it with `g_slice_new0()` so we'd need to change those as well and I don't see a critical need to do that.

Why do you prefer `g_new0()`? Because it's less of a risk when debugging and the performance gains of `g_slice` allocation are trivial here?

src/gs-page.c
treitter: Good catch. Fixed locally

plugins/flatpak/gs-flatpak.c
treitter: Fixed locally

plugins/flatpak/gs-flatpak.c
treitter: Sure. Fixed locally.

plugins/flatpak/gs-flatpak.c
treitter: Fixed locally

src/gs-details-page.c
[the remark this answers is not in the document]
joaquimrocha: I'd prefer if there was a limit, but there's actually no limit enforced, as long as it's "reasonable".

src/gs-details-page.c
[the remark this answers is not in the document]
joaquimrocha: Yeah, achieving the same effect with a different solution would probably be too complex anyway as you say. Let's leave it then.

src/gs-page.c
joaquimrocha: I read somewhere that `g_slice` was no longer a faster solution. But I hadn't noticed the free func, so let's keep it then.

plugins/flatpak/gs-flatpak.c
treitter: Fixed locally

plugins/flatpak/gs-flatpak.c
treitter: Fixed locally

plugins/flatpak/gs-flatpak.c
treitter: Fixed locally

plugins/flatpak/gs-flatpak.c
treitter: it was one tab in from the start of `g_warning` so I added the missing spaces locally.

plugins/flatpak/gs-flatpak.c
treitter: Fixed locally

plugins/flatpak/gs-flatpak.c
treitter: Fixed locally

plugins/flatpak/gs-flatpak.c
treitter: Fixed locally

plugins/flatpak/gs-flatpak.c
treitter: I thought it was convention to have no assignments (except for `NULL`) at the beginning of a block but maybe I was confusing `-Wdeclaration-after-statement` (which doesn't mind assignments mixed into the declaration section after all).

Fixed locally.

plugins/flatpak/gs-flatpak.c
treitter: I think you're right.

Fixed.

plugins/flatpak/gs-flatpak.c
treitter: Good idea.

Fixed locally.

plugins/flatpak/gs-flatpak.c
treitter: Do you mean independent? Because that makes sense.

Fixed locally.

plugins/flatpak/gs-flatpak.c
treitter: Fixed locally

plugins/flatpak/gs-flatpak.c
treitter: Fixed locally

plugins/flatpak/gs-flatpak.c
treitter: I think that's fine as long as we block here until `copy_process_watch_cb()` executes and breaks the conditional which is what I intended.

Do you see any potential races?

src/gs-details-page.c
treitter: No reason. Fixed locally.

src/gs-details-page.c
treitter: Fixed locally.

This conditional only needs to change when we support copying so I don't think it's necessarily better to move the support function to the previous commit given it isn't used yet by then (and would trigger an "unused function" warning when building at that point).

src/gs-details-page.c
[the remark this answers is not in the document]
treitter: OK. I joined the line.

treitter:
This is ready for a re-review.

The latest substantial change is that I moved the mount monitoring code from `details-page` to `plugin-loader`.

I also squashed my trivial fixes into two fixup commits to hopefully make this next review a little easier. I'll squash them before the final merge.

As discussed, this defers relocating the mount management to plugins and reporting copyability as a quirk. We'll do that at the time we support users selecting the target destination themselves.

src/gs-details-page.c
```diff
@@ -651,8 +626,8 @@ gs_details_page_refresh_progress (GsDetailsPage *self)
 		gtk_widget_set_visible (self->box_progress, FALSE);
 		break;
 	}
-	if (plugin_or_app_has_pending_action (self)) {
-	 	gtk_widget_set_visible (self->box_progress, TRUE);
+	if (plugin_has_pending_action (self)) {
+		gtk_widget_set_visible (self->box_progress, TRUE);
```
joaquimrocha: Nitpick: The curly braces.

src/gs-details-page.c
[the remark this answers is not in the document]
joaquimrocha: Yeah, let's change it if you don't mind.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3566,18 +3553,16 @@ gs_flatpak_app_copy (GsFlatpak *self,
 							     cancellable,
 							     cancelled_id);
 		g_child_watch_add (child_pid, copy_process_watch_cb, helper);
-	} else if (error && *error) {
+	} else
```
joaquimrocha: So, the thing about not having curly braces on one-liners is that if it's an if/else block where the counterpart has curly braces you should do it too... (Sorry if this is too nitpicky but I think this is how it usually goes upstream)

plugins/flatpak/gs-flatpak.c
joaquimrocha: I don't know exactly. I'd have to probe it, just wanted to make sure this is checked for races.

src/gs-details-page.c
```diff
@@ -2185,65 +2179,21 @@ gs_details_page_write_review_cb (GtkButton *button,
...
-	 */
-	if (existing_mount)
-		self->removable_mounts = g_list_remove (self->removable_mounts,
-							existing_mount);
-
-	self->removable_mounts = g_list_prepend (self->removable_mounts,
-						 g_object_ref (mount));
-
-	gs_details_page_update_copy_button (self);
+	if (self->copy_dests)
```
joaquimrocha: Check if `!= NULL` (upstream style...).

src/gs-details-page.c
```diff
@@ -286,9 +280,9 @@ gs_details_page_update_copy_button (GsDetailsPage *self)
 	if (self->plugin_status == GS_PLUGIN_STATUS_COPYING)
 		return;
 
-	mount_path = removable_destination_dup (self);
+	copy_dest = removable_destination_dup (self);
 	if (gs_app_is_installed (self->app)) {
-		if (mount_path) {
+		if (copy_dest) {
```
joaquimrocha: Check if `!= NULL` (upstream style).

lib/gs-plugin-loader.c
```diff
@@ -2459,6 +2463,96 @@ gs_plugin_loader_setup_again (GsPluginLoader *plugin_loader)
...
+static GList*
+mounts_dup_as_paths (GList *mounts)
+{
+	GList *paths = NULL;
+
+	for (GList *l = mounts; l; l = l->next) {
+		g_autoptr(GFile) root = g_mount_get_root (l->data);
+
+		if (root)
```
joaquimrocha: Check if `!= NULL` (upstream style).

plugins/flatpak/gs-flatpak.c
```diff
@@ -3396,6 +3411,167 @@ gs_flatpak_create_app_from_repo_dir (GsFlatpak *self,
...
+	const gchar *app_id = gs_flatpak_app_get_ref_name (app);
+	const gchar *app_arch = gs_flatpak_app_get_ref_arch (app);
+	const gchar *app_branch = gs_flatpak_app_get_ref_branch (app);
+	g_autofree gchar* app_ref = g_strdup_printf ("%s/%s/%s", app_id,
+						     app_arch, app_branch);
+	const gchar* app_collection_id = gs_flatpak_get_collection_id (self,
+								       app);
+
+	gboolean spawn_retval;
+	const gchar *argv[] = {"/usr/bin/flatpak", "create-usb", copy_dest,
```
mwleeds: Isn't it bad to assume flatpak was installed into that location, especially if this will be upstreamed? GNOME Builder seems to not need to do that: https://gitlab.gnome.org/GNOME/gnome-builder/blob/master/src/plugins/flatpak/gbp-flatpak-runtime.c#L119

mwleeds: Although in that case it's making sure `PATH` is set. So it might be better to use `which`.

mwleeds: I still don't think this is the right approach (for example on NixOS binaries are installed into `/root/.nix-profile/bin/` rather than `/usr/bin/`) but there is precedent for it in the codebase (search for `DPKG_DEB_BINARY`) so I guess leave it and see what upstream says.

PR #392, opened 2018-08-20
flatpak: Remove Blacklisting from USB Apps

mazen-asef:
Apps that are not on the native language of the system will be blacklisted by
gnome-software. Removing this category when it reads from an USB media and
when it verifies if an app is valid fixes the issue.

https://phabricator.endlessm.com/T23409

### DIFF
--- a/plugins/eos/gs-plugin-eos.c
+++ b/plugins/eos/gs-plugin-eos.c
@@ -896,7 +896,8 @@ gs_plugin_update_locale_cache_app (GsPlugin *plugin,
 	if (is_same_app (cached_app, app))
 		return;
 
-	if (cached_app && !gs_app_is_installed (cached_app)) {
+	if (cached_app && !gs_app_is_installed (cached_app) && 
+	    !gs_app_has_category (cached_app, "USB")) {
 		const char *app_id = gs_app_get_unique_id (app);
 		const char *cached_app_id = gs_app_get_unique_id (cached_app);
 
@@ -934,7 +935,8 @@ gs_plugin_eos_blacklist_kapp_if_needed (GsPlugin *plugin, GsApp *app)
 	/* last token may be the locale */
 	last_token = tokens[num_tokens - 1];
 
-	if (!gs_plugin_locale_is_compatible (plugin, last_token)) {
+	if (!gs_plugin_locale_is_compatible (plugin, last_token) &&
+	    !gs_app_has_category (app, "USB")) {
 		if (gs_app_is_installed (app))
 			return FALSE;
 
@@ -953,7 +955,8 @@ gs_plugin_eos_blacklist_kapp_if_needed (GsPlugin *plugin, GsApp *app)
 
 	/* skip if the cached app is already our best */
 	if (cached_app &&
-	    gs_plugin_app_is_locale_best_match (plugin, cached_app)) {
+	    gs_plugin_app_is_locale_best_match (plugin, cached_app)  && 
+	    !gs_app_has_category (cached_app, "USB")) {
 		if (!gs_app_is_installed (app)) {
 			g_debug ("Blacklisting '%s': cached app '%s' is best "
 				 "match", gs_app_get_unique_id (app),
@@ -1339,6 +1342,18 @@ gs_plugin_eos_blacklist_app_for_remote_if_needed (GsPlugin *plugin,
 		gs_app_add_category (app, "Blacklisted");
 
 	return do_blacklist;
+}
+
+static void
+gs_plugin_eos_remove_blacklist_from_usb_if_needed (GsPlugin *plugin,
+						  GsApp *app)
+{
+	if (!gs_app_has_category (app, "Blacklisted") || 
+	    !gs_app_has_category (app, "USB"))
+		return;
+
+	g_debug ("Removing blacklisting from '%s': app is from USB", gs_app_get_unique_id (app));
+	gs_app_remove_category (app, "Blacklisted");
 }
 
 static gboolean
@@ -1728,6 +1743,8 @@ gs_plugin_refine (GsPlugin		*plugin,
 
 		if (gs_plugin_eos_blacklist_app_for_remote_if_needed (plugin, app))
 			continue;
+		
+		gs_plugin_eos_remove_blacklist_from_usb_if_needed (plugin, app);
 
 		gs_plugin_eos_refine_popular_app (plugin, app);
 	}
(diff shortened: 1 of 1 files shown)


lib/gs-plugin-loader.c
```diff
@@ -1250,8 +1250,9 @@ gs_plugin_loader_app_is_valid (GsApp *app, gpointer user_data)
 		return FALSE;
 	}
 
-	/* don't show blacklisted apps */
-	if (gs_app_has_category (app, "Blacklisted")) {
+	/* don't show blacklisted apps if they are not from USB*/
+	if (gs_app_has_category (app, "Blacklisted") &&
+		!gs_app_has_category (app, "USB")) {
```
joaquimrocha: I don't think we should add changes to the plugin loader for this. Even though this is a simpler change, every change we add to the plugin loader will make it more complex and make rebases more difficult.

joaquimrocha: Thinking better about it. I could be okay with this change if adding it to the EOS plugin is much more complex. i.e. Have you verified if overriding `gs_plugin_add_category_apps` and modifying the `gs_plugin_refine` makes it too complex?

plugins/flatpak/gs-flatpak.c
```diff
@@ -432,6 +432,11 @@ gs_flatpak_mark_apps_from_usb_remote (GsFlatpak *self,
 		 * so we add it to the GsApp here too */
 		gs_app_add_category (app, "USB");
 
+		/* if the category of the app is Blacklisted, remove it */
+                if (gs_app_has_category (app, "Blacklisted")) {
+			gs_app_remove_category (app, "Blacklisted");
+		}
+
```
joaquimrocha: Again, I think it can all be done just from the EOS plugin. I am okay leaving this if the changes to the EOS plugin are too complex/error-prone. In which case you should also remove the curly brackets here (to make it more consistent with what we usually try to have as a coding style).

lib/gs-plugin-loader.c
mazen-asef: Will try that!

plugins/eos/gs-plugin-eos.c
```diff
@@ -1341,6 +1347,22 @@ gs_plugin_eos_blacklist_app_for_remote_if_needed (GsPlugin *plugin,
 	return do_blacklist;
 }
 
+static gboolean
+gs_plugin_eos_remove_blacklist_from_usb_if_needed (GsPlugin *plugin,
+						  GsApp *app)
+{
+	if (gs_app_has_category (app, "Blacklisted") &&
+	    gs_app_has_category (app, "USB")) {
+		g_debug ("Removing blacklisting from '%s': app is from "
```
treitter: Nitpick: capitalize "USB" in this message and join it with the string on the line above since I think that fits easily within the roughly 100+-line wrap of this code

plugins/eos/gs-plugin-eos.c
```diff
@@ -779,6 +781,7 @@ gs_plugin_initialize (GsPlugin *plugin)
 	/* we already deal with apps that need to be proxied, so let's impede
 	 * the other plugin from running */
 	gs_plugin_add_rule (plugin, GS_PLUGIN_RULE_CONFLICTS, "generic-updates");
+	priv->flatpaks = g_ptr_array_new_with_free_func ((GDestroyNotify) g_object_unref);
```
treitter: I just noticed this doesn't get freed - see how it's [done in the Flatpak plugin](https://github.com/endlessm/gnome-software/blob/master/plugins/flatpak/gs-plugin-flatpak.c#L95) in `eos_plugin_destroy()`.

mazen-asef:
Removed the copy from gs_plugin_add_category_apps from flatpak and just made sure to update the list with the new USB category and remove the Blacklisted catefory. Once gs_plugin_refine runs again, it will add the blacklisted category if needed.

plugins/eos/gs-plugin-eos.c
```diff
@@ -1728,6 +1746,9 @@ gs_plugin_refine (GsPlugin		*plugin,
 
 		if (gs_plugin_eos_blacklist_app_for_remote_if_needed (plugin, app))
 			continue;
+		
+		if (gs_plugin_eos_remove_blacklist_from_usb_if_needed (plugin, app))
+			continue;
```
joaquimrocha: We shouldn't continue here, because the app may need to be refined as popular, right? The other blocks call continue because they've just blacklisted an app, so there's no point in wasting time refining it.

plugins/eos/gs-plugin-eos.c
```diff
@@ -2028,6 +2049,24 @@ add_updates (GsPlugin *plugin,
...
+			     GsAppList *list,
+			     GCancellable *cancellable,
+			     GError **error)
+{
+	for (guint i = 0; i < gs_app_list_length (list); ++i) {
+		GsApp *app = gs_app_list_index (list, i);
+		
+		if (gs_app_has_category (app, "USB") && 
+		    gs_app_has_category (app, "Blacklisted"))
+			gs_app_remove_category (app, "Blacklisted");
```
joaquimrocha: Use the `gs_plugin_eos_remove_blacklist_from_usb_if_needed` function you added?

plugins/eos/gs-plugin-eos.c
```diff
@@ -1341,6 +1344,21 @@ gs_plugin_eos_blacklist_app_for_remote_if_needed (GsPlugin *plugin,
...
+gs_plugin_eos_remove_blacklist_from_usb_if_needed (GsPlugin *plugin,
+						  GsApp *app)
+{
+	if (gs_app_has_category (app, "Blacklisted") &&
+	    gs_app_has_category (app, "USB")) {
+		g_debug ("Removing blacklisting from '%s': app is from USB", gs_app_get_unique_id (app));
+		gs_app_remove_category (app, "Blacklisted");
+
+		return TRUE;
+	}
```
joaquimrocha: Nitpick: This is not too big a block, but we usually prefer early returns as it makes the code less hairy, especially if more logic is added in the future. (Remember to reomve the curly braces if you invert the returns)

plugins/eos/gs-plugin-eos.c
```diff
@@ -1341,6 +1344,20 @@ gs_plugin_eos_blacklist_app_for_remote_if_needed (GsPlugin *plugin,
 	return do_blacklist;
 }
 
+static void
+gs_plugin_eos_remove_blacklist_from_usb_if_needed (GsPlugin *plugin,
+						  GsApp *app)
+{
+	if (!gs_app_has_category (app, "Blacklisted"))
+		return;
+	else if (gs_app_has_category (app, "Blacklisted") &&
```
joaquimrocha: You don't need to check for this because you already have checked it before. It should be something like:
```
if (!gs_app_has_category (app, "Blacklisted") || !gs_app_has_category (app, "USB"))

PR #396, opened 2018-08-22
fix: Trailing whitespaces in gs-plugin-eos

mazen-asef:


ramcq:
Yeah maybe even doing a git commit --fixup - then it will appear 100% clearly as a patch to squash with the previous.

joaquimrocha:
In this case (since it's just trailing spaces) I agree with @ramcq . Let's go for a fixup.

PR #397, opened 2018-08-22
T23651

mwleeds:


mwleeds:
I have tested this and it seems to work as intended.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2249,6 +2249,11 @@ gs_flatpak_set_app_metadata (GsFlatpak *self,
 	}
 	g_debug ("runtime for %s is %s", name, runtime);
 
+	if (g_key_file_has_group (kf, "Extra Data"))
```
treitter: Nitpick: this could be a single line instead:

```
gs_flatpak_app_set_extra_data (app, g_key_file_has_group (kf, "Extra Data"));
```

plugins/flatpak/gs-flatpak.c
```diff
@@ -3418,8 +3423,41 @@ gs_flatpak_app_get_copyable (GsFlatpak *self,
...
+		GsApp *runtime;
+		g_autoptr(FlatpakRemote) app_remote = NULL;
+		g_autoptr(FlatpakRemote) runtime_remote = NULL;
+		g_autofree char *app_collection_id = NULL;
+		g_autofree char *runtime_collection_id = NULL;
+		gboolean is_copyable = TRUE;
+
+		/* Check if the app's remote has a collection ID set */
+		app_remote = flatpak_installation_get_remote_by_name (self->installation,
+								   gs_app_get_origin (app),
```
treitter: Nitpick: the upstream style is to align with the first character after the `(`

plugins/flatpak/gs-flatpak.c
```diff
@@ -3418,8 +3423,41 @@ gs_flatpak_app_get_copyable (GsFlatpak *self,
...
+		if (app_remote != NULL) {
+			app_collection_id = flatpak_remote_get_collection_id (app_remote);
+			if (app_collection_id == NULL)
+				is_copyable = FALSE;
+		}
+
+		/* Check if the runtime's remote has a collection ID set */
+		runtime = gs_app_get_runtime (app);
+		runtime_remote = flatpak_installation_get_remote_by_name (self->installation,
+								   gs_app_get_origin (runtime),
```
treitter: Nitpick: the upstream style is to align with the first character after the `(`

plugins/flatpak/gs-flatpak.c
```diff
@@ -3418,8 +3423,41 @@ gs_flatpak_app_get_copyable (GsFlatpak *self,
...
+	if (copyable) {
+		GsApp *runtime;
+		g_autoptr(FlatpakRemote) app_remote = NULL;
+		g_autoptr(FlatpakRemote) runtime_remote = NULL;
+		g_autofree char *app_collection_id = NULL;
+		g_autofree char *runtime_collection_id = NULL;
+		gboolean is_copyable = TRUE;
+
+		/* Check if the app's remote has a collection ID set */
+		app_remote = flatpak_installation_get_remote_by_name (self->installation,
```
treitter: As discussed in Slack, I'd prefer to invert the logic such that `is_copyable = (app_remote != NULL && app_collection_id != NULL && ...)`, to play it safe. That is, a Flatpak needs to be in a reasonable state/have reasonable metadata to support it being copied. This reduces exposure to issues like the crasher this patch solves and we can test and enable more-permissive branches if and when we find they're necessary and safe.

mwleeds:
I pushed fixups for the nitpicks and with the inverted logic. It seems to still work this way. Apps installed from bundles aren't copyable to USB regardless of whether you use GNOME Software or the command line.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3418,8 +3420,39 @@ gs_flatpak_app_get_copyable (GsFlatpak *self,
 			     GCancellable *cancellable,
 			     GError **error)
 {
-	if (copyable)
-		*copyable = (gs_flatpak_get_collection_id (self, app) != NULL);
+	if (copyable) {
```
ramcq: Can you just early return here if !copyable and save a whole block of indent? :)

plugins/flatpak/gs-flatpak.c
```diff
@@ -3418,8 +3420,39 @@ gs_flatpak_app_get_copyable (GsFlatpak *self,
...
+		/* Check if it's an extra-data app (which is not supported for P2P distribution) */
+		is_extra_data = gs_flatpak_app_get_extra_data (app);
+
+		if (app_collection_id != NULL && runtime_collection_id != NULL && !is_extra_data)
+			*copyable = TRUE;
+		else
+			*copyable = FALSE;
+	}
 
 	return TRUE;
```
ramcq: Is it right to return TRUE if copyable was not set? Seems like a programmer error - isn't assert more valid for this?

PR #399, opened 2018-08-27
T23674

uajain:
https://phabricator.endlessm.com/T23674

### DIFF
--- a/lib/gs-utils.c
+++ b/lib/gs-utils.c
@@ -1059,4 +1059,13 @@ gs_utils_update_install_timestamp (GSettings *settings)
 	g_settings_set (settings, "install-timestamp", "x", g_date_time_to_unix (now));
 }
 
+gboolean
+gs_utils_app_is_auto_updating (GsApp *app)
+{
+       GVariant *tmp = gs_app_get_metadata_variant (app, "GnomeSoftware::auto-updating");
+       if (tmp == NULL)
+               return FALSE;
+       return g_variant_get_boolean (tmp);
+}
+
 /* vim: set noexpandtab: */
--- a/plugins/flatpak/gs-flatpak.c
+++ b/plugins/flatpak/gs-flatpak.c
@@ -56,6 +56,7 @@ struct _GsFlatpak {
 G_DEFINE_TYPE (GsFlatpak, gs_flatpak, G_TYPE_OBJECT)
 
 #define APP_METADATA_MOGWAI_UPDATE_PRIORITY "Mogwai::update-priority"
+#define EOS_MIN_FREE_SPACE_NEEDED_BYTES (500 * 1024 * 1024)
 
 /* higher has the value that is used by eos-updater for OS updates */
 typedef enum {
@@ -3678,6 +3679,77 @@ app_has_local_source (GsApp *app)
 		(url != NULL && g_str_has_prefix (url, "file://"));
 }
 
+static gboolean
+get_installation_dir_free_space (GsFlatpak *self, guint64 *free_space, GError **error)
+{
+	g_autoptr (GFile) installation_dir = NULL;
+	g_autoptr (GFileInfo) info = NULL;
+
+	installation_dir = flatpak_installation_get_path (self->installation);
+
+	info = g_file_query_filesystem_info (installation_dir,
+					     G_FILE_ATTRIBUTE_FILESYSTEM_FREE,
+					     NULL,
+					     error);
+	if (info == NULL)
+		return FALSE;
+
+	*free_space = g_file_info_get_attribute_uint64 (info, G_FILE_ATTRIBUTE_FILESYSTEM_FREE);
+	return TRUE;
+}
+
+static gboolean
+gs_flatpak_has_space_to_install (GsFlatpak *self, GsApp *app, GsAppList *list)
+{
+	g_autoptr(GError) error = NULL;
+	guint64 free_space = 0;
+	guint64 space_required = 0;
+
+	space_required = gs_app_get_size_download (app);
+	if (space_required == GS_APP_SIZE_UNKNOWABLE) {
+		g_warning ("Failed to query download size: %s", gs_app_get_unique_id (app));
+		space_required = 0;
+	}
+	space_required = space_required + EOS_MIN_FREE_SPACE_NEEDED_BYTES;
+
+	if (!get_installation_dir_free_space (self, &free_space, &error)) {
+		g_warning ("Error getting the free space available for installing %s: %s",
+			   gs_app_get_unique_id (app), error->message);
+		/* Even if we fail to get free space, we don't want to block this user-intiated
+		 * install action. It might happen that there is enough space to install but
+		 * an error happened during querying the filesystem info. */
+		return TRUE;
+	}
+
+	return free_space >= space_required;
+}
+
+static gboolean
+gs_flatpak_has_space_to_update (GsFlatpak *self, GsApp *app, GsAppList *list, gboolean is_auto_update)
+{
+	g_autoptr(GError) error = NULL;
+	guint64 free_space = 0;
+	guint64 space_required = 0;
+
+	if (is_auto_update && gs_app_get_kind (app) != AS_APP_KIND_RUNTIME) {
+		for (guint i = 0; i < gs_app_list_length (list); i++) {
+			GsApp *app_temp = gs_app_list_index (list, i);
+			space_required += gs_app_get_size_installed (app_temp);
+		}
+	}
+
+	space_required = (space_required * 2) + EOS_MIN_FREE_SPACE_NEEDED_BYTES;
+	if (!get_installation_dir_free_space (self, &free_space, &error)) {
+		g_warning ("Error getting the free space available for updating %s: %s",
+			   gs_app_get_unique_id (app), error->message);
+		/* Only fail autoupdates if we cannot query the filesystem info.
+		 * Manual updates follow the pattern similar to install's case. */
+		return is_auto_update ? FALSE : TRUE;
+	}
+
+	return free_space >= space_required;
+}
+
 gboolean
 gs_flatpak_app_install (GsFlatpak *self,
 			GsApp *app,
@@ -3876,6 +3948,16 @@ gs_flatpak_app_install (GsFlatpak *self,
 			return FALSE;
 		}
 
+		if (!gs_flatpak_has_space_to_install (self, app, list)) {
+			g_debug ("Skipping installation for %s: not enough space on disk",
+				 gs_app_get_unique_id (app));
+			gs_app_set_state_recover (app);
+			g_set_error (error, GS_PLUGIN_ERROR, GS_PLUGIN_ERROR_NO_SPACE,
+				     "You don't have enough space to install %s. Please remove apps or documents to create more space.",
+				     gs_app_get_unique_id (app));
+			return FALSE;
+		}
+
 		/* install all the required packages */
 		phelper = gs_flatpak_progress_helper_new (self->plugin, app);
 		phelper->job_max = gs_app_list_length (list);
@@ -3939,6 +4021,7 @@ gs_flatpak_update_app (GsFlatpak *self,
 	g_autoptr(GsFlatpakProgressHelper) phelper = NULL;
 	gboolean all_updates_done = TRUE;
 	gboolean is_proxy_app = FALSE;
+	gboolean is_auto_update = FALSE;
 	GsApp *runtime = NULL;
 	GsApp *update_runtime = NULL;
 
@@ -3982,6 +4065,23 @@ gs_flatpak_update_app (GsFlatpak *self,
 			gs_app_set_state_recover (app);
 			return FALSE;
 		}
+	}
+
+	is_auto_update = gs_utils_app_is_auto_updating (app);
+	if (!gs_flatpak_has_space_to_update (self, app, list, is_auto_update)) {
+		g_debug ("Skipping %s for %s: not enough space on disk",
+			 (is_auto_update ? "automatic update" : "update"),
+			 gs_app_get_unique_id (app));
+		gs_app_set_state_recover (app);
+		if (is_auto_update) {
+			/* Do not return an error because we don't want to see a banner
+			 * showing up out of the blue for possibly multiple update failures */
+			return TRUE;
+		}
+		g_set_error (error, GS_PLUGIN_ERROR, GS_PLUGIN_ERROR_NO_SPACE,
+			     "You don't have enough space to install %s. Please remove apps or documents to create more space.",
+			     gs_app_get_unique_id (app));
+		return FALSE;
 	}
 
 	/* update all the required packages */
--- a/lib/gs-utils.h
+++ b/lib/gs-utils.h
@@ -92,6 +92,7 @@ void		 gs_utils_append_key_value	(GString	*str,
 						 const gchar	*value);
 guint		 gs_utils_get_memory_total	(void);
 void		 gs_utils_update_install_timestamp (GSettings *settings);
+gboolean	 gs_utils_app_is_auto_updating (GsApp *app);
 
 G_END_DECLS
 
--- a/src/gs-update-monitor.c
+++ b/src/gs-update-monitor.c
@@ -30,6 +30,7 @@
 
 #include "gs-update-monitor.h"
 #include "gs-common.h"
+#include "gs-utils.h"
 
 #define APP_METADATA_AUTO_UPDATING "GnomeSoftware::auto-updating"
 
@@ -243,15 +244,6 @@ app_set_auto_updating (GsApp *app,
 	}
 }
 
-static gboolean
-app_is_auto_updating (GsApp *app)
-{
-	GVariant *tmp = gs_app_get_metadata_variant (app, APP_METADATA_AUTO_UPDATING);
-	if (tmp == NULL)
-		return FALSE;
-	return g_variant_get_boolean (tmp);
-}
-
 static void
 app_update_finished_cb (GObject *source,
 			GAsyncResult *res,
@@ -333,7 +325,7 @@ download_now_cb (GObject *obj,
 	if (state == AS_APP_STATE_INSTALLING) {
 		/* if we cannot update at the moment, cancel any automatically
 		 * started update */
-		if (!app_is_auto_updating (helper->app))
+		if (!gs_utils_app_is_auto_updating (helper->app))
 			return;
 
 		g_debug ("Cancelling scheduled update of app %s, as "


lib/gs-utils.c
```diff
@@ -1059,4 +1059,13 @@ gs_utils_update_install_timestamp (GSettings *settings)
 	g_settings_set (settings, "install-timestamp", "x", g_date_time_to_unix (now));
 }
 
+gboolean
+gs_utils_app_is_auto_updating (GsApp *app)
```
joaquimrocha: In the commit message, make sure you write the reason for this move (that it will be used later in different parts of the code).

plugins/flatpak/gs-flatpak.c
```diff
@@ -3666,6 +3667,51 @@ app_has_local_source (GsApp *app)
...
 
+static gboolean
+gs_flatpak_allow_install_or_update (GsFlatpak *self,GsApp *app, GsAppList *list, gboolean is_update)
+{
+	g_autoptr (GFile) installation_dir = NULL;
+	g_autofree gchar *path = NULL;
+	GsApp *runtime, *update_runtime;
+	guint64 runtime_delta_size = 0;
+	guint64 min_free_space = 0;
+	guint i;
```
joaquimrocha: Add this in the `for` loop.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3666,6 +3667,51 @@ app_has_local_source (GsApp *app)
 		(url != NULL && g_str_has_prefix (url, "file://"));
 }
 
+static gboolean
+gs_flatpak_allow_install_or_update (GsFlatpak *self,GsApp *app, GsAppList *list, gboolean is_update)
```
joaquimrocha: Space needed before `GsApp`.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3666,6 +3667,51 @@ app_has_local_source (GsApp *app)
...
+	guint64 runtime_delta_size = 0;
+	guint64 min_free_space = 0;
+	guint i;
+	struct statvfs buf;
+
+	installation_dir = flatpak_installation_get_path (self->installation);
+	path = g_file_get_path (installation_dir);
+
+	if (statvfs (path, &buf) == -1) {
+		g_debug ("statvfs: Couldn't query disk space information");
```
joaquimrocha: This should be a warning (also say when/where it was).

plugins/flatpak/gs-flatpak.c
```diff
@@ -3666,6 +3667,51 @@ app_has_local_source (GsApp *app)
...
+		return FALSE;
+	}
+
+	runtime = gs_app_get_runtime (app);
+	update_runtime = gs_app_get_update_runtime (app);
+	if (is_update && runtime != update_runtime) {
+
```
joaquimrocha: No need for this extra newline.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3666,6 +3667,51 @@ app_has_local_source (GsApp *app)
...
+		/* Correctly setup the runtime sizes if they are updatable */
+		guint64 runtime_size = gs_app_get_size_installed (runtime);
+		guint64 update_runtime_size = gs_app_get_size_installed (update_runtime);
+
+		if ((runtime_delta_size = update_runtime_size - runtime_size) > update_runtime_size)
```
joaquimrocha: What if the `runtime_size` is actually greater than the update's?

joaquimrocha: Also, `update_runtime_size - runtime_size > update_runtime_size` means `runtime_size < 0`, which shouldn't really happen, meaning this if's body will never be used.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3666,6 +3667,51 @@ app_has_local_source (GsApp *app)
...
+			runtime_delta_size = runtime_size - update_runtime_size;
+
+		gs_app_list_remove (list, update_runtime);
+	}
+
+	min_free_space = runtime_delta_size;
+	for (i = 0; i < gs_app_list_length (list); i++) {
+		GsApp *app_temp = gs_app_list_index (list, i);
+		min_free_space += gs_app_get_size_installed (app_temp);
```
joaquimrocha: You should make sure that the `gs_app_get_size_installed` for every `app_temp` has a value (I don't know if at this point they already have it or need to be refined for it to be set up).

plugins/flatpak/gs-flatpak.c
```diff
@@ -3666,6 +3667,51 @@ app_has_local_source (GsApp *app)
...
+	}
+
+	if (buf.f_bfree < (min_free_space / buf.f_bsize + 1))
```
joaquimrocha: Why `+ 1` here? If it's to avoid that `f_bsize` may be zero (unlikely, right?), then you need to add parenthesis to it, as division takes precedence, and write a comment about why the `+ 1` is used.
However, I don't think `f_bsize` should be zero (but you should check it out).
Independently, I think maybe multiplying `f_bfree * f_bsize` makes it clearer.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3864,6 +3910,13 @@ gs_flatpak_app_install (GsFlatpak *self,
 			return FALSE;
 		}
 
+		/* Check disk space requirements for the list */
+		if (!gs_flatpak_allow_install_or_update (self, app, list, FALSE)) {
+			g_debug ("Skipping installation for: %s as prone to failure due to disk-space",
+				  gs_app_get_id (app));
```
joaquimrocha: Always log using the app's unique id (otherwise we may not know which app it is, if more than one have the same ID).

plugins/flatpak/gs-flatpak.c
```diff
@@ -3864,6 +3910,13 @@ gs_flatpak_app_install (GsFlatpak *self,
 			return FALSE;
 		}
 
+		/* Check disk space requirements for the list */
+		if (!gs_flatpak_allow_install_or_update (self, app, list, FALSE)) {
```
joaquimrocha: You're not checking if it's an auto update here.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3972,6 +4025,14 @@ gs_flatpak_update_app (GsFlatpak *self,
 		}
 	}
 
+	/* Check disk space requirements for the list */
```
joaquimrocha: Instead of having this comment, I think it'd be better to rename the `...allow_install_or_update` to `gs_flatpak_has_space_to_install_or_update`. Self-documented, so no need for that comment.

plugins/flatpak/gs-flatpak.c
uajain: `gs_app_get_size_installed` returns `guint64`. This check basically checks for wrapping around. If wrapping around is detected (means `runtime_size` > `update_runtime_size`), then reverse the parameters (the `if`'s body)

plugins/flatpak/gs-flatpak.c
uajain: This is the `gs_flatpak_app_install`. I check for autoupdating in `gs_flatpak_update_app`. It's what it is supposed to do right?

plugins/flatpak/gs-flatpak.c
uajain: > Why + 1 here? If it's to avoid that f_bsize may be zero (unlikely, right?), then you need to add parenthesis to it, as division takes precedence, and write a comment about why the + 1 is used.
> However, I don't think f_bsize should be zero (but you should check it out).

Yes, the `f_bsize`=0 is highliy unlikely and I wasn't trying to add `1` to `f_bsize` either.
Earlier I was using `<=` so I added `+1` to make sure that we compare with one extra block. But I remember that I shifted from `<=` to `<` but I think forgot to remove `+1`. I'll remove it now.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3666,6 +3667,52 @@ app_has_local_source (GsApp *app)
...
+		guint64 update_runtime_size = gs_app_get_size_installed (update_runtime);
+
+		if ((runtime_delta_size = update_runtime_size - runtime_size) > update_runtime_size)
+			runtime_delta_size = runtime_size - update_runtime_size;
+
+		gs_app_list_remove (list, update_runtime);
+	}
+
+	min_free_space = runtime_delta_size;
+	min_free_space +=  524288000; /* min-free-space-size = 500MB for Endless computers */
```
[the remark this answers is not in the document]
uajain: Yes, that's a hotpatch. I didn't find anything trivial to expose the min-free-space-* bytes needed to reserve other than reading from config and parsing it on our own.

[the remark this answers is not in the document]
pwithnall: I don’t think there’s suitable API in libflatpak for it, no. `flatpak_installation_get_config()` unfortunately only works for keys prefixed with `xa.`. You’d have to use the libostree API directly or (preferred) add new API to expose it in libflatpak. That could be a follow-up PR though.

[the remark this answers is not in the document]
uajain: Yes, that's the correct way of doing it and it was already in the back of my mind.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3972,6 +4033,13 @@ gs_flatpak_update_app (GsFlatpak *self,
 		}
 	}
 
+	if (gs_utils_app_is_auto_updating (app) &&
+	    !gs_flatpak_has_space_to_install_or_update (self, app, list, TRUE)) {
+		g_debug ("Skipping automatic update for: %s as prone to failure due to disk-space",
```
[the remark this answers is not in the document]
uajain: I am not much convinced with "no space left on disk" error message.
We are logging the message(which is not user-visible) of a situation that we are trying to avoid.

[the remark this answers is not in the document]
pwithnall: If I understand correctly, this message is printed when the amount of free space left on the disk is less than the minimum we estimated for the given app, so it’s not really a case of there being `no space left on disk`. However, I think `prone to failure due to disk-space` could be clearer.

How about `Skipping automatic update for ‘%s’, as it will likely run out of disk space`?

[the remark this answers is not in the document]
joaquimrocha: Okay. I just wanted to make it consistent with other error messages, but you're right it's not like we have 0 bytes of space left.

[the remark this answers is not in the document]
joaquimrocha: BTW, I'd still prefer something more compact like `Skipping automatic update of '%s': not enough space on disk`. Where "enough" means we decided it wasn't sufficient for a safe update...
In any case, I am also fine with Philip's proposal even if it seems more verbose to me.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3864,6 +3916,15 @@ gs_flatpak_app_install (GsFlatpak *self,
 			return FALSE;
 		}
 
+		if (!gs_flatpak_has_space_to_install_or_update (self, app, list, FALSE)) {
+			g_debug ("Skipping installation for: %s as prone to failure due to disk-space",
```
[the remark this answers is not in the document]
uajain: As I understand, you do this check when:

1) Installing an app (from details page)
2) auto-updates 

You don't do this check when:
1) User initiated Update.

@nedrichards Do you agree with this?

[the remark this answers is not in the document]
nedrichards: Yeah, we don't block user actions unless we absolutely know they will fail.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3666,6 +3666,58 @@ app_has_local_source (GsApp *app)
...
+	update_runtime = gs_app_get_update_runtime (app);
+	if (is_update && runtime != update_runtime) {
+		/* Correctly setup the runtime sizes if they are updatable */
+		guint64 runtime_size = gs_app_get_size_installed (runtime);
+		guint64 update_runtime_size = gs_app_get_size_installed (update_runtime);
+
+		if (update_runtime_size >= runtime_size)
+			runtime_delta_size  = update_runtime_size - runtime_size;
+		else
+			runtime_delta_size = runtime_size - update_runtime_size;
```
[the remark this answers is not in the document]
uajain: As discussed in the daily, we might need to ask upstream what's the implication of the static delta download when new runtime size is smaller than old runtime size.

plugins/flatpak/gs-flatpak.c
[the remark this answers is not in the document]
joaquimrocha: But "Installing an app (from details page)" is a user action. So according to what Nick's said, we shouldn't do it.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3666,6 +3666,58 @@ app_has_local_source (GsApp *app)
...
+		gs_app_list_remove (list, update_runtime);
+	}
+
+	space_required = runtime_delta_size;
+	for (guint i = 0; i < gs_app_list_length (list); i++) {
+		GsApp *app_temp = gs_app_list_index (list, i);
+		space_required += gs_app_get_size_installed (app_temp);
+	}
+
+	space_required = (space_required * 2) + (500 * 1024 * 1024); /* min-free-space-size = 500MB for Endless */
```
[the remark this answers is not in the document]
ramcq: Hmm - can't we fetch this min-free-space-size value from ostree?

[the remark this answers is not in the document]
pwithnall: See [this discussion](https://github.com/endlessm/gnome-software/pull/399#discussion_r213380938). tl;dr: Deferred pending new API.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3864,6 +3916,15 @@ gs_flatpak_app_install (GsFlatpak *self,
 			return FALSE;
 		}
 
+		if (!gs_flatpak_has_space_to_install_or_update (self, app, list, FALSE)) {
+			g_debug ("Skipping installation for: %s as prone to failure due to disk-space",
+				 gs_app_get_unique_id (app));
+			g_set_error (error, GS_PLUGIN_ERROR, GS_PLUGIN_ERROR_FAILED,
```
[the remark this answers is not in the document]
uajain: It shows up in the banner as per you asked to check, but the "Examine Disk" button leads to nowhere (T21533).
screenshot from 2018-09-06 16-56-53

plugins/flatpak/gs-flatpak.c
```diff
@@ -3678,6 +3678,75 @@ app_has_local_source (GsApp *app)
...
+	guint64 free_space = 0;
+
+	installation_dir = flatpak_installation_get_path (self->installation);
+
+	info = g_file_query_filesystem_info (installation_dir,
+					     G_FILE_ATTRIBUTE_FILESYSTEM_FREE,
+					     NULL,
+					     &local_error);
+	if (info == NULL)
+		g_warning ("Failed to query filesystem info: %s", local_error->message);
```
joaquimrocha: You are not returning here, so it will proceed with a NULL `info` and (likely) crash.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3678,6 +3678,75 @@ app_has_local_source (GsApp *app)
...
+gs_flatpak_has_space_to_install (GsFlatpak *self, GsApp *app, GsAppList *list)
+{
+	guint64 free_space = 0;
+	guint64 space_required = 0;
+
+	space_required = gs_app_get_size_download (app);
+	if (space_required == GS_APP_SIZE_UNKNOWABLE)
+		g_warning ("Failed to query download size: %s", gs_app_get_unique_id (app));
+
+	space_required = space_required + (500 * 1024 * 1024); /* min-free-space-size = 500MB for Endless */
```
joaquimrocha: Please declare a variable or a define with this value. So it's easier to spot, and by giving it a proper name, the code will be easier to read too. Something like
`#define EOS_MIN_FREE_SPACE_NEEDED_BYTES (500 * 1024 * 1024)`

plugins/flatpak/gs-flatpak.c
```diff
@@ -3678,6 +3678,75 @@ app_has_local_source (GsApp *app)
...
+	free_space = get_installation_dir_free_space (self);
+
+	if (free_space < space_required)
+		return FALSE;
+
+	return TRUE;
```
joaquimrocha: Nitpick: You could also `return free_space >= space_required;`

plugins/flatpak/gs-flatpak.c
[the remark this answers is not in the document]
joaquimrocha: You haven't addressed this, and now you're repeating the value in the other function too.

plugins/flatpak/gs-flatpak.c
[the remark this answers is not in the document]
joaquimrocha: I think there was a task to address this button (either downstream or upstream, I cannot remember). Please check that out, and if this is not true, please open a new task.
So we should still use the `NO_SPACE_ERROR` and address the button problem in a different task.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3678,6 +3678,75 @@ app_has_local_source (GsApp *app)
...
+	guint64 free_space = 0;
+	guint64 space_required = 0;
+
+	/* If an app has an updatable runtime, the runtime is not updated
+	 * as a part of app update.
+	 *
+	 * Having said that, runtime updates and their related refs (see T22716)
+	 * are done together and below heuristics might not be good for if GsApp
+	 * app is a runtime.
+	 */
```
joaquimrocha: What does this mean? It doesn't matter that the runtime is not updated together with the app for this task.
But you should check this code for runtime updates too. I think it will work fine, but in any case this comment is confusing and shouldn't be necessary (please do test both apps + runtimes).

plugins/flatpak/gs-flatpak.c
```diff
@@ -3984,6 +4062,12 @@ gs_flatpak_update_app (GsFlatpak *self,
 		}
 	}
 
+	if (!gs_flatpak_has_space_to_update (self, app, list)) {
+		g_debug ("Skipping automatic update for %s: not enough space on disk",
```
joaquimrocha: The function is also used for manual updates (it will check the minimum space only). So this message is misleading. Maybe you should pass the `is_auto_update` as a parameter to the function (making it more clear), and use it here to filter two appropriate messages.

plugins/flatpak/gs-flatpak.c
uajain: This is something I need some discussion on. The heuristics is bad if GsApp is a Runtime. For e.g. GNOME runtime is around ~1.6GB when installed, so if there is a runtime update pending; according to this heuristics it will dictate to have `(1.6GB * 2 + 500MB)` space required for update to happen.

ramcq: Yeah, the assumption I detailed in the ticket was that we should only use this heuristic for app upgrades, on the basis that most files are likely to be changed. For runtimes, that's (hopefully!) far less likely. If we could get the delta size, it would be best, but for the moment we should just leave it as checking the min space.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3719,32 +3723,22 @@ gs_flatpak_has_space_to_install (GsFlatpak *self, GsApp *app, GsAppList *list)
...
 
-	/* If an app has an updatable runtime, the runtime is not updated
-	 * as a part of app update.
-	 *
-	 * Having said that, runtime updates and their related refs (see T22716)
-	 * are done together and below heuristics might not be good for if GsApp
-	 * app is a runtime.
-	 */
-	if (gs_utils_app_is_auto_updating (app)) {
+	if (is_auto_update && gs_app_get_kind (app) != AS_APP_KIND_RUNTIME) {
```
joaquimrocha: Notice that if the app needs a new runtime, the runtime's size will be included in the required space which is multipled. So if the filter for the runtime is to never require e.g. 1.6 GB * 2 + 500 MB, we need to do some extra checks here.

plugins/flatpak/gs-flatpak.c
```diff
@@ -4062,9 +4057,16 @@ gs_flatpak_update_app (GsFlatpak *self,
...
-		g_debug ("Skipping automatic update for %s: not enough space on disk",
-			 gs_app_get_unique_id (app));
+	is_auto_update = gs_utils_app_is_auto_updating (app);
+	if (!gs_flatpak_has_space_to_update (self, app, list, is_auto_update)) {
+		if (is_auto_update)
+			g_debug ("Skipping automatic update for %s: not enough space on disk",
+				 gs_app_get_unique_id (app));
+		else
+			g_debug ("Skipping update for %s: not enough space on disk",
+				 gs_app_get_unique_id (app));
```
joaquimrocha: How about:
`g_debug ("Skipping %s for %s: not enough space on disk", (is_auto_update ? " automatic update" : "update"),
		 gs_app_get_unique_id (app));`

plugins/flatpak/gs-flatpak.c
```diff
@@ -4062,9 +4057,16 @@ gs_flatpak_update_app (GsFlatpak *self,
...
+
+		gs_app_set_state_recover (app);
```
joaquimrocha: You shouldn't return `FALSE` without setting the `error` output parameter.

plugins/flatpak/gs-flatpak.c
uajain: I don't understand. If it's a **new** runtime and  we need to make it fit in that would require double disk space, no? (as it's a `--system` pull)

plugins/flatpak/gs-flatpak.c
uajain: Then it would show up in the error banner, no? I think we want to avoid that case basically.

plugins/flatpak/gs-flatpak.c
joaquimrocha: Ah okay. I think you're right then.

plugins/flatpak/gs-flatpak.c
joaquimrocha: Ah, because this is the skip (sorry I was in a rush). In this case then you have to return `TRUE`, as this is the value plugin methods return also when they ignore a request.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3692,10 +3693,13 @@ get_installation_dir_free_space (GsFlatpak *self)
 					     G_FILE_ATTRIBUTE_FILESYSTEM_FREE,
 					     NULL,
 					     &local_error);
-	if (info == NULL)
+	if (info == NULL) {
 		g_warning ("Failed to query filesystem info: %s", local_error->message);
+		goto out;
```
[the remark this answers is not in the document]
uajain: I am not very sure of this one. APIs  like `g_file_query_filesystem_info` seems to very stable, no? But yeah I get the notion of things going wrong. But at the same time, I am concerned not to over-engineer stuff. 

Although here one can argue that, there is an error getting the **basic** filesystem info, so does it make sense to take a step further? (trying to install the app). The reason being there that something inherently is going wrong with the filesystem. Not sure.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3679,72 +3680,64 @@ app_has_local_source (GsApp *app)
...
 
 	info = g_file_query_filesystem_info (installation_dir,
 					     G_FILE_ATTRIBUTE_FILESYSTEM_FREE,
 					     NULL,
-					     &local_error);
-	if (info == NULL)
-		g_warning ("Failed to query filesystem info: %s", local_error->message);
+					     error);
+	if (info == NULL) {
+		g_warning ("Failed to query filesystem info");
```
joaquimrocha: Since we now have an error, instead of using just as a sentinel, we should print also when happened. So the warning should include `: %s", error->message` at the end. And it's better in this case to move the warning to the caller, so you can use a specific message ("error updating", "error installing", etc.) which also prevents us from having to deal with whether the error is passed as `NULL` at some point in the future.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3679,72 +3680,64 @@ app_has_local_source (GsApp *app)
...
 		g_warning ("Failed to query download size: %s", gs_app_get_unique_id (app));
 
-	space_required = space_required + (500 * 1024 * 1024); /* min-free-space-size = 500MB for Endless */
-	free_space = get_installation_dir_free_space (self);
-
-	if (free_space < space_required)
-		return FALSE;
+	space_required = space_required + EOS_MIN_FREE_SPACE_NEEDED_BYTES;
+	free_space = get_installation_dir_free_space (self, &error);
+	if (error != NULL)
```
joaquimrocha: Following my previous comment, please print the warning here instead.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3679,72 +3680,64 @@ app_has_local_source (GsApp *app)
...
 
-	space_required = (space_required * 2) + (500 * 1024 * 1024); /* min-free-space-size = 500MB for Endless */
-	free_space = get_installation_dir_free_space (self);
-
-	if (free_space < space_required)
-		return FALSE;
+	space_required = (space_required * 2) + EOS_MIN_FREE_SPACE_NEEDED_BYTES;
+	free_space = get_installation_dir_free_space (self, &error);
+	if (error != NULL)
+		return is_auto_update ? FALSE : TRUE;
```
joaquimrocha: Add a comment here explaining why we differ when an error occurs (otherwise in 2 weeks we'll have to waste brain power just understanding why we did it).

plugins/flatpak/gs-flatpak.c
```diff
@@ -3679,72 +3680,73 @@ app_has_local_source (GsApp *app)
...
 
-	space_required = space_required + (500 * 1024 * 1024); /* min-free-space-size = 500MB for Endless */
-	free_space = get_installation_dir_free_space (self);
-
-	if (free_space < space_required)
-		return FALSE;
+	space_required = space_required + EOS_MIN_FREE_SPACE_NEEDED_BYTES;
+	free_space = get_installation_dir_free_space (self, &error);
+	if (error != NULL) {
+		g_warning ("Error installing: %s", error->message);
```
joaquimrocha: It was not an error installing, it was an error getting the free space. And you should always mention which app it happened for, otherwise the message will not be very useful.
`Error getting the free space available for installing %s: %s` -> app unique id , error msg

plugins/flatpak/gs-flatpak.c
```diff
@@ -3679,72 +3680,73 @@ app_has_local_source (GsApp *app)
...
 
-	space_required = (space_required * 2) + (500 * 1024 * 1024); /* min-free-space-size = 500MB for Endless */
-	free_space = get_installation_dir_free_space (self);
-
-	if (free_space < space_required)
-		return FALSE;
+	space_required = (space_required * 2) + EOS_MIN_FREE_SPACE_NEEDED_BYTES;
+	free_space = get_installation_dir_free_space (self, &error);
+	if (error != NULL) {
+		g_warning ("Error updating: %s", error->message);
```
joaquimrocha: See my comment about the installation error.

plugins/flatpak/gs-flatpak.c
```diff
@@ -4062,9 +4066,17 @@ gs_flatpak_update_app (GsFlatpak *self,
...
 
-	if (!gs_flatpak_has_space_to_update (self, app, list)) {
-		g_debug ("Skipping automatic update for %s: not enough space on disk",
+	is_auto_update = gs_utils_app_is_auto_updating (app);
+	if (!gs_flatpak_has_space_to_update (self, app, list, is_auto_update)) {
+		g_debug ("Skipping %s for %s: not enough space on disk",
+			 (is_auto_update ? "automatic update" : "update"),
 			 gs_app_get_unique_id (app));
+		gs_app_set_state_recover (app);
+		if (!is_auto_update)
```
joaquimrocha: Add a comment about why we're not reporting an error.
`/* Do not return an error because we don't want to see a banner showing up out of the blue for possibly multiple update failures */`

plugins/flatpak/gs-flatpak.c
```diff
@@ -3678,6 +3679,76 @@ app_has_local_source (GsApp *app)
...
+static gboolean
+gs_flatpak_has_space_to_install (GsFlatpak *self, GsApp *app, GsAppList *list)
+{
+	g_autoptr(GError) error = NULL;
+	guint64 free_space = 0;
+	guint64 space_required = 0;
+
+	space_required = gs_app_get_size_download (app);
+	if (space_required == GS_APP_SIZE_UNKNOWABLE)
+		g_warning ("Failed to query download size: %s", gs_app_get_unique_id (app));
```
ramcq: GS_APP_SIZE_UNKNOWABLE is G_MAXUINT64 - it seems a bit weird to go ahead and add EOS_MIN_FREE_SPACE_NEEDED_BYTES to this - it will probably give us EOS_MIN_FREE_SPACE_NEEDED_BYTES - 1 but also causes an overflow condition. Would be better to explicitly set space_required to 0 if we can't fetch the value, wouldn't it?

plugins/flatpak/gs-flatpak.c
```diff
@@ -3678,6 +3679,76 @@ app_has_local_source (GsApp *app)
...
+	g_autoptr (GFileInfo) info = NULL;
+
+	installation_dir = flatpak_installation_get_path (self->installation);
+
+	info = g_file_query_filesystem_info (installation_dir,
+					     G_FILE_ATTRIBUTE_FILESYSTEM_FREE,
+					     NULL,
+					     error);
+	if (info == NULL)
+		return 0;
```
ramcq: It doesn't necessarily need a change, because in both call sites, we check the value of the GError * rather than the return value, but perhaps as some stylistic advice it's better to have an error return value which isn't a valid return value - either G_MAXUINT64 or make the function into a gboolean and have the space filled into a guint64 *. Because in typical glib (@pwithnall can correct me if I am wrong!), GError ** may be NULL, and this means that a caller then can't distinguish an error (whatever it is) from there being literally 0 free space - and/or if someone did if (return value == 0) to detect an error, *error would be unset if there was no error (rather than genuinely no space), potentially causing a SEGV in the calling code.

plugins/flatpak/gs-flatpak.c
joaquimrocha: Oh! Great catch! I had the impresion that UNKNOWABLE was actually 0...
So yes, the `space_required` needs to be reset in that case.

plugins/flatpak/gs-flatpak.c
pwithnall: > @pwithnall can correct me if I am wrong!

You are right. FWIW, I’d prefer returning the `guint64` as an out parameter, rather than as a return value with `g_assert (error != NULL)` because, while adding the `g_assert()` is correct, it’s an anti-pattern for public APIs. So I don’t really want people to ever think it’s acceptable.

uajain:
Pushed yet another fixup! 
Please do _not_ merge. Due to volatile internet connectivity today, I couldn't QA this PR properly.

ramcq:
Fixup LGTM, squash and we can merge once you're happy!

uajain:
PR in action / QA satisfactory. Merging now.

PR #400, opened 2018-08-27
flatpak: Assert on programmer error rather than allowing it

mwleeds:
There shouldn't be any reason to call gs_flatpak_app_get_copyable()
without passing in a pointer to @copyable, so assert if it's NULL. I've
tested this to make sure the "Copy to USB" feature still works with this
patch.

https://phabricator.endlessm.com/T23651

### DIFF
--- a/plugins/flatpak/gs-flatpak.c
+++ b/plugins/flatpak/gs-flatpak.c
@@ -3420,39 +3420,39 @@ gs_flatpak_app_get_copyable (GsFlatpak *self,
 			     GCancellable *cancellable,
 			     GError **error)
 {
-	if (copyable) {
-		GsApp *runtime;
-		g_autoptr(FlatpakRemote) app_remote = NULL;
-		g_autoptr(FlatpakRemote) runtime_remote = NULL;
-		g_autofree char *app_collection_id = NULL;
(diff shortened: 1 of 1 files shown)


plugins/flatpak/gs-flatpak.c
```diff
@@ -3420,39 +3420,39 @@ gs_flatpak_app_get_copyable (GsFlatpak *self,
...
-		is_extra_data = gs_flatpak_app_get_extra_data (app);
+	/* Check if it's an extra-data app (which is not supported for P2P distribution) */
+	is_extra_data = gs_flatpak_app_get_extra_data (app);
 
-		if (app_collection_id != NULL && runtime_collection_id != NULL && !is_extra_data)
-			*copyable = TRUE;
-		else
-			*copyable = FALSE;
-	}
+	if (app_collection_id != NULL && runtime_collection_id != NULL && !is_extra_data)
```
treitter: personally, I'd do:

```
*copyable = (!app_collection_id && ...);
```
(though I guess upstream would want explicit `!= NULL`); or:

```
*copyable = FALSE;
if (app_collection_id != NULL && ...);
```

to save a line or more but it's a fine point.

joaquimrocha:
LGTM. Merged.

PR #405, opened 2018-09-07
T23724: handle long-running (install, update, etc) ops at idle IO scheduler class

ramcq:
https://phabricator.endlessm.com/T23724


ramcq:
Going for some pre-review (and CI) here before I submit upstream.

joaquimrocha:
I see. Well, let's try it then.

PR #406, opened 2018-09-13
flatpak: Fix heap-use-after-free of a GError

mwleeds:
This patch fixes a heap-use-after-free of a GError and subsequent seg
fault in the case that you use the "Copy to USB" button and the `flatpak
create-usb` command fails, such as because the `ostree-metadata` ref
isn't available for one of the remotes relevant to the app being copied.
This happens for example if you reflash a computer and never go online
before using the "Copy to USB" feature.

https://phabricator.endlessm.com/T23754


uajain:
LGTM.

PR #407, opened 2018-09-20
T23422 usb os updates

treitter:
DO NOT MERGE

This is ready for early review but isn't ready to merge as I still need to fill in the "empty state" message for the USB category when the disk has no apps (there's a `TODO` comment) and need to add the "copy OS to USB" button which will call `eos-updater-prepare-volume` (which has another `TODO` comment). That last piece not yet being in place means that the critical feature doesn't yet work but that should be straightforward as soon as I come up with a decent way to manage a `GtkBox` at the top of `GsCategoryPage` that contains the button. The simplest way seems to be adding it to the `.ui` file (invisible by default) and show or hide it upon reload based on whether the category is USB

https://phabricator.endlessm.com/T23422


lib/gs-plugin-loader.c
```diff
@@ -2501,23 +2486,16 @@ mount_added_cb (GVolumeMonitor *volume_monitor,
...
-	g_autoptr(GMount) existing_mount = NULL;
 
 	if (!g_drive_is_removable (drive))
 		return;
 
-	existing_mount = find_mount_in_list (priv->removable_mounts, mount);
-
-	/* update the list to remove any equivalent mounts and move this to the
-	 * front so we prioritize mounts by recency
+	/* update the list to remove this mount if it already exists mounts and
```
joaquimrocha: `if it already exists mounts`
Maybe there's a comma missing or an extra word here.

lib/gs-plugin-loader.c
```diff
@@ -2526,12 +2504,12 @@ mount_removed_cb (GVolumeMonitor *volume_monitor,
...
 {
+	GList *list_orig;
 	GsPluginLoaderPrivate *priv = gs_plugin_loader_get_instance_private (plugin_loader);
-	GMount *existing_mount = find_mount_in_list (priv->removable_mounts,
-						     mount);
-	if (existing_mount) {
-		priv->removable_mounts = g_list_remove (priv->removable_mounts,
-							existing_mount);
+
+	list_orig = priv->removable_mounts;
```
joaquimrocha: Nitpick: You can declare + assign this by moving the declaration under the priv one.

lib/gs-plugin-loader.c
```diff
@@ -2526,12 +2504,12 @@ mount_removed_cb (GVolumeMonitor *volume_monitor,
...
+	priv->removable_mounts = g_list_remove (priv->removable_mounts, mount);
+	if (list_orig != priv->removable_mounts) {
```
joaquimrocha: Nitpick: Remove the curly brackets now that it's a one-liner if body.

src/gs-overview-page.c
```diff
@@ -486,8 +486,20 @@ get_categories_cb (GObject *source_object,
 	gs_shell_side_filter_clear_categories (priv->shell);
 
 	for (guint i = 0; i < list->len; i++) {
+		gboolean usb_category_inserted = FALSE;
```
joaquimrocha: Commenting here because there's no way of commenting in the commit message: Please capitalize the first word (display -> Display). This is not done when the first word is the plugin/module's name of course.

src/gs-overview-page.c
```diff
@@ -486,8 +486,20 @@ get_categories_cb (GObject *source_object,
...
 		cat = GS_CATEGORY (g_ptr_array_index (list, i));
-		if (gs_category_get_size (cat) == 0)
+
+		if (g_strcmp0 (gs_category_get_id (cat), "usb") == 0) {
+			g_autoptr(GList) copy_dests = gs_plugin_loader_dup_copy_dests (plugin_loader);
+
+			if (copy_dests != NULL && g_list_length (copy_dests) > 0)
```
joaquimrocha: Do we need to use the `g_list_length` here to test for whether the list has an element? i.e. can the list have no elements but not be `NULL`? From that function's doc:
`To check whether the list is non-empty, it is faster to check list against NULL.`

src/gs-overview-page.c
```diff
@@ -486,8 +486,20 @@ get_categories_cb (GObject *source_object,
...
+				usb_category_inserted = TRUE;
+		}
+
+		// allow empty categories for USB since there are usable actions
+		// (such as copy OS to USB) in the USB category even when it has
+		// no apps available
```
joaquimrocha: Use C-style comments.

src/gs-shell.c
```diff
@@ -599,8 +599,14 @@ on_overview_categories_loaded (GsOverviewPage *page, gpointer user_data)
 					      user_data);
 
 	if (category == NULL) {
-		g_warning ("No %s category after overview categories are refreshed!",
-			   data->category_name);
+		/* this is only an error condition if a non-NULL category name
+		 * is provided; otherwise, selecting the category after creation
+		 * was not intended */
+		if (data->category_name != NULL) {
```
joaquimrocha: This sounds like a different issue that should be addressed in a separate patch. Or is this problem introduced by your patch? The comment also doesn't make it clear for me.

src/gs-shell.c
```diff
@@ -599,8 +599,14 @@ on_overview_categories_loaded (GsOverviewPage *page, gpointer user_data)
...
+			g_warning ("No %s category after overview categories are refreshed!",
+				   data->category_name);
+		}
```
joaquimrocha: Nitpick: No need for brackets.

src/gs-shell.c
```diff
@@ -619,7 +626,6 @@ reload_overview_and_select_category (GsShell *shell, const gchar *category_name)
 	data->shell = shell;
 	data->category_name = g_strdup (category_name);
 	page = GS_PAGE (g_hash_table_lookup (priv->pages, "overview"));
-
```
joaquimrocha: Don't remove this newline.

src/gs-shell.c
```diff
@@ -782,6 +788,37 @@ gs_shell_reload_cb (GsPluginLoader *plugin_loader, GsShell *shell)
 	}
 }
 
+static void
+gs_shell_reload_and_select_removable_media_category (GsShell *shell,
+						     gboolean reload_overview,
+						     gboolean select_category,
+						     const gchar *category_name)
```
joaquimrocha: I guess this is well aligned but Github shows it badly? (Just making sure, because it's happened before)

src/gs-shell.c
```diff
@@ -782,6 +788,37 @@ gs_shell_reload_cb (GsPluginLoader *plugin_loader, GsShell *shell)
...
+{
+	/* using NULL for this below indicates it should not be selected */
```
joaquimrocha: This seems like a thing that may cause trouble down the road. Should we have a simple `reload_overview` that selects no category? Or is this functionality being used more times?

src/gs-shell.c
```diff
@@ -782,6 +788,37 @@ gs_shell_reload_cb (GsPluginLoader *plugin_loader, GsShell *shell)
...
+	const char *cat_name = (select_category ? category_name : NULL);
+
+	if (!reload_overview) {
+		reload_overview_and_select_category (shell, cat_name);
```
joaquimrocha: This reads as: if not reload_overview, then reload overview and select category...
Is this correct?

src/gs-shell.c
```diff
@@ -782,6 +788,37 @@ gs_shell_reload_cb (GsPluginLoader *plugin_loader, GsShell *shell)
...
+	} else {
+		GsCategory *category = side_filter_get_category_by_name (shell, cat_name);
+		if (category == NULL) {
+			reload_overview_and_select_category (shell, cat_name);
+		} else {
+			gs_shell_change_mode (shell, GS_SHELL_MODE_CATEGORY,
+					      (gpointer) category, TRUE);
+		}
```
joaquimrocha: Nitpick: No need for curly brackets.

src/gs-shell.c
```diff
@@ -782,6 +788,37 @@ gs_shell_reload_cb (GsPluginLoader *plugin_loader, GsShell *shell)
...
+	}
```
joaquimrocha: Maybe this can be rewritten as (notice though, I am keeping the unlogical `!reload_overview`):
```
GsCategory *category = !reload_overview ? NULL : side_filter_get_category_by_name (shell, cat_name);
if (category != NULL)
  gs_shell_change_mode (shell, GS_SHELL_MODE_CATEGORY, ...)
else
  reload_overview_and_select_category (shell, cat_name);
```

src/gs-shell.c
```diff
@@ -782,6 +788,37 @@ gs_shell_reload_cb (GsPluginLoader *plugin_loader, GsShell *shell)
...
 }
 
+static void
+gs_shell_reload_and_select_removable_media_category (GsShell *shell,
+						     gboolean reload_overview,
+						     gboolean select_category,
+						     const gchar *category_name)
+{
+	/* using NULL for this below indicates it should not be selected */
+	const char *cat_name = (select_category ? category_name : NULL);
```
joaquimrocha: Do you need to have a flag for selecting the category? Can't you just pass `category_name` as `NULL`?

src/gs-shell.c
```diff
@@ -782,6 +788,37 @@ gs_shell_reload_cb (GsPluginLoader *plugin_loader, GsShell *shell)
...
+	}
+	gs_shell_activate (shell);
+}
+
+static void
+gs_shell_copy_dests_notify_cb (GsPluginLoader *plugin_loader,
+			       GParamSpec *pspec,
+			       GsShell *shell)
+{
+	gs_shell_reload_and_select_removable_media_category (shell, TRUE, FALSE, "usb");
```
joaquimrocha: Yeah, this is a bit cryptic, `gs_shell_reload_and_select_removable_media_category (shell, TRUE, NULL);` would be better. But actually only reloading the shell or selecting the removable category would even be clearer.

src/gs-shell.c
```diff
@@ -2090,19 +2131,11 @@ gs_shell_file_to_app_cb (GObject *source,
...
-			else {
-				gs_shell_change_mode (shell, GS_SHELL_MODE_CATEGORY,
-						      (gpointer) category, TRUE);
-			}
-		}
-		gs_shell_activate (shell);
+		gboolean reload_overview =
+			(gs_app_get_metadata_item (app, "EndlessOS::ReloadOverview") == NULL);
+		gs_shell_reload_and_select_removable_media_category (shell, reload_overview,
+								     TRUE, removable_media_cat);
```
joaquimrocha: Again, this could be made more clear.

src/gs-category-page.c
```diff
@@ -280,6 +280,16 @@ gs_category_page_create_filter_list (GsCategoryPage *self,
 			g_debug ("not showing %s/%s as no apps",
 				 gs_category_get_id (category),
 				 gs_category_get_id (s));
+
+			// re-filter USB category with no apps so the
+			// placeholder app tiles get cleared out, then set
+			// "empty state" message for an empty USB disk
```
joaquimrocha: Use C-style comments.

src/gs-category-page.c
```diff
@@ -280,6 +280,16 @@ gs_category_page_create_filter_list (GsCategoryPage *self,
...
+			if (g_strcmp0 (gs_category_get_id (category), "usb") == 0) {
+				gs_category_page_populate_filtered (self, s);
+
+				// TODO: show a "No apps available on USB" message
```
joaquimrocha: Use a C-style comment, and add the comment above the function call.

plugins/flatpak/gs-flatpak.c
```diff
@@ -442,6 +442,11 @@ gs_flatpak_mark_apps_from_usb_remote (GsFlatpak *self,
 				 app_collection_id, ref_collection_id);
 			continue;
 		}
+
+		// TODO: to the USB category (possibly not in here), add top
+		// banner with "copy OS to USB" button (which calls
+		// eos-updater-prepare-flatpak upon click)
```
joaquimrocha: Use C-style comments.

plugins/flatpak/gs-flatpak.c
```diff
@@ -442,6 +442,11 @@ gs_flatpak_mark_apps_from_usb_remote (GsFlatpak *self,
 				 app_collection_id, ref_collection_id);
 			continue;
 		}
+
+		// TODO: to the USB category (possibly not in here), add top
```
joaquimrocha: Yeah, plugins cannot change the UI directly. So this needs to be like you had said: add it to the category page UI file, then show it only when appropriate.

lib/gs-plugin-loader.c
treitter: Fixed locally

lib/gs-plugin-loader.c
treitter: Fixed locally

lib/gs-plugin-loader.c
treitter: Fixed locally

src/gs-overview-page.c
treitter: Fixed locally

src/gs-overview-page.c
treitter: You're right. It's easy to forget that `GList*` is a unique snowflake where the default state of the object is not actually being an object :smile:

Fixed locally.

src/gs-overview-page.c
treitter: Fixed locally

src/gs-shell.c
treitter: I split this out in a separate patch for clarity and cleaned up this comment

src/gs-shell.c
treitter: Fixed locally

src/gs-shell.c
treitter: Fixed locally

src/gs-shell.c
treitter: It seems to be a Builder bug. It looks fine in Builder (double-checked all the tabs and spacing) but it also shows mis-aligned in `gnome-terminal`, so I guess Builder is displaying it incorrectly for some reason.

Fixed locally.

src/gs-shell.c
treitter: Yeah, splitting it seems like the cleanest option. There's trivial code duplication but it's three lines and one of those is just fetching `priv`, etc. This helps clean up the code a decent amount and lets me cut out one of the commits I described above.

src/gs-shell.c
treitter: I noticed that earlier but apparently lost my note to address it.

It's carried directly over from this commit:
https://github.com/endlessm/gnome-software/commit/54863ceaf54cf5128f19b7f6e4338905a1b155d0#diff-b556be19f0ccc1af7d4666e07e9b7c04R2086

Do you know why that was written the way it was? I think I see a way to simplify it but I'm worried about somehow missing some important but subtle side effect.

As it is, I added a commit in my local branch to note the apparent confusion but haven't addressed it.

After the clean-up I mentioned in my last comment, this code is basically intact as it was in your commit I linked in this comment.

src/gs-shell.c
treitter: Now that I'm leaving this code unmodified and unmoved, I'll leave the `else` braces in place on the assumption it's not worth changing the code history. Let me know if you'd still like it changed.

src/gs-category-page.c
treitter: Fixed locally (though this will disappear when implemented)

src/gs-shell.c
treitter: This confusing portion goes away with recent simplification in my local branch

src/gs-shell.c
treitter: This goes away in recent simplification

src/gs-shell.c
treitter: This also goes away in recent simplification

src/gs-shell.c
treitter: This also goes away in recent simplification

src/gs-category-page.c
treitter: Fixed locally

plugins/flatpak/gs-flatpak.c
treitter: Fixed locally (though this will disappear when implemented)

plugins/flatpak/gs-flatpak.c
treitter: Will do

src/gs-shell.c
joaquimrocha: In Emacs it's usually pretty consistent, you can verify with it too if you want.

treitter:
DO NOT MERGE

This version should address the issues you raised above and it's a bit simpler.

There are a couple outstanding issues (hence the "DO NOT MERGE"):
* the "NEEDS WORK -- Add 'copy OS to USB' button to USB category" commit should add a GtkBox with a button to the top of the USB category. For some reason, they don't appear and all other categories print a number of `GTK_IS_*()` check failures for other widgets in the layout. It's almost like the wrong pointers are being checked in each case like there's some struct alignment issue. Except, I added the pointers for the new widgets at the end of the struct and none of them have IDs that conflict with existing widgets in the `.ui` file. If you could look at that, maybe you'll find some obvious error I'm making
* I need to set up the `clicked` handler for the "Copy OS to USB" button to call `eos-updater-prepare-volume` but that's trivial; I just haven't done it because the issue above means the button isn't even displayed currently

lib/gs-plugin-loader.c
```diff
@@ -2527,13 +2505,11 @@ mount_removed_cb (GVolumeMonitor *volume_monitor,
...
 	GsPluginLoaderPrivate *priv = gs_plugin_loader_get_instance_private (plugin_loader);
-	GMount *existing_mount = find_mount_in_list (priv->removable_mounts,
-						     mount);
-	if (existing_mount) {
-		priv->removable_mounts = g_list_remove (priv->removable_mounts,
-							existing_mount);
+	GList *list_orig = priv->removable_mounts;
+
+	priv->removable_mounts = g_list_remove (priv->removable_mounts, mount);
+	if (list_orig != priv->removable_mounts)
```
joaquimrocha: This means the notify will only happen if the mount that got removed was the first of the list, is this the intention? I think it should always notify if one of the mount points got removed, right? (it's also what the code was doing before)

src/gs-shell.c
```diff
@@ -2091,6 +2091,9 @@ gs_shell_file_to_app_cb (GObject *source,
 	removable_media_cat = gs_app_get_metadata_item (app, "EndlessOS::RemovableMediaCategory");
 	if (removable_media_cat != NULL) {
 		GsCategory *category;
+		/* XXX: this seems to be a logic error; if the app does not have
+		 * metadata indicating the overview should be reloaded, it's
+		 * reloaded anyway */
```
joaquimrocha: Interesting... This is definitely worth investigating.

src/gs-category-page.c
```diff
@@ -214,6 +216,23 @@ gs_category_page_reload (GsPage *page)
 		gtk_widget_set_visible (self->infobar_category_shell_extensions, FALSE);
 	}
 
+
+	if (g_strcmp0 (gs_category_get_id (self->category), "usb") == 0) {
+		// FIXME: cut
```
joaquimrocha: Use a C-style comment. Also, what does this mean?

src/gs-category-page.c
```diff
@@ -214,6 +216,23 @@ gs_category_page_reload (GsPage *page)
...
+		g_debug ("category-page: reloading USB");
+
+
+		gtk_widget_set_visible (self->usb_action_box, TRUE);
+
+		/* TODO: hook up "clicked" listener that calls eos-updater-prepare-flatpak
+		 */
+	} else {
+		// FIXME: cut
```
joaquimrocha: Previous comment applies here too.

src/gs-category-page.c
```diff
@@ -257,6 +276,8 @@ filter_selected (GtkListBox *filters, GtkListBoxRow *row, gpointer data)
 	if (row == NULL)
 		return;
 
+	//FIXME: use GDB and fatal warnings to track this down
+	// FIXME: this or something similar in this file assumes children are the content box but that's no longer true due to the addition of the USB box
```
joaquimrocha: Use C-style comments. And use the `/* FIXME` consistently (notice one has a space before, the other has none).

lib/gs-plugin-loader.c
```diff
@@ -2501,23 +2486,16 @@ mount_added_cb (GVolumeMonitor *volume_monitor,
...
-	/* update the list to remove any equivalent mounts and move this to the
-	 * front so we prioritize mounts by recency
+	/* update the list to remove this mount if it already exists mounts move
+	 * it to the front so we prioritize mounts by recency
 	 */
-	if (existing_mount)
-		priv->removable_mounts = g_list_remove (priv->removable_mounts,
-							existing_mount);
-
+	priv->removable_mounts = g_list_remove (priv->removable_mounts, mount);
```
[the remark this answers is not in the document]
treitter: good catch. Fixed locally

lib/gs-plugin-loader.c
treitter: Ah, I mixed that up slightly. Yes, it should happen as long as there was any change, so I'm now comparing based on length before and after the removal. Fixed locally.

src/gs-category-page.c
treitter: I leave myself a number of FIXMEs and TODOs while developing and always check for those strings before merging (and usually before sharing, hence why I didn't bother using C-style comments for these, generally don't bother wrapping them well, etc.).

At any rate, they're removed for the feature-complete version of the branch I'm about to push.

lib/gs-plugin-loader.c
```diff
@@ -2505,10 +2505,11 @@ mount_removed_cb (GVolumeMonitor *volume_monitor,
...
 {
 	GsPluginLoaderPrivate *priv = gs_plugin_loader_get_instance_private (plugin_loader);
-	GList *list_orig = priv->removable_mounts;
+	guint orig_len = g_list_length (priv->removable_mounts);
 
-	priv->removable_mounts = g_list_remove (priv->removable_mounts, mount);
-	if (list_orig != priv->removable_mounts)
+	if (priv->removable_mounts != NULL)
+		priv->removable_mounts = g_list_remove (priv->removable_mounts, mount);
+	if (orig_len != g_list_length (priv->removable_mounts))
```
joaquimrocha: This Is why I think it'd be better to keep a `has_mount` function as before, or a `remove_if_exists -> boolean`.

src/gs-category-page.c
```diff
@@ -340,12 +343,13 @@ gs_category_page_create_filter_list (GsCategoryPage *self,
...
 			if (g_strcmp0 (gs_category_get_id (category), "usb") == 0) {
-				/* TODO: show a "No apps available on USB"
-				 * message (after populating?) */
-
 				gs_category_page_populate_filtered (self, s);
+				category_is_usb = TRUE;
 			}
 
+			gtk_widget_set_visible (self->no_apps_box, category_is_usb);
+			gtk_widget_set_visible (self->scrolledwindow_category, !category_is_usb);
```
joaquimrocha: Nitpick: You could bind the `visible` properties with the invert option. But we can merge this as is as I know you're in a rush.

PR #408, opened 2018-09-21
T23856: migrate 61 apps to Flathub

ramcq:
Remove blacklist entries, and add anything we were shipping on ARM on eos-apps to the flathub whitelist. (This is no guarantee that the Flathub version is working/useful, but is a reasonable starting point.)


ramcq:
https://phabricator.endlessm.com/T23856

PR #409, opened 2018-09-23
T23856: migrate 61 apps to Flathub

ramcq:
Remove blacklist entries, and add anything we were shipping on ARM on eos-apps to the flathub whitelist. (This is no guarantee that the Flathub version is working/useful, but is a reasonable starting point.)

https://phabricator.endlessm.com/T23856


plugins/eos/gs-plugin-eos.c
```diff
@@ -1045,6 +1045,7 @@ gs_plugin_eos_blacklist_app_for_remote_if_needed (GsPlugin *plugin,
 		"com.github.gkarsay.parlatype",
 		"com.github.gyunaev.spivak",
 		"com.github.hluk.copyq",
+		"com.github.labyrinth_team.labyrinth",
```
wjt: @nedrichards did you say you were testing some of these?

PR #418, opened 2018-10-05
T23944 os copy button hiding

treitter:
Only show the "Copy OS to USB" button if an OS copy has a chance of succeeding.

For the eos plugin, this means a collection ID is set on the remote and a USB disk is inserted.

https://phabricator.endlessm.com/T23944


plugins/eos/meson.build
```diff
@@ -25,5 +27,5 @@ shared_module(
   install : true,
   install_dir: plugin_dir,
   c_args : cargs,
-  dependencies : [ plugin_libs, flatpak ]
+  dependencies : [ plugin_libs, flatpak, ostree, ]
```
mwleeds: not sure if the trailing comma here makes a difference

pwithnall: Shouldn’t do. Meson ignores trailing commas (thank god).

lib/gs-plugin-types.h
```diff
@@ -320,6 +320,7 @@ typedef enum {
 	GS_PLUGIN_ACTION_PURCHASE,
 	GS_PLUGIN_ACTION_GET_COPYABLE,
 	GS_PLUGIN_ACTION_COPY,
+	GS_PLUGIN_ACTION_OS_GET_COPYABLE,
```
pwithnall: Needs to be added to the documentation above.

pwithnall: I assume it’s not a problem that adding it here will shift the values of the following elements in the enum? i.e. The numeric values of the elements are not part of any API?

lib/gs-plugin-vfuncs.h
```diff
@@ -623,6 +623,26 @@ gboolean	 gs_plugin_app_copy			(GsPlugin	*plugin,
...
+ * @copy_dest: an absolute path of the destination directory
+ * @copyable: whether this app can be copied
+ * @cancellable: a #GCancellable, or %NULL
+ * @error: a #GError, or %NULL
+ *
+ * Determine whether gs_plugin_os_copy() with this @app and @copy_dest should
+ * succeed (barring runtime errors which cannot be known in advance).
+ *
+ * Returns: %FALSE if copying the OS to @copy_dest will certainly fail; %TRUE
+ * otherwise
```
pwithnall: How does the return value differ from the value returned in `copyable`?

plugins/eos/gs-plugin-eos.c
```diff
@@ -2290,6 +2291,62 @@ gs_plugin_app_upgrade_download (GsPlugin *plugin,
...
+	ostree_parse_refspec (refspec, &remote, NULL, NULL);
+	if (remote == NULL)
+		return NULL;
+
+	repo = ostree_repo_new_default ();
+
+	if (!ostree_repo_open (repo, NULL, NULL))
+		return NULL;
+
+	ostree_repo_get_remote_option (repo, remote, "collection-id", NULL, &collection_id, &error);
```
pwithnall: Nitpick: `error` is unused here.

lib/gs-plugin-vfuncs.h
pwithnall: From the implementations, it seems like the return value actually corresponds to `*error == NULL`.

src/gs-category-page.c
```diff
@@ -230,6 +235,97 @@ copy_os_to_usb_button_cb (GtkButton *button, GsCategoryPage *self)
...
+	}
+
+	gtk_widget_set_visible (self->usb_action_box, TRUE);
+	gtk_widget_set_sensitive (self->copy_os_to_usb_button, FALSE);
+	gtk_widget_set_visible (self->copy_os_to_usb_button, FALSE);
+
+	if (self->copying)
+		return;
+
+	copy_dest = removable_destination_dup (self);
```
pwithnall: Why does `removable_destination_dup()` need to `g_strdup()` the string? The duplicate is only used locally here, then freed.

src/gs-category-page.c
```diff
@@ -230,6 +235,97 @@ copy_os_to_usb_button_cb (GtkButton *button, GsCategoryPage *self)
...
+	if (copy_dest != NULL) {
+		gtk_button_set_label (GTK_BUTTON (self->copy_os_to_usb_button), _("Copy OS to USB"));
+		gtk_widget_set_sensitive (self->copy_os_to_usb_button, TRUE);
+
+		/* also check (asynchronously) whether the OS will
+		 * certainly fail a copy so we can update the UI
+		 * accordingly. Cases where the plugin would know a copy
+		 * would fail include an OSTree remote which lacks a collection
+		 * ID or broader issues like the user lacking write
+		 * access to the destination. */
```
pwithnall: I don’t see any code which checks whether the user has write access to the destination.

pwithnall: Would you really want to hide the button in that case? It seems more user friendly to show the button, and then present the user with a suitable error message. Or, even better, chmod the destination so we actually do have write access.

Conversely, I think it makes sense to disable the button if no collection ID is set, because that’s beyond the control of any (non-technical) user.

src/gs-category-page.c
```diff
@@ -230,6 +235,97 @@ copy_os_to_usb_button_cb (GtkButton *button, GsCategoryPage *self)
...
+			"copy-dest", copy_dest,
+			 NULL);
+		gs_plugin_loader_job_os_get_copyable_async (
+			self->plugin_loader,
+			plugin_job,
+			self->cancellable,
+			gs_category_page_os_get_copyable_cb,
+			self);
+	} else {
+		gtk_button_set_label (GTK_BUTTON (self->copy_os_to_usb_button), _("Insert USB to copy"));
```
pwithnall: Nitpick: The GNOME style guide says that button labels should be in Title Case, i.e. `Insert USB to Copy`. A mnemonic would also be good.

pwithnall: Should it also be `Insert USB to Copy To`?

src/gs-category-page.c
mwleeds: To be consistent with the behavior when no collection ID is configured on an app remote, the button shouldn't show at all.

src/gs-category-page.c
```diff
@@ -230,6 +235,97 @@ copy_os_to_usb_button_cb (GtkButton *button, GsCategoryPage *self)
...
+	gtk_widget_set_sensitive (self->copy_os_to_usb_button, FALSE);
+	gtk_widget_set_visible (self->copy_os_to_usb_button, FALSE);
+
+	if (self->copying)
+		return;
+
+	copy_dest = removable_destination_dup (self);
+	if (copy_dest != NULL) {
+		gtk_button_set_label (GTK_BUTTON (self->copy_os_to_usb_button), _("Copy OS to USB"));
+		gtk_widget_set_sensitive (self->copy_os_to_usb_button, TRUE);
```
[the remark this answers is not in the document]
treitter: In my local branch, I cleaned this all up by making it all invisible here then centralizing all later `gtk_widget_set_{sensitive,visible}()` calls to the callback

lib/gs-plugin-types.h
treitter: It's technically public API, so I've fixed the ordering locally.

In practice, it probably won't matter since this would only be an API break downstream at this point and we'll keep the library and all dependents synchronized. The only dependents are plugins and all the ones we use, as far as I know, are in this source tree. It's possible third-party plugins exist but I don't know.

lib/gs-plugin-vfuncs.h
treitter: `copyable` is independent of the return value. The return value of the function does match `*error == NULL` as all the other similar functions. It's redundant with `*error` for convenience (so you can put it in an `if` statement and in case you don't care about the actual `error`). This function signature pattern is also extremely common in `g-s` so I don't think I should break it.

plugins/eos/gs-plugin-eos.c
treitter: Locally, I filled in all the potential uses of `error` in this function.

src/gs-category-page.c
treitter: > I don’t see any code which checks whether the user has write access to the destination.

None exists; that's just an example of what could be done. It's mostly highlighting the design of the API in case other plugin developers want to support that use case (or in case we eventually do).

> Would you really want to hide the button in that case? It seems more user friendly to show the button, and then present the user with a suitable error message. Or, even better, chmod the destination so we actually do have write access.

I can see that argument though what I currently do follows the convention used in similar contexts (eg, app copying which follows the convention for making an app's desktop shortcut, etc.).

The UX needs review from a designer in general, so I'd prefer to follow convention and punt until we get to that point.

As far as ensuring the proper permissions, as of recently, we ensure the `.ostree` dir is writable whenever we make it. If the `.ostree` dir doesn't exist and we don't have write access, I don't think we should try `chmod`ing the parent dir.

> Conversely, I think it makes sense to disable the button if no collection ID is set, because that’s beyond the control of any (non-technical) user.

I agree if we're showing the button as you suggested above but let's address that if/when we reverse the convention.

src/gs-category-page.c
treitter: > Nitpick: The GNOME style guide says that button labels should be in Title Case, i.e. `Insert USB to Copy`. A mnemonic would also be good.

I'll file a follow-up task. We just got this string translated into a few languages and I assume you need a perfect match to take advantage of the translations.

> Should it also be `Insert USB to Copy To`?

Shouldn't it be `Insert USB to Which to Copy`? :)

The string I used has trivially different meaning but the point of both is the same. In that case, shorter wins. As it stands, the German translation is already long enough: "Einfügen von USB zum Kopieren". (I'm not sure the "von" is needed but Google Translate seems to be reasonably accurate for German and I get increasingly inaccurate with time).

treitter:
I just pushed some changes which should address everyone's concerns.

plugins/eos/gs-plugin-eos.c
mwleeds: Could use an `if (error != NULL) goto err_out;` here

src/gs-category-page.c
```diff
@@ -230,6 +235,82 @@ copy_os_to_usb_button_cb (GtkButton *button, GsCategoryPage *self)
...
+{
+	GsPluginLoader *plugin_loader = GS_PLUGIN_LOADER (source_object);
+	GsCategoryPage *self = GS_CATEGORY_PAGE (user_data);
+	g_autoptr(GError) error = NULL;
+	gboolean copyable;
+	const char *copy_button_label;
+
+	copyable = gs_plugin_loader_job_os_get_copyable_finish (plugin_loader,
+								res, &error);
+	copy_button_label = (copyable ? _("Copy OS to USB") : _("Insert USB to copy"));
```
mwleeds: The way it's written now doesn't distinguish between the "Insert USB to copy" and "Copying is impossible" cases. Why set the label just to then make the button invisible?

mwleeds:
Alright with the latest fixup this looks good. For a minute I was confused because I thought we were losing the feature where it tells you to "Insert USB to copy" but that only makes sense for the app details pages, since the USB category doesn't show with no USBs plugged in.

lib/gs-plugin-vfuncs.h
pwithnall: Documentation needs updating then, I think.

PR #419, opened 2018-10-05
src: Fix Category Box Visibility

mazen-asef:
Fix issue where the empty apps layout box appears on the USB categories
when a USB with loaded apps is inserted by checking after a page reload
if there are really no apps in it.

https://phabricator.endlessm.com/T23418


plugins/flatpak/gs-flatpak.c
```diff
@@ -3641,7 +3641,7 @@ gs_flatpak_mount_removed (GVolumeMonitor *volume_monitor,
 	GsFlatpak *self = GS_FLATPAK (user_data);
 	GHashTableIter iter;
 	gpointer key;
-	gboolean should_reload = FALSE;
+	//gboolean should_reload = FALSE;
```
treitter: I think all the changes to this file need to be reverted since they cause a regression: if you have the USB inserted and copy an app to it, the USB category no longer gets reloaded to reflect the changes (as it did before this patch).

I don't immediately see why leaving this in place would conflict with the rest of the patch.

src/gs-category-page.c
```diff
@@ -342,15 +342,28 @@ gs_category_page_create_filter_list (GsCategoryPage *self,
 			 * placeholder app tiles get cleared out, then set
 			 * "empty state" message for an empty USB disk
 			 */
-			if (g_strcmp0 (gs_category_get_id (category), "usb") == 0) {
-				gs_category_page_populate_filtered (self, s);
+			if (g_strcmp0 (gs_category_get_id (category), "usb") != 0) {
+				gtk_widget_set_visible (self->no_apps_box, category_is_usb);
+				gtk_widget_set_visible (self->scrolledwindow_category,
```
treitter: There's trailing whitespace on this line and some others in this patch.

`git diff` should color this whitespace in red for you if you need help finding it or just do a regex search for ` $`

src/gs-category-page.c
```diff
@@ -342,15 +342,28 @@ gs_category_page_create_filter_list (GsCategoryPage *self,
 			 * placeholder app tiles get cleared out, then set
 			 * "empty state" message for an empty USB disk
 			 */
-			if (g_strcmp0 (gs_category_get_id (category), "usb") == 0) {
-				gs_category_page_populate_filtered (self, s);
+			if (g_strcmp0 (gs_category_get_id (category), "usb") != 0) {
```
treitter: You should be able to factor out these common calls to `gtk_widget_set_visible()` and `continue` with something like:

```
gboolean have_usb_apps = FALSE;
if (g_strcmp0 (gs_category_get_id (category), "usb") == 0) {
    				GPtrArray *usb_apps;
￼ 				category_is_usb = TRUE;
￼ 				gs_category_page_populate_filtered (self, s);
￼ 				/* After reloading the page it will get the actual apps for the
￼                                  * for the usb so we test if the category will truly have the 
￼                                  * no apps box.
￼                                  */
￼ 				usb_apps = gs_category_get_children (category);
￼ 				if (usb_apps != NULL && usb_apps->len >= 0)
￼ 				￼ 	  have_usb_apps = TRUE;
}

if (!have_usb_apps) {
					gtk_widget_set_visible (self->no_apps_box, category_is_usb);
					gtk_widget_set_visible (self->scrolledwindow_category, 
							!category_is_usb);
					continue;
}
```

(but fix the horrific indentation in my example)

src/gs-category-page.c
```diff
@@ -342,15 +342,28 @@ gs_category_page_create_filter_list (GsCategoryPage *self,
...
+			} else {
+				GPtrArray *usb_apps;
 				category_is_usb = TRUE;
+				gs_category_page_populate_filtered (self, s);
+				/* After reloading the page it will get the actual apps for the
+                                 * for the usb so we test if the category will truly have the 
+                                 * no apps box.
+                                 */
+				usb_apps = gs_category_get_children (category);
+				if (usb_apps->len == 0) {
```
treitter: I hope you like segfaults :)

This needs to be:
```
if (usb_apps != NULL && usb_apps->len == 0) {
```
or you'll crash when you insert a USB disk with no apps.

src/gs-category-page.c
```diff
@@ -342,15 +342,28 @@ gs_category_page_create_filter_list (GsCategoryPage *self,
...
+				gtk_widget_set_visible (self->scrolledwindow_category, 
+							!category_is_usb);
+
+				continue;
+			} else {
+				GPtrArray *usb_apps;
 				category_is_usb = TRUE;
+				gs_category_page_populate_filtered (self, s);
+				/* After reloading the page it will get the actual apps for the
+                                 * for the usb so we test if the category will truly have the
```
treitter: It looks like this and the following lines aren't aligned with the `*` in the initial line

mwleeds:
@mazen-asef I'm confused because the title of the ticket, the description of the ticket, and the commit message seem to all three describe distinct behaviors. Can we synchronize them? Are you sure the duplicate apps problem is gone? What is the "empty apps layout box" problem you're describing here?

mazen-asef:
@mwleeds It was regarding the issue on https://phabricator.endlessm.com/T23418 (after I changed the name of the ticket) , where inserting a USB a second time didn't show the apps installed on it, and if you booted up gnome-software without a USB, it would show that there are no apps installed in it.

Also, I renamed the PR decription and commit.

src/gs-category-page.c
```diff
@@ -333,24 +333,34 @@ gs_category_page_create_filter_list (GsCategoryPage *self,
...
+					apps_detected = TRUE;
+			}
+			/* Then set "empty state" message for an empty USB disk. */
+			if (!apps_detected) {
+				g_debug ("not showing %s/%s as no apps",
+					 gs_category_get_id (category),
+					 gs_category_get_id (s));
+
+				gtk_widget_set_visible (self->no_apps_box, category_is_usb);
+				gtk_widget_set_visible (self->scrolledwindow_category, !category_is_usb);
```
joaquimrocha: I think this code would be much simpler if you rename the var `category_is_usb` as `show_no_apps_box` and just set its state, instead of having now `apps_detected` and `category_is_usb`.

src/gs-category-page.c
```diff
@@ -332,25 +332,31 @@ gs_category_page_create_filter_list (GsCategoryPage *self,
...
+			}
+			/* Then set "empty state" message for an empty USB disk. */
+			if ((usb_apps == NULL) || (usb_apps->len < 1)) {
+	 			g_debug ("not showing %s/%s as no apps",
+					 gs_category_get_id (category),
+					 gs_category_get_id (s));
+
+ 				gtk_widget_set_visible (self->no_apps_box, show_no_apps_box);
+				gtk_widget_set_visible (self->scrolledwindow_category, !show_no_apps_box);
+	 			continue;
```
joaquimrocha: This is more of less doing what you had before, right?
Can you move the `gtk_widget_set_visible` calls outside, as they were initially? Otherwise this logic is not always running and is harder to understand. You should just set the `show_no_apps_box` variable from where you currently are, but also assign it according to the `usb_apps` value. (Don't forget to declare `usb_apps` inside that if's body, which makes it also all more contained)

src/gs-category-page.c
```diff
@@ -332,25 +332,31 @@ gs_category_page_create_filter_list (GsCategoryPage *self,
...
+				g_debug ("not showing %s/%s CATEGORY USB CARAI",
+					 gs_category_get_id (category),
+					 gs_category_get_id (s));
 				gs_category_page_populate_filtered (self, s);
-				category_is_usb = TRUE;
+				usb_apps = gs_category_get_children (category);
+				show_no_apps_box = TRUE;
+			}
+			/* Then set "empty state" message for an empty USB disk. */
+			if ((usb_apps == NULL) || (usb_apps->len < 1)) {
```
joaquimrocha: Use `usb_apps->len == 0` instead of `usb_apps->len < 1`. it is unsigned, so being `< 1` always means it's 0.

src/gs-category-page.c
```diff
@@ -332,25 +332,31 @@ gs_category_page_create_filter_list (GsCategoryPage *self,
...
-				 gs_category_get_id (category),
-				 gs_category_get_id (s));
-
-			/* re-filter USB category with no apps so the
+ 			/* re-filter USB category with no apps so the
 			 * placeholder app tiles get cleared out, then set
 			 * "empty state" message for an empty USB disk
 			 */
 			if (g_strcmp0 (gs_category_get_id (category), "usb") == 0) {
+				g_debug ("not showing %s/%s CATEGORY USB CARAI",
```
joaquimrocha: :scream_cat:  Maybe make the debug message less slangish.

src/gs-category-page.c
```diff
@@ -332,25 +332,31 @@ gs_category_page_create_filter_list (GsCategoryPage *self,
...
+					 gs_category_get_id (category),
+					 gs_category_get_id (s));
 				gs_category_page_populate_filtered (self, s);
-				category_is_usb = TRUE;
+				usb_apps = gs_category_get_children (category);
+				show_no_apps_box = TRUE;
+			}
+			/* Then set "empty state" message for an empty USB disk. */
```
joaquimrocha: This comment is repeated from the previous one.

pwithnall:
@mazen-asef, this needs to be rebased to sit on top of the rebased GNOME Software 3.32, and needs Joaquim’s review comments to be taken care of.

mwleeds:
This problem is addressed by #489

PR #420, opened 2018-10-05
plugins: fix documentation for gs_plugin_os_get_copyable()

treitter:
It had a reference to an `app` parameter that doesn't exist.

### DIFF
--- a/lib/gs-plugin-vfuncs.h
+++ b/lib/gs-plugin-vfuncs.h
@@ -594,8 +594,10 @@ gboolean	 gs_plugin_app_purchase			(GsPlugin	*plugin,
  * Determine whether gs_plugin_app_copy() with this @app and @copy_dest should
  * succeed (barring runtime errors which cannot be known in advance).
  *
- * Returns: %FALSE if copying @app to @copy_dest will certainly fail; %TRUE
- * otherwise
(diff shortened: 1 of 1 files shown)


lib/gs-plugin-vfuncs.h
```diff
@@ -631,8 +631,8 @@ gboolean	 gs_plugin_app_copy			(GsPlugin	*plugin,
  * @cancellable: a #GCancellable, or %NULL
  * @error: a #GError, or %NULL
  *
- * Determine whether gs_plugin_os_copy() with this @app and @copy_dest should
- * succeed (barring runtime errors which cannot be known in advance).
+ * Determine whether gs_plugin_os_copy() with this @copy_dest should succeed
+ * (barring runtime errors which cannot be known in advance).
  *
  * Returns: %FALSE if copying the OS to @copy_dest will certainly fail; %TRUE
  * otherwise
```
pwithnall: I was meaning that the `Returns` documentation seems to be about `copyable`, rather than whether determining the copyability failed.

treitter: Ah, got it. I fixed this and the same mistake for `gs_plugin_app_get_copyable()`, rebased on master, and force-pushed. Please re-review.

PR #423, opened 2018-10-09
T23939: fix state tracking/transition around LOADING state

ramcq:
There were a few different bugs where the shell and the application's idea of whether the shell had finished loading could be bypassed or get out of sync. This branch cleans the shell logic up so that the shell doesn't emit the "loading" signal or leave the LOADING mode until the initial refresh has been completed, and the shell and application can use these guarantees to prevent other actions being started too early.

The main "culprit" here was the USB app installation code reloading the overview page as new mount points appeared, which happens asynchronously during the application startup. This was starting the loading of categories and popular apps before the initial refresh had completed, which was causing live locks between the plugins, and the categories and popular apps staying empty for the lifespan of the app.

Also squash various other new warnings caused by the USB app export code, things being accessed before being initialised, or updated un-necessarily.

https://phabricator.endlessm.com/T23939

### DIFF
--- a/lib/gs-plugin-loader.c
+++ b/lib/gs-plugin-loader.c
@@ -2620,7 +2620,7 @@ mount_added_cb (GVolumeMonitor *volume_monitor,
 	GsPluginLoaderPrivate *priv = gs_plugin_loader_get_instance_private (plugin_loader);
 	g_autoptr(GDrive) drive = g_mount_get_drive (mount);
 
-	if (!g_drive_is_removable (drive))
+	if (drive == NULL || !g_drive_is_removable (drive))
 		return;
 
 	/* remove this mount if it already exists in the list and add it to the
--- a/src/gs-application.c
+++ b/src/gs-application.c
@@ -883,9 +883,6 @@ wrapper_action_activated_cb (GSimpleAction *action,
 								       G_SIMPLE_ACTION (real_action),
 								       g_variant_ref (parameter));
 
-		g_signal_handler_disconnect (app->shell, app->shell_loaded_handler_id);
-		app->shell_loaded_handler_id = 0;
-
 		g_signal_connect_swapped (app->shell, "loaded",
 					  G_CALLBACK (activate_on_shell_loaded_cb), helper);
 		return;
--- a/src/gs-details-page.c
+++ b/src/gs-details-page.c
@@ -2448,6 +2448,11 @@ gs_details_page_plugin_status_changed_cb (GsPluginLoader *plugin_loader,
                                           GsPluginStatus status,
                                           GsDetailsPage *self)
 {
+	self->plugin_status = status;
+
+	if (app == NULL)
+		return;
+
 	if (status == GS_PLUGIN_STATUS_COPYING) {
 		/* prevent the app from being deleted, etc. while it's being copied */
 		gs_app_add_quirk (app, AS_APP_QUIRK_COMPULSORY);
@@ -2455,8 +2460,6 @@ gs_details_page_plugin_status_changed_cb (GsPluginLoader *plugin_loader,
 		/* allow removal once copying has completed */
 		gs_app_remove_quirk (app, AS_APP_QUIRK_COMPULSORY);
 	}
-
-	self->plugin_status = status;
 }
 
 static gboolean
@@ -2484,7 +2487,6 @@ gs_details_page_setup (GsPage *page,
 	g_signal_connect (plugin_loader, "notify::copy-dests",
 			  G_CALLBACK (gs_details_page_copy_dests_notify_cb),
 			  self);
-	gs_details_page_copy_dests_notify_cb (plugin_loader, NULL, self);
 
 	/* show review widgets if we have plugins that provide them */
 	self->enable_reviews =
--- a/src/gs-shell.c
+++ b/src/gs-shell.c
@@ -786,6 +786,12 @@ gs_shell_copy_dests_notify_cb (GsPluginLoader *plugin_loader,
 	GsShellPrivate *priv = gs_shell_get_instance_private (shell);
 	GsPage *page = GS_PAGE (g_hash_table_lookup (priv->pages, "overview"));
 
+	/* reloading the overview page before the initial refresh is completed
+	   means the categories and popular apps don't get initialised properly
+	   and the UI appears very broken */
+	if (priv->mode == GS_SHELL_MODE_LOADING)
+		return;
+
 	gs_page_reload (page);
 }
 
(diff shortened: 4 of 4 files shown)


lib/gs-plugin-loader.c
```diff
@@ -2620,7 +2620,7 @@ mount_added_cb (GVolumeMonitor *volume_monitor,
 	GsPluginLoaderPrivate *priv = gs_plugin_loader_get_instance_private (plugin_loader);
 	g_autoptr(GDrive) drive = g_mount_get_drive (mount);
 
-	if (!g_drive_is_removable (drive))
+	if (!drive || !g_drive_is_removable (drive))
```
joaquimrocha: Nitpick: In g-s the convention it to use `var != NULL` instead of `!var` (I dislike it but tried to follow when I remembered, as it increases the chances of upstreaming).

src/gs-application.c
```diff
@@ -260,7 +260,7 @@ theme_changed (GtkSettings *settings, GParamSpec *pspec, GsApplication *app)
 static void
 gs_application_shell_loaded_cb (GsShell *shell, GsApplication *app)
 {
-	gs_shell_set_mode (app->shell, GS_SHELL_MODE_OVERVIEW);
+	g_signal_handler_disconnect (app->shell, app->shell_loaded_handler_id);
```
joaquimrocha: Is this whole commit really needed? Is the problem that `gs_application_shell_loaded_cb` is getting called before the `initial_overview_load_done` is called?

src/gs-shell.c
```diff
@@ -1963,6 +1963,9 @@ gs_shell_setup (GsShell *shell, GsPluginLoader *plugin_loader, GCancellable *can
 
 	/* coldplug */
 	gs_shell_rescan_events (shell);
+
+	/* show loading page, which triggers the initial refresh */
+	gs_shell_change_mode (shell, GS_SHELL_MODE_LOADING, NULL, TRUE);
```
joaquimrocha: Isn't this commit equivalent to what was being done (and hence not necessary)? AFAICS only the application calls the `gs_shell_setup`, so these changes seem to only call the mentioned method before the `_add_window`.

src/gs-details-page.c
```diff
@@ -2448,6 +2448,9 @@ gs_details_page_plugin_status_changed_cb (GsPluginLoader *plugin_loader,
                                           GsPluginStatus status,
                                           GsDetailsPage *self)
 {
+	if (!app)
+		return;
```
joaquimrocha: Maybe you should still set the `self->plugin_status` (last line not shown here).

src/gs-details-page.c
```diff
@@ -2448,6 +2448,9 @@ gs_details_page_plugin_status_changed_cb (GsPluginLoader *plugin_loader,
                                           GsPluginStatus status,
                                           GsDetailsPage *self)
 {
+	if (!app)
```
joaquimrocha: Nitpick: The thing I wrote before about `app != NULL` (Richard's demands).

src/gs-shell.c
```diff
@@ -767,9 +767,6 @@ main_window_closed_cb (GtkWidget *dialog, GdkEvent *event, gpointer user_data)
 	g_application_withdraw_notification (g_application_get_default (),
 					     "install-resources");
 
-	/* When the window is closed, reset the initial mode to overview */
-	priv->mode = GS_SHELL_MODE_OVERVIEW;
-
```
joaquimrocha: Maybe we need to make this conditional on whether the `mode` is `LOADING`. Because the idea for this commit is to always open g-s in the overview mode.

src/gs-shell.c
```diff
@@ -790,6 +790,10 @@ gs_shell_copy_dests_notify_cb (GsPluginLoader *plugin_loader,
 	GsShellPrivate *priv = gs_shell_get_instance_private (shell);
 	GsPage *page = GS_PAGE (g_hash_table_lookup (priv->pages, "overview"));
 
+	/* the overview page has not been loaded yet */
```
joaquimrocha: I'd prefer a comment stating why this is important. Something like: `Don't reload if we're still loading, it may break the state/refresh`.
And good catch, BTW!

src/gs-shell.c
```diff
@@ -1972,18 +1972,7 @@ void
...
-	/* if we're loading a different mode at startup then don't wait for
-	 * the overview page to load before showing content */
-	if (mode != GS_SHELL_MODE_OVERVIEW) {
-		GsPage *page = g_hash_table_lookup (priv->pages, "overview");
-		matched = g_signal_handlers_disconnect_by_func (page,
-								initial_overview_load_done,
-								shell);
-		if (matched > 0)
-			g_signal_emit (shell, signals[SIGNAL_LOADED], 0);
-	}
```
joaquimrocha: Did you verify that launching g-s with the `--mode` option still works after this change?

joaquimrocha:
Just a couple of comments. Also, it'd be nice if the commits followed the convention that only plugin changes get a prefix, and would be in the form:
`plugin-name: Capitalized title`
for no plugins, it'd be:
`Capitalized title`

But the changes make sense! Great job!

src/gs-shell.c
ramcq: Yes.

src/gs-shell.c
ramcq: This is also handled by GsApplication setting the mode to overview when it activates the window - I think that it's a bad idea to modify the state like this without also updating the other stuff - so this should either use gs_shell_change_mode and do it "properly", or leave it to the GsApplication code - but doing both is really weird to me.

lib/gs-plugin-loader.c
ramcq: Will fixup.

src/gs-application.c
ramcq: It's not _needed_, no - I just think it's confusing to split the state machine between application and shell, unless application is doing something different to the default.

src/gs-shell.c
ramcq: Correct, it doesn't change the behaviour, although if the code I deleted in a77b763 hadn't been so broken, this would never have worked - which for a while I thought was the bug. Per above it feels more consistent to me that the shell would advance itself through the state machine rather than splitting the state advances between two classes.

src/gs-details-page.c
ramcq: :+1:

src/gs-shell.c
ramcq: Done.

ramcq:
I've fixed up the fixups - I will squash and fix the commit message titles if the code looks OK now.

src/gs-shell.c
joaquimrocha: That makes sense. I trust you tested reopening the window once it's closed?

joaquimrocha:
I've still prefer we didn't include the initial commits that just move things around (unless we're positive they may go upstream, as it's another thing that can break in a rebase).
But it's fine to leave it since we are in a rush, if you prefer not to have to retest things.

src/gs-shell.c
ramcq: Yep!

ramcq:
I've dropped the renaming and moving stuff around commits, and made a separate T23939-full branch that still has them. I will try and upstream all of the applicable parts tomorrow, and pick back the relevant things into master/eos3.5 if they are accepted (I had a chat to Kalev and I think he is amenable), but we can drop them for eos3.4 indeed.

PR #424, opened 2018-10-11
T23939: additional couple of fixes

ramcq:
The patch c4579fb dropped from https://github.com/endlessm/gnome-software/pull/423 is actually needed because it prevents the shell from needlessly going to the overview mode before the set_mode GAction is handled - we unconditionally keep the shell_loaded_handler_id as an indicator of whether loading has completed, and the shell decides whether the loading->overview transition is needed or not.

This in turn exposed another bug where if you go from overview to another mode before the categories have loaded, the side filters get shown on the other mode.

https://phabricator.endlessm.com/T23939


joaquimrocha:
LGTM!

PR #426, opened 2018-10-29
Update hardcoded list of popular apps for 3.5

nedrichards:


ramcq:
Fixed in #427. Monday reviewing fail.

PR #431, opened 2018-11-16
updates section: Be more defensive before triggering reboots

ramcq:
Instead of checking that app state != UPDATABLE_LIVE, check that it is
UPDATABLE. This ensures that we don't do a reboot in case of an
unexpected state, such as INSTALLING.

https://gitlab.gnome.org/GNOME/gnome-software/issues/483
(cherry picked from commit c85432c779d76f566eb3e303a1cdaa577ccb3af0)
Signed-off-by: Robert McQueen <rob@endlessm.com>

Moved equivalent code back to to src/gs-updates-page.c where it used to be in
our version.

https://phabricator.endlessm.com/T19785


ramcq:
YES DO IT!

PR #433, opened 2018-11-26
T24017 Apply parental controls to Launch button

pwithnall:
https://phabricator.endlessm.com/T24017


pwithnall:
> So the OARS filter only applies to installing apps, not to running them?

Correct.



> Otherwise on a system with multiple users (one adult, one child) the adult would have to take care to explicitly blacklist any intense applications they install for their own use.

@nedrichards?

nedrichards:
Absolutely and correct.

wjt:
Okay, I see now that this is explicitly addressed in the design document, with this rationale:

> OARS filtering must only apply to listing and installing apps, and not to running already-installed apps. Otherwise it would be impossible for a parent to choose that a certain app *should* be available to their kid and whitelist it despite its OARS rating.

pwithnall:
> I presume the idea is to upstream adding the quirk to the enum now, but not the other patch?

[Yes](https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/147), sorry for the delay.

PR #434, opened 2018-11-29
trivial: Never show cancelled error messages to the user

uajain:
https://phabricator.endlessm.com/T23928

(cherry-picked from 62f6314f9f1262ac7b3c5d78bb96fe57240f7d4f)

### DIFF
--- a/lib/gs-plugin-loader.c
+++ b/lib/gs-plugin-loader.c
@@ -457,6 +457,12 @@ gs_plugin_error_handle_failure (GsPluginLoaderHelper *helper,
 		return TRUE;
 	}
 
+        /* this is only ever informational */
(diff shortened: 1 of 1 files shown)


lib/gs-plugin-loader.c
```diff
@@ -457,6 +457,12 @@ gs_plugin_error_handle_failure (GsPluginLoaderHelper *helper,
 		return TRUE;
 	}
 
+        /* this is only ever informational */
```
wjt: Is this patch already upstream? If not: the indentation doesn't match the surrounding code.

ramcq: Yeah... is this actually a cherry pick or is it copy/pasted?! It doesn't match https://gitlab.gnome.org/GNOME/gnome-software/commit/62f6314f9f1262ac7b3c5d78bb96fe57240f7d4f

PR #437, opened 2019-01-17
Fix displaying Flathub apps on Featured tab

wjt:
Since we moved various applications to Flathub, they fell out of our Featured tab, even though we attempted to update the hard-coded list.

There turned out to be two bugs here:

* The app IDs in the hard-coded list are neither Flatpak app IDs, nor .desktop file IDs: they are the <id> element from the appdata, which may be one or the other.
* We effectively filtered out any app which does not have a tile background image, which (since this is Endless-specific) effectively filtered out any Flathub apps.

Fixing the second point causes Hatch Previewer, a development tool we have published with a com.endlessm.* app ID to Flathub, to appear in the popular app rotation; I addressed this with a nasty special-case. Alternative suggestions welcome!

https://phabricator.endlessm.com/T23953

### DIFF
--- a/plugins/eos/gs-plugin-eos.c
+++ b/plugins/eos/gs-plugin-eos.c
@@ -2360,6 +2360,9 @@ gs_plugin_add_popular (GsPlugin *plugin,
 		if (!g_strcmp0 (origin, "eos-apps") &&
 		    g_str_has_prefix (app_id, "com.endlessm."))
 			add = TRUE;
+		/* all com.endlessnetwork. apps */
+		else if (g_str_has_prefix (app_id, "com.endlessnetwork."))
+			add = TRUE;
 
 		if (add)
 			gs_app_list_add (new_list, app);
--- a/plugins/eos/gs-plugin-eos.c
+++ b/plugins/eos/gs-plugin-eos.c
@@ -2293,27 +2293,42 @@ gs_plugin_add_popular (GsPlugin *plugin,
 {
 	g_autoptr(AsProfileTask) ptask = NULL;
 	GsAppList *new_list = NULL;
+	/* These IDs must match the <component><id> element in the app's
+	 * appdata. Note in particular that some apps have a .desktop suffix
+	 * there, and some do not.
+	 */
 	const gchar *popular_apps[] = {
 		"com.calibre_ebook.calibre.desktop",
 		"com.google.Chrome.desktop",
 		"com.spotify.Client.desktop",
-		"com.transmissionbt.Transmission.desktop",
+		"com.transmissionbt.Transmission",
 		"com.valvesoftware.Steam.desktop",
 		"io.github.mmstick.FontFinder.desktop",
-		"net.minetest.Minetest.desktop",
-		"net.scribus.Scribus.desktop",
-		"org.audacityteam.Audacity.desktop",
-		"org.gimp.GIMP.desktop",
-		"org.gnome.chess.desktop",
-		"org.gnome.SwellFoop.desktop",
-		"org.inkscape.Inkscape.desktop",
+		"net.minetest.Minetest",
+		"net.scribus.Scribus",
+		"org.audacityteam.Audacity",
+		"org.gimp.GIMP",
+		"org.gnome.Chess",
+		"org.gnome.SwellFoop",
+		"org.inkscape.Inkscape",
 		"org.kde.gcompris.desktop",
-		"org.kde.krita.desktop",
+		"org.kde.krita",
 		"org.libreoffice.LibreOffice.desktop",
 		"org.mozilla.Firefox.desktop",
 		"org.telegram.desktop.desktop",
 		"org.tuxpaint.Tuxpaint.desktop",
 		"org.videolan.VLC.desktop",
+		/* TODO: this doesn't end up in the list of apps used by the
+		 * overview. Relevant log output:
+		 *
+		 *   As  run 0x562bb3977e30~GsPlugin::appstream(gs_plugin_add_popular;gs_plugin_refine_wildcard)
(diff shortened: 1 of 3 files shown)


plugins/flatpak/gs-flatpak.c
```diff
@@ -594,7 +594,8 @@ gs_flatpak_add_apps_from_xremote (GsFlatpak *self,
 
 		/* Mark com.endlessm. apps with GnomeSoftware::popular kudo. See discussion on T23152 */
 		if (!as_app_has_kudo (app, "GnomeSoftware::popular") &&
-		     g_str_has_prefix (as_app_get_id (app), "com.endlessm."))
+		     g_str_has_prefix (as_app_get_id (app), "com.endlessm.") &&
+		     g_strcmp0 (as_app_get_id (app), "com.endlessm.HatchPreviewer") != 0)
```
ramcq: I know it might be a little hair-splitting but my gut feeling is this is more reliably expressed as all com.endlessm apps from eos-apps should be auto-featured, and perhaps we also wish to feature (from flathub) com.endlessm.photos?

plugins/eos/gs-plugin-eos.c
```diff
@@ -2190,27 +2190,42 @@ gs_plugin_add_popular (GsPlugin *plugin,
 {
 	g_autoptr(AsProfileTask) ptask = NULL;
 	GsAppList *new_list = NULL;
+	/* These IDs must match the <component><id> element in the app's
```
ramcq: I wonder if there is a better function to dig in the cached apps to match without hard-coding whether or not the app ID has a .desktop in? https://github.com/endlessm/gnome-software/blob/master/plugins/core/gs-plugin-hardcoded-blacklist.c seems to do some globbing for example.

plugins/eos/gs-plugin-eos.c
```diff
@@ -2190,27 +2190,42 @@ gs_plugin_add_popular (GsPlugin *plugin,
...
+		 *
+		 *   As  run 0x562bb3977e30~GsPlugin::appstream(gs_plugin_add_popular;gs_plugin_refine_wildcard)
+		 *   Gs  not using simple-scan.desktop for wildcard as no bundle or pkgname
+		 *
+		 * This app is built into the ostree. Its ID is, without spaces:
+		 *
+		 *   system / * / * / desktop / simple-scan.desktop / *
+		 *            ^   ^ these are probably bundle and pkgname?
+		 */
 		"simple-scan.desktop",
```
ramcq: It might also be possible to match this with a glob too. :)

wjt:
@ramcq thanks for the review. Yes, it does seem like it may be possible to better, by moving this logic into a refine function rather than splitting it between 2 plugins in this way. I'll see if I can find time for that.

plugins/flatpak/gs-flatpak.c
ramcq: I have implemented this - com.endlessm is now only matched on *.endlessm.com origins.

plugins/eos/gs-plugin-eos.c
ramcq: This is still broken, and in fact - this isn't particularly easy to match with a glob if there isn't anything placing it into the popular apps list in the first place. Leaving out of this ticket.

plugins/eos/gs-plugin-eos.c
ramcq: Whether or not there is a .desktop isn't (supposed to be) mutable - the app ID shouldn't change unless there is some weirdness going on. So we don't need to worry about being smarter here, just maintaining this list accurately.

plugins/eos/gs-plugin-eos.c
```diff
@@ -2348,7 +2349,19 @@ gs_plugin_add_popular (GsPlugin *plugin,
...
-		if (g_str_has_prefix (gs_app_get_id (app), "com.endlessm."))
+		const gchar *app_id = gs_app_get_id (app);
+		const gchar *origin = gs_app_get_origin (app);
+		gboolean add = FALSE;
+
+		if (!app_id || !origin)
+			continue;
+
+		/* .endlessm.com apps from eos-apps */
+		if (!g_strcmp0 (origin, "eos-apps") &&
```
wjt: Strong use of „ersatz“ in the commit message!

plugins/eos/gs-plugin-eos.c
```diff
@@ -2333,7 +2349,22 @@ gs_plugin_add_popular (GsPlugin *plugin,
...
 	        GsApp *app = gs_app_list_index (list, i);
-		if (g_str_has_prefix (gs_app_get_id (app), "com.endlessm."))
+		const gchar *app_id = gs_app_get_id (app);
+		const gchar *origin = gs_app_get_origin (app);
+		gboolean add = FALSE;
+
+		if (!app_id || !origin)
+			continue;
+
+		/* .endlessm.com apps from eos-apps */
```
wjt:
```suggestion
		/* com.endlessm. apps from eos-apps */
```

plugins/eos/gs-plugin-eos.c
```diff
@@ -2333,7 +2349,22 @@ gs_plugin_add_popular (GsPlugin *plugin,
...
+		if (!g_strcmp0 (origin, "eos-apps") &&
+		    g_str_has_prefix (app_id, "com.endlessm."))
+			add = TRUE;
+		/* all .endlessnetwork.com apps */
```
wjt:
```suggestion
		/* all com.endlessnetwork. apps */
```

ramcq:
Merging with @wjt's suggested comment fixups applied.

PR #438, opened 2019-01-28
gs-updates-page: Don’t crash if mogwai-scheduled stops

pwithnall:
The updates page holds a reference to the Mogwai scheduler, which should
normally stop it being invalidated while the page needs it. However, if
some feckless developer runs `systemctl stop mogwai-scheduled`, the hold
on the scheduler will be ignored, and it will disappear from the bus.

Previously, this would cause gnome-software to emit a critical warning
and abort. We probably want it to gracefully drop its scheduler instance
and carry on without scheduling support instead.

Spotted while working on T24117.

### DIFF
--- a/src/gs-updates-page.c
+++ b/src/gs-updates-page.c
@@ -1799,9 +1799,11 @@ gs_updates_page_upgrade_cancel_cb (GsUpgradeBanner *upgrade_banner,
 static void
 scheduler_invalidated_cb (GsUpdatesPage *self)
 {
-	g_warning ("The Mogwai Scheduler has just been invalidated when "
-		   "it should still be held!");
-	g_assert_not_reached ();
+	/* The scheduler shouldn’t normally be invalidated, since we Hold() it
+	 * until we’re done with it. However, if the scheduler is stopped by
+	 * systemd (`systemctl stop mogwai-scheduled`) this signal will be
+	 * emitted. */
+	g_clear_object (&self->scheduler);
 }
 
 static void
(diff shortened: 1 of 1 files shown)


src/gs-updates-page.c
```diff
@@ -1799,9 +1799,11 @@ gs_updates_page_upgrade_cancel_cb (GsUpgradeBanner *upgrade_banner,
...
 scheduler_invalidated_cb (GsUpdatesPage *self)
 {
-	g_warning ("The Mogwai Scheduler has just been invalidated when "
-		   "it should still be held!");
-	g_assert_not_reached ();
+	/* The scheduler shouldn’t normally be invalidated, since we Hold() it
+	 * until we’re done with it. However, if the scheduler is stopped by
+	 * systemd (`systemctl stop mogwai-scheduled`) this signal will be
+	 * emitted. */
+	g_clear_pointer (&self->scheduler);
```
GeorgesStavracas: I think we still would want to warn and abort on regular builds... can't you wrap this up in some sort of `#ifdef G_DEBUG` or something like that?

pwithnall: I don’t think we do. While we’ve requested that the scheduler be held open, we can’t guarantee its behaviour. It’s not very robust if one process exiting causes another to abort.

GeorgesStavracas: Okay. I have no other comments on this MR.

pwithnall: As [Jenkins has pointed out](https://ci.endlessm-sf.com/job/obs-gnome-software-pr/330/console), it would work better had I actually passed both arguments to `g_clear_pointer()`.

PR #439, opened 2019-01-28
gs-details-page: Possible fix for an assertion failure warning

pwithnall:
Spotted while looking at OpenQA logs for T24855. This is a potential fix
for the following warning:
  Jan 28 15:22:57 endless gnome-software[1717]: gs_app_is_installed: assertion 'GS_IS_APP (app)' failed

Spotted by reading the code, rather than any more precise debugging
technique. Other places in gs-details-page.c check (self->app != NULL)
before doing anything with the app, so it’s obviously possible for the
app to be NULL. Guard against that case when updating the ‘Copy to USB’
button too.

https://phabricator.endlessm.com/T24855

### DIFF
--- a/src/gs-details-page.c
+++ b/src/gs-details-page.c
@@ -286,7 +286,7 @@ gs_details_page_update_copy_button (GsDetailsPage *self)
 		return;
 
 	copy_dest = removable_destination_dup (self);
-	if (gs_app_is_installed (self->app)) {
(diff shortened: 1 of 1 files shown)


pwithnall:
Note: This is literally just from reading the code. I haven’t even tried to compile it. It’s a drive-by patch.

PR #440, opened 2019-02-06
Improve capitalisation and mnemonics of USB buttons

pwithnall:
Untested, reckless, drive-by patch spotted while working on OpenQA things.

### DIFF
--- a/src/gs-details-page.c
+++ b/src/gs-details-page.c
@@ -310,7 +310,7 @@ gs_details_page_update_copy_button (GsDetailsPage *self)
 				gs_details_page_app_get_copyable_cb,
 				self);
 		} else {
-			gtk_button_set_label (GTK_BUTTON (self->button_copy), _("Insert USB to copy"));
+			gtk_button_set_label (GTK_BUTTON (self->button_copy), _("Insert USB to Copy To"));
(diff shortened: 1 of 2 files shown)


src/gs-details-page.c
```diff
@@ -310,7 +310,7 @@ gs_details_page_update_copy_button (GsDetailsPage *self)
 				gs_details_page_app_get_copyable_cb,
 				self);
 		} else {
-			gtk_button_set_label (GTK_BUTTON (self->button_copy), _("Insert USB to copy"));
+			gtk_button_set_label (GTK_BUTTON (self->button_copy), _("Insert USB to Copy To"));
```
wjt: Both the old and new strings are kind of hawkward to read. I am explicitly not advocating “Insert USB to which to Copy”, but if we must go with “insert usb to copy to”, can we titlecase “to” consistently?

pwithnall: I agree that it’s an awkward string; this is the best I could come up with. The idea with the capitalisation was that the first ‘to’ is a connective; the second is part of the verb ‘copy to’. In the style guide (IIRC), connectives are lower case, verbs are Title Case.

wjt: OK, so there is method to it! I can't think of anything better either ☹

PR #443, opened 2019-02-21
flatpak: Upgrade related apps to proxied apps

joaquimrocha:
When an app is made updatable because one of its related apps has an
update, but the (main) app itself is proxied by another app, then the
update is only attempted on the proxied app, i.e. related apps of
proxied apps never get updated.

This was happening in the case of the ClippyExtension, since this is an
extension of the HackComponents app, so if the ClippyExtension has an
update, the HackComponents app is set as updatable, which by turn makes
the Hack proxy app updatable; however, GNOME Software just attempted to
update the HackComponents app individually, thus, the ClippyExtension
could never be updated.

This patch fixes that by using the gathering the list needed for an app
update of a proxied app that is also used in regular (non-proxied) apps
which means it'll contain all its related apps + the app itself, thus
solving the aforementioned issue.

Subsequent patches arrange for Hack apps to be installed if they are missing.

https://phabricator.endlessm.com/T25498


joaquimrocha:
@cosimoc Yup, it solved it for me when I downgraded ClippyExtension in order to get the issue (command in the Phab task).

plugins/flatpak/gs-flatpak.c
```diff
@@ -4068,8 +4068,17 @@ gs_flatpak_update_app (GsFlatpak *self,
...
-			gs_app_list_add (list, g_ptr_array_index (proxied_apps, i));
+		for (guint i = 0; i < proxied_apps->len; ++i) {
+			g_autoptr(GsAppList) app_related_list = NULL;
+			GsApp *proxied_app = g_ptr_array_index (proxied_apps, i);
+
+			/* Add not only the app but also any eventual related apps to be
+			 * updated; otherwise, the related apps of proxied apps may never
+			 * get updated. */
+			app_related_list = gs_flatpak_get_list_for_update (self, proxied_app,
+									   cancellable, error);
```
wjt: You are not handling the case where this call fails and `error` is set.

joaquimrocha: Rushy rushy! I will address it.

joaquimrocha:
Hey, I have updated this PR with the fix to the issue that @wjt found, and the "install if missing" feature for the Proxy apps.

lib/gs-app.c
```diff
@@ -3328,6 +3328,29 @@ gs_app_add_related (GsApp *app, GsApp *app2)
 	g_ptr_array_add (priv->related, g_object_ref (app2));
 }
 
+/**
+ * gs_app_clear_related:
+ * @app: a #GsApp
+ * @app2: a #GsApp
```
cosimoc: There's no `app2` in this method

plugins/flatpak/gs-flatpak.c
```diff
@@ -4070,13 +4070,24 @@ gs_flatpak_update_app (GsFlatpak *self,
...
 			 * updated; otherwise, the related apps of proxied apps may never
 			 * get updated. */
 			app_related_list = gs_flatpak_get_list_for_update (self, proxied_app,
-									   cancellable, error);
+									   cancellable,
+									   &local_error);
+			if (app_related_list == NULL) {
+				g_warning ("Error getting the list for updating the proxied "
+					   "app %s: %s; adding only the main app itself.",
+					   gs_app_get_unique_id(proxied_app),
```
cosimoc: Missing space before the paren

plugins/flatpak/gs-flatpak.c
```diff
@@ -4133,7 +4133,7 @@ gs_flatpak_update_app (GsFlatpak *self,
 
 		/* either install or update the ref */
 		ref_display = gs_flatpak_app_get_ref_display (app_tmp);
-		if (!is_proxy_app && !g_hash_table_contains (hash_installed, ref_display)) {
```
cosimoc: Could it have been that the intention here was to check whether `app_tmp` was a proxy app instead? That would make more sense.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2577,6 +2577,43 @@ gs_flatpak_refine_appstream (GsFlatpak *self, GsApp *app, GError **error)
...
+
+	real_related_apps = gs_app_list_new ();
+
+	/* get a list of the real apps (not eventual wildcard ones) */
+	for (guint i = 0; i < proxied_apps->len; ++i) {
+		GsApp *app_cached = NULL;
+		GsApp *proxied_app = g_ptr_array_index (proxied_apps, i);
+
+		if (gs_app_has_quirk (proxied_app, GS_APP_QUIRK_IS_WILDCARD) &&
+		    gs_app_get_bundle_kind (proxied_app) == AS_BUNDLE_KIND_FLATPAK) {
```
cosimoc: Could we even get here with something that is not a flatpak?

plugins/flatpak/gs-flatpak.c
```diff
@@ -2577,6 +2577,43 @@ gs_flatpak_refine_appstream (GsFlatpak *self, GsApp *app, GError **error)
...
+			const gchar *id = gs_app_get_unique_id (proxied_app);
+			app_cached = gs_plugin_cache_lookup (self->plugin, id);
+			if (app_cached == NULL)
+				continue;
+		}
+
+		gs_app_list_add (real_related_apps, app_cached);
```
cosimoc: Does this line not belong inside the block above? Otherwise you can end up adding `NULL` to the list. (Or alternatively the `app_cached == NULL` check could be moved outside of the block.

plugins/eos/gs-plugin-eos.c
```diff
@@ -2068,31 +2107,59 @@ static void
...
 		       GsAppList *list,
 		       GsApp *proxy_app,
+		       gboolean install_missing,
 		       const char **proxied_apps)
 {
-	g_autoptr(GSList) proxied_updates = NULL;
+	g_autoptr(GsAppList) proxied_updates = gs_app_list_new ();
+
+	GsApp *cached_proxy_app = gs_app_list_lookup (list,
+						      gs_app_get_unique_id(proxy_app));
```
cosimoc: Missing space before paren

plugins/eos/gs-plugin-eos.c
```diff
@@ -2068,31 +2107,59 @@ static void
...
+	if (cached_proxy_app != NULL)
+		gs_app_list_remove (list, cached_proxy_app);
+
+	/* add all the apps if we should install the missing ones; the real sorting for
+	 * whether they're already installed is done in the refine method */
+	if (install_missing) {
+		for (guint i = 0; proxied_apps[i] != NULL; ++i) {
+		        g_autoptr(GsApp) app = gs_app_new (proxied_apps[i]);
+			gs_app_add_quirk (app, GS_APP_QUIRK_IS_WILDCARD);
+			gs_app_set_bundle_kind(app, AS_BUNDLE_KIND_FLATPAK);
```
cosimoc: Missing space before paren

plugins/eos/gs-plugin-eos.c
```diff
@@ -2068,31 +2107,59 @@ static void
...
 		GsApp *app = gs_app_list_index (list, i);
+		GsApp *added_app = NULL;
 		const char *id = gs_app_get_id (app);
 
 		if (!g_strv_contains (proxied_apps, id) ||
 		    gs_app_get_scope (proxy_app) != gs_app_get_scope (app))
 			continue;
 
-		proxied_updates = g_slist_prepend (proxied_updates, app);
+		added_app = gs_app_list_lookup (proxied_updates, gs_app_get_unique_id(app));
```
cosimoc: Missing space before paren

plugins/eos/gs-plugin-eos.c
```diff
@@ -2068,31 +2107,59 @@ static void
...
+		if (added_app == app)
+			continue;
+
+		gs_app_list_remove (proxied_updates, added_app);
+
+		gs_app_set_state (app, AS_APP_STATE_UPDATABLE_LIVE);
+		gs_app_list_add (proxied_updates, app);
 	}
 
 	if (!proxied_updates)
```
cosimoc: This will now never be true; you may want to check the length of the list instead?

plugins/eos/gs-plugin-eos.c
```diff
@@ -256,6 +257,21 @@ get_personality (void)
...
 
+static char *
+get_product_name (const char *image_version)
+{
+	char *hyphen_index = NULL;
+
+	if (image_version == NULL)
+		return NULL;
+
+	hyphen_index = index (image_version, '-');
```
cosimoc: `index()` is deprecated in favor of the equivalent `strchr()`

plugins/eos/gs-plugin-eos.c
```diff
@@ -108,6 +109,7 @@ struct GsPluginData
 	int applications_changed_id;
 	SoupSession *soup_session;
 	char *personality;
+	char *product_name;
```
cosimoc: Should be freed in the destroy func

plugins/flatpak/gs-flatpak.c
joaquimrocha: I think so, but not because there may be a proxy app related to others (and this is not what was being done).

plugins/flatpak/gs-flatpak.c
joaquimrocha: Actually we could! I think it could be prevented if we added something in https://github.com/endlessm/gnome-software/blob/master/plugins/flatpak/gs-plugin-flatpak.c#L343
However, maybe the idea was also to have (outside of EOS) proxy apps matching apps in different systems. So let's keep it here for now.

plugins/flatpak/gs-flatpak.c
joaquimrocha: Good catch!

plugins/eos/gs-plugin-eos.c
joaquimrocha: Great catch!

plugins/eos/gs-plugin-eos.c
```diff
@@ -1772,6 +1772,40 @@ gs_plugin_eos_refine_popular_app (GsPlugin *plugin,
...
+
+	for (guint i = 0; i < proxied_apps->len; ++i) {
+		GsApp *proxied_app = g_ptr_array_index (proxied_apps, i);
+
+		if (gs_app_get_state (proxied_app) == AS_APP_STATE_AVAILABLE ||
+		    gs_app_is_updatable (proxied_app)) {
+			g_debug ("App %s has an update or needs to be installed/updated "
+				 "by the proxy app %s",
+				 gs_app_get_unique_id(proxied_app),
+				 gs_app_get_unique_id(app));
```
cosimoc: For these two lines, there's a missing space before paren

joaquimrocha:
Hi Cosimo, thanks for the review. The tests in g-s for EOS needed a lot of love last time I checked. There's this open task [here](https://phabricator.endlessm.com/T21809) and I have added a notice about doing it for Proxy apps too. I will address your last comment in this PR and squash the commits.

joaquimrocha:
Merging it myself after addressing @cosimoc 's last comment.

ramcq:
:thinking: I think this passed the bar of non-trivial change that required a review from the Desktop team before merge. As in, Will asked the team to review it yesterday...

plugins/flatpak/gs-flatpak.c
```diff
@@ -2577,6 +2577,43 @@ gs_flatpak_refine_appstream (GsFlatpak *self, GsApp *app, GError **error)
...
+	}
+
+	/* set the cached apps as the related ones instead */
+	gs_app_clear_related (app);
+
+	for (guint i = 0; i < gs_app_list_length (real_related_apps); ++i) {
```
pwithnall: Looks like this loop could be folded into the one above.

plugins/eos/gs-plugin-eos.c
```diff
@@ -1772,6 +1772,40 @@ gs_plugin_eos_refine_popular_app (GsPlugin *plugin,
...
+	}
+
+	gs_app_clear_related (app);
+
+	for (guint i = 0; i < gs_app_list_length (related_filtered); ++i) {
+		GsApp *related_app = gs_app_list_index (related_filtered, i);
+		gs_app_add_related (app, related_app);
+	}
+
+	gs_app_set_state (app, AS_APP_STATE_UNKNOWN);
```
pwithnall: If `proxied_apps->len == 0`, then the function will mark the app as `STATE_UNKNOWN` unnecessarily (and will otherwise be a no-op). Is that correct?

plugins/eos/gs-plugin-eos.c
```diff
@@ -2068,31 +2107,59 @@ static void
...
 	for (guint i = 0; i < gs_app_list_length (list); ++i) {
 		GsApp *app = gs_app_list_index (list, i);
+		GsApp *added_app = NULL;
 		const char *id = gs_app_get_id (app);
 
 		if (!g_strv_contains (proxied_apps, id) ||
 		    gs_app_get_scope (proxy_app) != gs_app_get_scope (app))
 			continue;
 
-		proxied_updates = g_slist_prepend (proxied_updates, app);
```
pwithnall: The behaviour of `proxied_updates` changes in the case where `!install_missing`. From the commit message, I would have assumed that the behaviour should only change when `install_missing` is `TRUE`. Is that not the case?

plugins/eos/gs-plugin-eos.c
```diff
@@ -2205,13 +2224,16 @@ add_updates (GsPlugin *plugin,
...
+	gboolean is_hack_product = g_strcmp0 (priv->product_name, "hack") == 0;
+
 	process_proxy_updates (plugin, list,
 			       framework_proxy_app,
 			       FALSE,
 			       framework_proxied_apps);
 	process_proxy_updates (plugin, list,
 			       hack_proxy_app,
-			       FALSE,
+			       is_hack_product,
```
pwithnall: Do we really want the behaviour of gnome-software to differ based on the OS? That doubles the testing space. Could this just be based on whether the Hack proxy app is installed in the first place?

plugins/flatpak/gs-flatpak.c
joaquimrocha: Hmm, we're clearing the related apps of the app in the call above (meaning it does clear the object it returns). So for me to fold this into the loop above I'd have to copy those apps elsewhere.

plugins/eos/gs-plugin-eos.c
joaquimrocha: This is fine in any case and deliberate, as an unknown state ensures the proxy app will never be shown anywhere.

plugins/eos/gs-plugin-eos.c
joaquimrocha: It doesn't really change, unless I am missing some issue. This code simply checks if there's already an app matching the one we're about to add (in order to remove it first). In case `!install_missing`, the list we're doing the lookup on is empty, so it will always add any updatable apps that are in the `proxied_apps` list. Thus, the code changes, but not the functionality.

plugins/eos/gs-plugin-eos.c
joaquimrocha: Proxy apps aren't really "installed", they're created and shown as updatable just for masking the updates. So their existence is ephemeral.

It's also not trivial to understand whether an app is installed or not in this place (hence the amount of patches just for making sure that we don't show the Hack proxy app if all its apps are already installed and up-to-date).

joaquimrocha:
@ramcq / @pwithnall , I am sorry I may have not given Desktop time for a review after Cosimo's last comments. I assumed that it since the PR had been in CR Ready for a day (hopefully giving the chance to those tagged in it), and reviewed by Cosimo, it'd be fine to merge it. But I agree that someone from the Desktop team should have taken a final look.
I will keep that in mind next time.

plugins/flatpak/gs-flatpak.c
pwithnall: Or move the `gs_app_clear_related (app)` call above the first loop?

plugins/eos/gs-plugin-eos.c
pwithnall: Maybe that could be added in a comment above the `gs_app_set_state()` call.

plugins/eos/gs-plugin-eos.c
pwithnall: The app gets its state changed with `gs_app_set_state (app, AS_APP_STATE_UPDATABLE_LIVE);` now; the code didn’t previously do that. Is that OK?

plugins/eos/gs-plugin-eos.c
pwithnall: What would happen if `TRUE` was unconditionally passed to this function on both versions of the OS, rather than passing `is_hack_product`?

pwithnall:
> I assumed that it since the PR had been in CR Ready for a day

Sorry, my work queue is often longer than a day so I didn’t get to this in time. Pinging someone from the desktop team on Slack about whether it’s OK to merge something like this if you haven’t heard from us would be useful.

plugins/flatpak/gs-flatpak.c
joaquimrocha: That means we wouldn't be able to loop through the `GPtrArray` we assigned before because it'll be cleared in the `gs_app_clear_related` call.

plugins/eos/gs-plugin-eos.c
joaquimrocha: I will add that now.

plugins/eos/gs-plugin-eos.c
joaquimrocha: The app is already created with that state because they'd only be added if there were updates anyway. Now there are more moving parts, so if anything this code ensures the state is the desired one (so it doesn't affect the state differently than before for the needed outcome).

plugins/eos/gs-plugin-eos.c
joaquimrocha: @pwithnall all apps belonging to either proxy apps would end up installed.

plugins/eos/gs-plugin-eos.c
pwithnall: Ideally, there could be a comment in the code about this. :+1:

PR #444, opened 2019-02-21
eos: update list of proxied apps for the HackComponents

erikos:
Add the future HackAccountService to the list of proxied apps for the HackComponents.

https://phabricator.endlessm.com/T25480


plugins/eos/gs-plugin-eos.c
```diff
@@ -2127,6 +2127,7 @@ add_updates (GsPlugin *plugin,
 	const char *hack_proxied_apps[] = {"com.endlessm.Clubhouse.desktop",
 					   "com.endlessm.Fizzics.desktop",
 					   "com.endlessm.GameStateService.desktop",
+					   "com.endlessm.HackAccountService.desktop",
```
wjt: What's the behaviour if (as is true now) this app just doesn't exist?

joaquimrocha: Nothing happens. That list is used to match any already updatable apps with it. So if there's a string that doesn't get matched, it's not a problem.

PR #445, opened 2019-02-21
add Learn to Code category

treitter:
This category will be used for games developed by Endless Interactive to
highlight them separately from other games.

As of this writing, there are no apps available in this category so it
will not be displayed. However, I've confirmed that the category does
get populated when apps for the associated category are available.

For apps to appear in this category, they will need to include "Endless
Games" in their AppStream <categories>. See the AppStream documentation
for more details:

https://www.freedesktop.org/software/appstream/docs/chap-CollectionData.html

This, of course, is an invented category not among the FreeDesktop Menu
Specification's Registered Categories (but it's also not among the
Reserved Categories):

https://specifications.freedesktop.org/menu-spec/latest/apa.html

https://phabricator.endlessm.com/T25518

### DIFF
--- a/plugins/core/gs-desktop-common.c
+++ b/plugins/core/gs-desktop-common.c
@@ -169,6 +169,13 @@ static const GsDesktopMap map_games[] = {
 	{ NULL }
 };
 
+static const GsDesktopMap map_learn_to_code[] = {
+	{ "all",		NC_("Menu of Learn to Code", "All"),
+					{ "LearnToCode",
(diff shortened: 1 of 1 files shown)


treitter:
Currently, I'm just using the generic "Game" category icon so it will be identical to the adjacent "Games" category. Endless Network is fine with using that as a placeholder for now, so we don't need to block on that but we should have a final version by mid next week or so.

PR #447, opened 2019-03-01
eos: Add a comment clarifying the state change of proxy apps

joaquimrocha:
This state change happens in the function that refines proxy apps, and
ensures that the proxy apps' state is set according to whether it has
apps to be processed; however, especially for developers not very
experienced with GNOME Software, it's good to document why that change
is done the way it is.

### DIFF
--- a/plugins/eos/gs-plugin-eos.c
+++ b/plugins/eos/gs-plugin-eos.c
@@ -1818,6 +1818,9 @@ gs_plugin_refine_proxy_app (GsPlugin	*plugin,
 		gs_app_add_related (app, related_app);
 	}
 
+	/* mark the state as unknown so 1) we're always allowed to change the state below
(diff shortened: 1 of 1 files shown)


pwithnall:
Thanks a lot. This change looks good to me, but it would also be nice to see a comment [for this bit of the diff too](https://github.com/endlessm/gnome-software/pull/443#discussion_r262063746).

joaquimrocha:
> Thanks a lot. This change looks good to me, but it would also be nice to see a comment [for this bit of the diff too](https://github.com/endlessm/gnome-software/pull/443#discussion_r262063746).

There you go (sorry for having gotten lost so badly in the review/comments): https://github.com/endlessm/gnome-software/pull/448

pwithnall:
Thanks! Both merged.

PR #449, opened 2019-03-25
T25495 fix pending updates

joaquimrocha:

https://phabricator.endlessm.com/T26042

### DIFF
--- a/plugins/flatpak/gs-flatpak.c
+++ b/plugins/flatpak/gs-flatpak.c
@@ -1665,9 +1665,14 @@ gs_flatpak_add_updates_pending (GsFlatpak *self, GsAppList *list,
 
 		update_priority = get_app_update_priority (app);
 
-		main_app = get_real_app_for_update (self, app, cancellable, error);
-		if (main_app == NULL)
-			return FALSE;
+		main_app = get_real_app_for_update (self, app, cancellable, &error_local);
+		/* do not fail if one of the updates cannot be done (otherwise the update
+		 * list will be empty), just inform the user */
+		if (main_app == NULL) {
+			g_warning ("Failed to get real app when updating %s: %s",
+				   gs_app_get_unique_id (app), error_local->message);
+			continue;
+		}
 
 		/* set the update highest update priority between the main app
 		 * and any important related apps (e.g. if we have a content
--- a/plugins/flatpak/gs-flatpak.c
+++ b/plugins/flatpak/gs-flatpak.c
@@ -1418,6 +1418,7 @@ get_main_app_of_related (GsFlatpak *self,
 	g_autoptr(FlatpakInstalledRef) ref = NULL;
 	const gchar *ref_name;
 	g_auto(GStrv) app_tokens = NULL;
+	FlatpakRefKind ref_kind = FLATPAK_REF_KIND_RUNTIME;
 
 	ref_name = gs_flatpak_app_get_main_app_ref_name (related_app);
 	if (ref_name == NULL) {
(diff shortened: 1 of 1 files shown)


plugins/flatpak/gs-flatpak.c
```diff
@@ -1665,9 +1665,18 @@ gs_flatpak_add_updates_pending (GsFlatpak *self, GsAppList *list,
 
 		update_priority = get_app_update_priority (app);
 
-		main_app = get_real_app_for_update (self, app, cancellable, error);
-		if (main_app == NULL)
-			return FALSE;
+		main_app = get_real_app_for_update (self, app, cancellable, &error_local);
```
cosimoc: In the commit messages there are a few typos:
- `it whole` -> `the whole`
- `updage` -> `update`

plugins/flatpak/gs-flatpak.c
```diff
@@ -1665,9 +1665,18 @@ gs_flatpak_add_updates_pending (GsFlatpak *self, GsAppList *list,
...
+		/* do not fail if one of the updates cannot be done (otherwise the update
+		 * list will be empty), just inform the user */
+		if (main_app == NULL) {
+			g_autoptr(GsPluginEvent) event = gs_plugin_event_new ();
```
cosimoc: It's not entirely clear to me the effect that this will have to the user; will it create a in-app notification? I want to avoid a gibberish/technical error to pop up to the user if possible.

plugins/flatpak/gs-flatpak.c
joaquimrocha: It was late. Sorry...

plugins/flatpak/gs-flatpak.c
joaquimrocha: Yes. It's an in-app notification. I did it like that because it matches the approach in the `add_updates` function (used for deploying pre-downloaded updates) which is not usually called in our case.
I can remove this event and let it continue silently (maybe with a warning) if that's better UX.

joaquimrocha:
I have updated the PR and dropped the in-app notification if an error occurs (instead, it logs the issue so we can get it in the journal if needed).

plugins/flatpak/gs-flatpak.c
```diff
@@ -1665,9 +1665,14 @@ gs_flatpak_add_updates_pending (GsFlatpak *self, GsAppList *list,
...
-		main_app = get_real_app_for_update (self, app, cancellable, error);
-		if (main_app == NULL)
-			return FALSE;
+		main_app = get_real_app_for_update (self, app, cancellable, &error_local);
+		/* do not fail if one of the updates cannot be done (otherwise the update
+		 * list will be empty), just inform the user */
+		if (main_app == NULL) {
+			g_warning ("Failed to get real app when updating %s: %s",
+				   gs_app_get_unique_id (app), error_local->message);
+			continue;
```
pwithnall: What if all the `get_real_app_for_update()` calls fail (for all the `xrefs`), and `list` ends up empty that way? Should `gs_flatpak_add_updates_pending()` return an error in that case?

joaquimrocha: I pondered exactly that, since I think we take a similar approach elsewhere in g-s IIRC. However, due to the limited time for testing, the fact that we do get the messages in the log (in case we need a detailed report from users), and that this code will be modified by the rebase, I think we should proceed without that.
Does this make sense?

pwithnall: > However, due to the limited time for testing, the fact that we do get the messages in the log (in case we need a detailed report from users), and that this code will be modified by the rebase, I think we should proceed without that.
> Does this make sense?

IMO it’s less a case of whether we get messages in the log (though that is good), and more a case of whether the calling code expects an empty list to ever be returned if the function doesn’t set an error.

joaquimrocha: That is okay AFAICS because it will just assume there was no updates at all, and the same could happen before when there were no updates to be done (the list is passed as parameter BTW, and it's okay to not touch it).

pwithnall: OK.

PR #450, opened 2019-03-27
Add and use “Learn to Code” category icon

wjt:
This is much better than using the same icon as the adjacent "Games" category.

https://phabricator.endlessm.com/T25518


wjt:
Screenshot from 2019-03-27 22-23-59

wjt:
Updated with a better icon.

Screenshot from 2019-03-28 18-34-34

PR #451, opened 2019-04-09
T26187 Fix crash when a USB drive is inserted

mwleeds:
https://phabricator.endlessm.com/T26187

### DIFF
--- a/lib/gs-category.c
+++ b/lib/gs-category.c
@@ -525,6 +525,30 @@ gs_category_sort_children (GsCategory *category)
 			  gs_category_sort_children_cb);
 }
 
+/**
+ * gs_category_equal:
+ * @a: a #GsCategory
+ * @b: a #GsCategory
+ *
+ * Returns: %TRUE if @a and @b have the same category id, and %FALSE otherwise.
+ **/
+gboolean
+gs_category_equal (GsCategory *a,
+		   GsCategory *b)
+{
+	const char *a_id, *b_id;
+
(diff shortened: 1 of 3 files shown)


lib/gs-category.c
```diff
@@ -525,6 +525,27 @@ gs_category_sort_children (GsCategory *category)
...
+ **/
+gboolean
+gs_category_equal (GsCategory *a,
+		   GsCategory *b)
+{
+	const char *a_id, *b_id;
+
+	g_return_val_if_fail (GS_IS_CATEGORY (a), FALSE);
+	g_return_val_if_fail (GS_IS_CATEGORY (b), FALSE);
+
```
cosimoc: I think it would be worth adding the pointer comparison here as well as a small optimization for when the category object did not actually change.

mwleeds: Pushed a fixup

lib/gs-category.c
```diff
@@ -536,14 +536,17 @@ gboolean
...
-
 	g_return_val_if_fail (GS_IS_CATEGORY (a), FALSE);
 	g_return_val_if_fail (GS_IS_CATEGORY (b), FALSE);
 
-	a_id = gs_category_get_id (a);
-	b_id = gs_category_get_id (b);
-	return g_strcmp0 (a_id, b_id) == 0;
+	if (a == b) {
+		return TRUE;
+	} else {
```
joaquimrocha: You don't need, and shouldn't use the else block here (since you return and never reach it otherwise). It's also a common use in g-s, so IMO we should do it.

PR #455, opened 2019-05-30
T26511 Split out eos-blacklist plugin from eos plugin

pwithnall:
This splits the blacklist code out of the [`eos` plugin](https://github.com/endlessm/gnome-software/tree/master/plugins/eos) and drops a few bits of it which seem outdated. See the commit messages for rationale, and also see [T26511](https://phabricator.endlessm.com/T26511).

I’ve based this PR on the new `T25367-rebase-in-progress` branch so that only the commits which are relevant to the blacklist are reviewed here. I plan to build up the rebase-in-progress branch with bits of the actual rebase as they get reviewed and merged.

@joaquimrocha, @ramcq, I’ve tagged you two for review because you’ve both had a lot of dealings with these blacklists in the past and can tell me whether some of them still need to stay (or whether more of them can be dropped) than I can.

For context, there’s a big comment in T26511 which explains all the blacklists, what I’ve done with them in this rebase, and what I think should eventually be done with them.

https://phabricator.endlessm.com/T26511

### DIFF
--- a/plugins/eos-blacklist/gs-plugin-eos-blacklist.c
+++ b/plugins/eos-blacklist/gs-plugin-eos-blacklist.c
@@ -198,17 +198,6 @@ app_get_flatpak_ref_name (GsApp *app)
 }
 
 static gboolean
-app_is_renamed (GsApp *app)
-{
-	/* Apps renamed by eos-desktop get the desktop attribute of
-	 * X-Endless-CreatedBy assigned to the desktop's name;
-	 * Starting with EOS 3.2 apps can no longer be renamed so
-	 * we keep it for legacy reasons */
-	return g_strcmp0 (gs_app_get_metadata_item (app, "X-Endless-CreatedBy"),
-			  "eos-desktop") == 0;
-}
-
-static gboolean
 gs_plugin_locale_is_compatible (GsPlugin *plugin,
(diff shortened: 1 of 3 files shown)


plugins/eos-blacklist/gs-plugin-eos-blacklist.c
```diff
@@ -628,13 +628,6 @@ gs_plugin_eos_blacklist_app_for_remote_if_needed (GsPlugin *plugin,
 		NULL
 	};
 
-	/* Legacy apps that have been replaced by other versions in Flathub */
-	static const char *legacy_apps[] = {
-		"com.spotify.Client",
```
ramcq: Agree with this change; but for posterity/context - the reason these were blacklisted in the reverse (from Endless rather than Flathub) is that unlike the other apps migrated from eos-apps to flathub, we kept these ones alive in eos-apps so that OS images which had "Get App" launchers / autoinstalls would still be able to fetch them. We later decided to just deal with that regression on older images because maintaining the parallel copies of them inside our infrastructure was sufficiently painful (especially VLC, a very fragile extra-data patchwork quilt of .debs) and somewhat voided the purpose of moving them to Flathub.

PR #456, opened 2019-05-31
eos: Drop popular apps support

pwithnall:
It’s now implemented by adding the relevant kudos to the eos-extra.xml
appstream file, which can be updated independently from gnome-software.

### DIFF
--- a/plugins/eos/gs-plugin-eos.c
+++ b/plugins/eos/gs-plugin-eos.c
@@ -2079,65 +2079,6 @@ gs_plugin_add_updates (GsPlugin *plugin,
 	return add_updates (plugin, list, cancellable, error);
 }
 
-gboolean
-gs_plugin_add_popular (GsPlugin *plugin,
-		       GsAppList *list,
-		       GCancellable *cancellable,
-		       GError **error)
-{
-	g_autoptr(AsProfileTask) ptask = NULL;
-	GsAppList *new_list = NULL;
-	const gchar *popular_apps[] = {
-		"com.calibre_ebook.calibre.desktop",
-		"com.google.Chrome.desktop",
-		"com.spotify.Client.desktop",
-		"com.transmissionbt.Transmission.desktop",
-		"com.valvesoftware.Steam.desktop",
-		"io.github.mmstick.FontFinder.desktop",
-		"net.minetest.Minetest.desktop",
-		"net.scribus.Scribus.desktop",
-		"org.audacityteam.Audacity.desktop",
-		"org.gimp.GIMP.desktop",
-		"org.gnome.chess.desktop",
-		"org.gnome.SwellFoop.desktop",
-		"org.inkscape.Inkscape.desktop",
-		"org.kde.gcompris.desktop",
-		"org.kde.krita.desktop",
-		"org.libreoffice.LibreOffice.desktop",
-		"org.mozilla.Firefox.desktop",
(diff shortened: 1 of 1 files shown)


plugins/eos/gs-plugin-eos.c
```diff
@@ -2079,65 +2079,6 @@ gs_plugin_add_updates (GsPlugin *plugin,
...
-	for (guint i = 0; popular_apps[i] != NULL; ++i) {
-		g_autoptr(GsApp) app = gs_app_new (popular_apps[i]);
-		gs_app_add_quirk (app, AS_APP_QUIRK_MATCH_ANY_PREFIX);
-		gs_app_list_add (new_list, app);
-	}
-
-	/* get all the popular apps that are Endless' ones */
-	for (guint i = 0; i < gs_app_list_length (list); ++i) {
-	        GsApp *app = gs_app_list_index (list, i);
-		if (g_str_has_prefix (gs_app_get_id (app), "com.endlessm."))
```
wjt: https://github.com/endlessm/gnome-software-data/pull/10 has superseded the loop above, but it hasn't superseded this logic. Do some/all of our own-brand apps' own appdata have `GnomeSoftware::popular`, and is it respected? Or is the intention that we'd put some curated set of apps into eos-extra before this change is released? (If so, let's add a task for that.) I had some memory that it's ignored from some sources, but I can't find anything implementing that in a cursory search.

pwithnall: So I actually missed the `com.endlessm.` prefix check when I was working on this, so it got omitted from my thinking entirely.

The machinery in `eos-build` which generates the appdata for (some of?) our apps used to add the kudo directly into the appdata, but that isn’t allowed by `appstream-validate` any more, so it [no longer does](https://github.com/endlessm/eos-build/commit/05977432eb2a624a436caedbcb92f9acd41581de).

We’d either have to list all the `com.endlessm.`-prefixed apps (or some subset of them) in `eos-extra.xml`, or drop part of this PR. Since time is tight, I suggest:
 * Reworking this PR so the prefix check stays in gnome-software, but the hard-coded popular apps list is dropped.
 * The hard-coded popular apps list is added to `eos-extra.xml` (already done in https://github.com/endlessm/gnome-software-data/pull/10).
 * A follow-up task is filed to rethink this, and potentially dynamically generate `eos-extra.xml` to list all `com.endlessm.` apps for the kudo.

The reason I’m not recommending adding the kudo for all `com.endlessm.` apps to `eos-extra.xml` now is because there’s currently no build machinery for `eos-extra.xml`, so that would have to be added.

pwithnall: Actually, re-reading this again and again, this code doesn’t add any Endless apps to the popular list which aren’t already there. All it does is filter the existing popular list to remove the non-Endless apps (and not remove the hard-coded popular apps it’s just added). So in order for this to function as it currently does, we only need to un-tag non-Endless apps as popular. They’re likely coming from the hard-coded popular plugin, which I’ll disable in the build.

tl;dr: I think we can actually drop all this code. Any change in behaviour can be tweaked later by changing which apps are marked as popular in the metadata.

pwithnall: Dropped by dropping the “eos: Set up the list of popular apps directly” commit which originally introduced this code, as part of the rebase in `T25367-new-downstream-master2`.

PR #460, opened 2019-06-13
eos: rename org.gnome.chess to org.gnome.Chess

wjt:
This was renamed upstream. I have not verified that it is still unusable
after this update.

https://phabricator.endlessm.com/T27018


wjt:
Maddening.

– Will

PR #463, opened 2019-07-08
T27182 flatpak dropped changes

pwithnall:
Various commits to reinstate features from
https://phabricator.endlessm.com/T27182

This is part of the [gnome-software 3.32 rebase](https://phabricator.endlessm.com/T25367).

### DIFF
--- a/plugins/core/gs-appstream.c
+++ b/plugins/core/gs-appstream.c
@@ -1048,6 +1048,32 @@ gs_appstream_search (GsPlugin *plugin,
 		}
 	}
 
+	/* If the caller has provided the custom "Endless::HasDiscoveryFeedContent"
+	 * search string (which is squashed to lower case by the search machinery),
+	 * add a search for it in the custom metadata dictionary of each component.
+	 *
+	 * FIXME: Technically we should also check ` and text()='true'` in the
+	 * XPath selector, but libxmlb doesn’t support that. So this will return
+	 * all components which specify
+	 *    <custom><value key="Endless::HasDiscoveryFeedContent">false</value></custom>
+	 * too. */
+	if (g_strv_contains (values, "endless::hasdiscoveryfeedcontent")) {
+		g_autoptr(GError) error_query = NULL;
+
+		/* AppStream versions 0.9 and above call it <custom>; older
+		 * versions call it <metadata>, but thankfully the upgrade is
+		 * handled already using gs_plugin_appstream_upgrade_cb(). */
+		g_autoptr(XbQuery) query = xb_query_new (silo, "custom/value[@key='Endless::HasDiscoveryFeedContent']", &error_query);
+		if (query != NULL) {
+			GsAppstreamSearchHelper *helper = g_new0 (GsAppstreamSearchHelper, 1);
+			helper->match_value = AS_APP_SEARCH_MATCH_KEYWORD;
+			helper->query = g_steal_pointer (&query);
+			g_ptr_array_add (array, g_steal_pointer (&helper));
+		} else {
+			g_debug ("ignoring: %s", error_query->message);
+		}
+	}
+
 	/* get all components */
 	components = xb_silo_query (silo, "components/component", 0, &error_local);
 	if (components == NULL) {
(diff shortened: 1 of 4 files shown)


src/gs-loading-page.c
```diff
@@ -114,9 +114,23 @@ gs_loading_page_load (GsLoadingPage *self)
...
+	settings = g_settings_new ("org.gnome.software");
+	if (g_settings_get_boolean (settings, "first-run"))
+		cache_age = 60 * 60 * 24;  /* 24 hours */
+	else
+		cache_age = G_MAXUINT;
 
-	/* ensure that at least some metadata of any age is present, and also
-	 * spin up the plugins enough as to prime caches */
 	plugin_job = gs_plugin_job_newv (GS_PLUGIN_ACTION_REFRESH,
 					 "age", (guint64) G_MAXUINT,
```
uajain: You set a new `cache_age` variable above but I think you forgot to use that variable here.

uajain:
I eye-rolled these patches for quite sometime, and seems broadly LGTM to me. 
There are major new contexts attached to two of the commits. I did go deep enough (not too deep) to understand the context of these patches.

If I can help with some testing, I would have been more confident about these patches (and its behavior). But I think @pwithnall has probably done testing around it to see if they work.

plugins/eos/gs-plugin-eos.c
```diff
@@ -488,6 +488,71 @@ gs_plugin_eos_refine_popular_app (GsPlugin *plugin,
...
+	g_autoptr(GsApp) services_app = NULL;
+
+	runtime = gs_app_get_runtime (app);
+	if (runtime == NULL)
+		return;
+
+	/* add services flatpak */
+	services_app = gs_flatpak_get_services_app_if_needed (plugin, app, runtime, cancellable);
+	if (services_app != NULL)
+		gs_app_add_related (app, services_app);
```
wjt: Does this cause the related app to be upgraded if it is already installed but an upgrade is available?

plugins/core/gs-appstream.c
```diff
@@ -1048,6 +1048,32 @@ gs_appstream_search (GsPlugin *plugin,
...
+	/* If the caller has provided the custom "Endless::HasDiscoveryFeedContent"
+	 * search string (which is squashed to lower case by the search machinery),
+	 * add a search for it in the custom metadata dictionary of each component.
+	 *
+	 * FIXME: Technically we should also check ` and text()='true'` in the
+	 * XPath selector, but libxmlb doesn’t support that. So this will return
+	 * all components which specify
+	 *    <custom><value key="Endless::HasDiscoveryFeedContent">false</value></custom>
+	 * too. */
+	if (g_strv_contains (values, "endless::hasdiscoveryfeedcontent")) {
```
wjt: It seems that previously all search terms were applied to all apps' metadata, whereas this is a more targeted approach. What component makes this query? I couldn't find it in a quick search – maybe worth a cross-reference in the commit message and/or this comment?

src/gs-loading-page.c
wjt: Yep, as written this patch will have no effect.

pwithnall: Good spot. Fixed in the upstream MR and backported again.

plugins/eos/gs-plugin-eos.c
pwithnall: I think so, but I haven’t actually tested this EKN stuff at all. I was going for the approach of getting it rebased and merged, and then potentially reworking it after merge/QA, once it’s in an image and testing is easier (and fixes can be distributed amongst more people than just me).

plugins/core/gs-appstream.c
pwithnall: The query is made in `src/gs-discovery-feed-content-provider.c`. I’ll update the commit message.

plugins/eos/gs-plugin-eos.c
wjt: OK – can you make sure that we document this as a (dev?) AC somewhere? I guess it would be:

* Ensure EknServicesMultiplexer is //not// the latest version (perhaps we need instructions to downgrade it)
* Install a new knowledge app
* [ ] EknServicesMultiplexer is upgraded to the latest version

pwithnall: Added to the ACs of https://phabricator.endlessm.com/T27182 (although I think an equivalent AC was already there as ‘re-check all the ACs of T20936)

PR #470, opened 2019-07-24
 flatpak: Mark Endless apps with popular kudo 

uajain:

### DIFF
--- a/plugins/core/gs-appstream.c
+++ b/plugins/core/gs-appstream.c
@@ -1403,6 +1403,25 @@ gs_appstream_component_add_keyword (XbBuilderNode *component, const gchar *str)
 }
 
 void
+gs_appstream_component_add_kudo (XbBuilderNode *component, const gchar *str)
+{
+	g_autoptr(XbBuilderNode) kudos = NULL;
+	g_autoptr(XbBuilderNode) kudo = NULL;
+
+	/* create <kudos> if it does not already exist */
+	kudos = xb_builder_node_get_child (component, "kudos", NULL);
+	if (kudos == NULL)
+		kudos = xb_builder_node_insert (component, "kudos", NULL);
+
+	/* create <kudo>str<kudo> if it does not already exist */
+	kudo = xb_builder_node_get_child (kudos, "kudo", str);
+	if (kudo == NULL) {
+		kudo = xb_builder_node_insert (kudos, "kudo", NULL);
+		xb_builder_node_set_text (kudo, str, -1);
+	}
+}
+
+void
 gs_appstream_component_add_provide (XbBuilderNode *component, const gchar *str)
 {
 	g_autoptr(XbBuilderNode) provide = NULL;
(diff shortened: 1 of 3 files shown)


uajain:
This basically worked for me.
At some point, I saw some `xb_builder_node_get_child: assertion 'XB_IS_BUILDER_NODE (self)' failed` failures in `--verbose` log, so will be looking into that tomorrow.

plugins/core/gs-appstream.c
```diff
@@ -1408,6 +1402,25 @@ gs_appstream_component_add_keyword (XbBuilderNode *component, const gchar *str)
...
+	g_autoptr(XbBuilderNode) kudos = NULL;
+	g_autoptr(XbBuilderNode) kudo = NULL;
+
+	/* create <kudos> if it does not already exist */
+	kudos = xb_builder_node_get_child (component, "kudos", NULL);
+	if (kudos == NULL)
+		kudos = xb_builder_node_insert (component, "kudos", NULL);
+
+	/* create <kudo>str<kudo> if it does not already exist */
+	kudo = xb_builder_node_get_child (kudo, "kudo", str);
```
uajain: This was causing the warnings. Fixed up locally. 
```suggestion
	kudo = xb_builder_node_get_child (kudos, "kudo", str);
```

plugins/core/gs-appstream.h
```diff
@@ -71,5 +71,7 @@ void		 gs_appstream_component_add_icon	(XbBuilderNode	*component,
 							 const gchar	*str);
 void		 gs_appstream_component_add_provide	(XbBuilderNode	*component,
 							 const gchar	*str);
+void		 gs_appstream_component_add_kudo	(XbBuilderNode	*component,
```
pwithnall: Shouldn’t this new function be called from somewhere?

uajain: Sorry, I made a `fixup/squash/rebase` fiasco and had hunks not pushed on the PR. 
Pushed now.

plugins/flatpak/gs-flatpak.c
```diff
@@ -323,6 +323,22 @@ gs_flatpak_add_flatpak_keyword_cb (XbBuilderFixup *self,
...
 }
 
+static gboolean
+gs_flatpak_mark_endlessm_apps_popular_cb (XbBuilderFixup *self,
+					   XbBuilderNode *bn,
+					   gpointer user_data,
+					   GError **error)
+{
+	if (g_strcmp0 (xb_builder_node_get_element (bn), "component") == 0) {
+		g_autoptr(XbBuilderNode) id = xb_builder_node_get_child (bn, "id", NULL);
```
pwithnall: It’s possible for `id` to be `NULL` if the XML is malformed (a `<component>` doesn’t contain an `<id>`), which will cause a crash.

plugins/flatpak/gs-flatpak.c
```diff
@@ -323,6 +323,22 @@ gs_flatpak_add_flatpak_keyword_cb (XbBuilderFixup *self,
...
+		if (g_str_has_prefix (xb_builder_node_get_text (id), "com.endlessm.") ||
+		    g_str_has_prefix (xb_builder_node_get_text (id), "com.endlessnetwork.")) {
```
pwithnall: It’s possible for `xb_builder_node_get_text()` to return `NULL` (if the XML contains `<id/>` or `<id></id>`), which `g_str_has_prefix()` won’t be happy with.

pwithnall:
[Not needed for now](https://phabricator.endlessm.com/T26507#740922), so I’ll close this.

PR #473, opened 2019-07-31
gs-details-page: Multiple UI fixes

uajain:
https://phabricator.endlessm.com/T27332

### DIFF
--- a/src/gs-details-page.c
+++ b/src/gs-details-page.c
@@ -1877,6 +1877,9 @@ _set_app (GsDetailsPage *self, GsApp *app)
 	g_signal_connect_object (self->app, "notify::license",
 				 G_CALLBACK (gs_details_page_notify_state_changed_cb),
 				 self, 0);
+	g_signal_connect_object (self->app, "notify::quirk",
+				 G_CALLBACK (gs_details_page_notify_state_changed_cb),
+				 self, 0);
 	g_signal_connect_object (self->app, "notify::progress",
 				 G_CALLBACK (gs_details_page_progress_changed_cb),
 				 self, 0);
--- a/src/gs-details-page.ui
+++ b/src/gs-details-page.ui
@@ -153,7 +153,7 @@
                             <property name="margin_bottom">14</property>
                             <property name="spacing">9</property>
                             <child>
-                              <object class="GtkLinkButton" id="button_details_add_shortcut">
+                              <object class="GtkButton" id="button_details_add_shortcut">
                                 <property name="visible">False</property>
                                 <property name="can_focus">True</property>
                                 <property name="use_underline">True</property>
@@ -161,7 +161,7 @@
                               </object>
                             </child>
                             <child>
-                              <object class="GtkLinkButton" id="button_details_remove_shortcut">
+                              <object class="GtkButton" id="button_details_remove_shortcut">
                                 <property name="visible">False</property>
                                 <property name="can_focus">True</property>
                                 <property name="use_underline">True</property>
--- a/src/gs-details-page.ui
+++ b/src/gs-details-page.ui
@@ -197,18 +197,6 @@
                               </object>
                             </child>
                             <child>
-                              <object class="GtkButton" id="button_remove">
-                                <property name="visible">False</property>
-                                <property name="use_underline">True</property>
-                                <property name="label" translatable="yes">_Uninstall</property>
-                                <property name="width_request">105</property>
-                                <property name="can_focus">True</property>
-                                <property name="receives_default">True</property>
-                                <property name="halign">start</property>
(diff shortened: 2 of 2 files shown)


wjt:
(Why the `trivial: ` prefix? Why are these 3 changes all in 1 commit?)

src/gs-details-page.ui
```diff
@@ -153,15 +153,18 @@
...
                             <property name="spacing">9</property>
                             <child>
-                              <object class="GtkLinkButton" id="button_details_add_shortcut">
+                              <object class="GtkButton" id="button_details_add_shortcut">
                                 <property name="visible">False</property>
                                 <property name="can_focus">True</property>
                                 <property name="use_underline">True</property>
                                 <property name="label" translatable="yes" comments="Translators: A label for a button to add a shortcut to the selected application.">_Add to Desktop</property>
+                                <style>
+                                  <class name="suggested-action"/>
```
pwithnall: Does this need to change dynamically as different buttons are made visible/hidden?

src/gs-details-page.c
```diff
@@ -1877,6 +1877,9 @@ _set_app (GsDetailsPage *self, GsApp *app)
 	g_signal_connect_object (self->app, "notify::license",
 				 G_CALLBACK (gs_details_page_notify_state_changed_cb),
 				 self, 0);
+	g_signal_connect_object (self->app, "notify::quirk",
+				 G_CALLBACK (gs_details_page_notify_state_changed_cb),
+				 self, 0);
```
pwithnall: This should definitely go upstream.

pwithnall:
I agree with all of Will's suggestions too.

src/gs-details-page.ui
uajain: Not sure why being dynamic here? I have taken this change from `eos3.6` branch (i.e. this is a fallout from the 3.32 rebase.) 

Both `_add_shortcut` and `_remove_shortcut` are separate buttons and visibility is toggled according to the given situation in the icon-grid. Not sure why we need to make the style class being dynamically applied

pwithnall: As in, buttons on the details page are often shown/hidden based on the state of the app. We don’t ever want the situation where *two* buttons are visible, both with class `suggested-action`.

I don’t know if that’s the case after applying this PR, but please check.

uajain:
> I believe we are the only user of this (upstream) UI. It was originally added upstream by Joaquim. Shouldn't most/all of this PR go upstream too?

Sorry, I was not aware aware of this fact! I never saw desktop shortcuts in gnome-software hence, assumed its downstream only. I will push all this upstream

src/gs-details-page.ui
uajain: No, that's not the case. I can see it in action. It is making the behavior consistent with what we have/had in `3.6` OS's gnome-software. 

(I have a 3.6 release OS image VM which I use to compare and contrast these UI fixes).

uajain:
I have upstreamed most of the changes : https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/295

Just one changed ineeds to kept downstream, which is the "Uninstall" button / UI placement.

src/gs-details-page.ui
wjt: I think Philip's point is that both this button, and "Launch", have the `suggested-action` style class. Won't this mean that, when the app is installed but not on the desktop, both "Launch" and "Add to Desktop" will be blue? Could you take screenshots of the 3 states (not installed, installed but not on desktop, installed and on desktop)?

uajain: **Not installed**
Screenshot from 2019-08-01 20-08-39


**Installed and not on desktop** 
Screenshot from 2019-08-01 20-08-15


**Installed and  on desktop**
Screenshot from 2019-08-01 20-08-22

wjt: https://developer.gnome.org/hig/stable/buttons.html.en

> Each view should only include a single suggested or destructive button.

De-emphasising Launch was a deliberate downstream change, see discussion in https://phabricator.endlessm.com/T14949 and designs linked there. I think we should just drop the `suggested-action` on the launch button.

uajain: P

> https://developer.gnome.org/hig/stable/buttons.html.en
> 
> > Each view should only include a single suggested or destructive button.
> 
> De-emphasising Launch was a deliberate downstream change, see discussion in https://phabricator.endlessm.com/T14949 and designs linked there. I think we should just drop the `suggested-action` on the launch button.

Pushed a `fixup!` which is linking where this particular changed should have been introduced downstream. Also, made a note in the commit message of the original commit_id on `eos3.6` branch.

PR #477, opened 2019-08-09
flatpak: Introduce a utility to resolve default arch for refs

uajain:
This is originated to address the case of external-appstream
that can be used to mention additional appdata for a given GsApp.
However, the `source` parameter for a flatpak GsApp is dependent
on the arch of the system. This patch allows the external appstream
to optionally omit the arch and gnome-software will assume the
default arch and re-create the source parameter with the correct ref.
The corrected ref is parsed through `flatpak_ref_parse` for sanity.

Example of an external appstream 'source' parameter resolution:
<bundle type="flatpak">app/org.telegram.desktop//stable</flatpak>
resolves to: `app/org.telegram.desktop/x86_64/stable` for a
x86_64 platform.

### DIFF
--- a/plugins/flatpak/gs-flatpak.c
+++ b/plugins/flatpak/gs-flatpak.c
@@ -1701,6 +1701,7 @@ gs_refine_item_metadata (GsFlatpak *self, GsApp *app,
 			 GError **error)
 {
 	g_autoptr(FlatpakRef) xref = NULL;
+	const gchar *source;
 
 	/* already set */
 	if (gs_flatpak_app_get_ref_name (app) != NULL)
@@ -1710,9 +1711,10 @@ gs_refine_item_metadata (GsFlatpak *self, GsApp *app,
 	if (gs_app_get_kind (app) == AS_APP_KIND_SOURCE)
 		return TRUE;
 
+	source = gs_app_get_source_default (app);
 	/* AppStream sets the source to appname/arch/branch, if this isn't set
 	 * we can't break out the fields */
-	if (gs_app_get_source_default (app) == NULL) {
+	if (source == NULL) {
 		g_autofree gchar *tmp = gs_app_to_string (app);
 		g_warning ("no source set by appstream for %s: %s",
 			   gs_plugin_get_name (self->plugin), tmp);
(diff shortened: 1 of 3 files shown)


plugins/flatpak/gs-flatpak.c
```diff
@@ -1710,22 +1711,32 @@ gs_refine_item_metadata (GsFlatpak *self, GsApp *app,
...
+		g_autofree gchar *resolved_source;
+		resolved_source = gs_flatpak_app_source_resolve_default_arch (source);
+
+		/* Sources mentioned in external-appstream might use refs without specifying
+		 * the arch. Hence, try to parse the arch-resolved ref and see if it succeeds */
+		xref = flatpak_ref_parse (resolved_source, NULL);
+		if (xref == NULL) {
+			gs_flatpak_error_convert (error);
+			g_prefix_error (error, "failed to parse '%s': ",
+					gs_app_get_source_default (app));
```
pwithnall: Nitpick: I think this could use `source` rather than calling `gs_app_get_source_default()` again.

plugins/flatpak/gs-flatpak.c
```diff
@@ -1710,22 +1711,32 @@ gs_refine_item_metadata (GsFlatpak *self, GsApp *app,
...
 
 	/* parse the ref */
-	xref = flatpak_ref_parse (gs_app_get_source_default (app), error);
+	xref = flatpak_ref_parse (source, error);
 	if (xref == NULL) {
-		gs_flatpak_error_convert (error);
-		g_prefix_error (error, "failed to parse '%s': ",
-				gs_app_get_source_default (app));
-		return FALSE;
+		g_autofree gchar *resolved_source;
```
pwithnall: Nitpick: Either initialise this to `NULL` at declaration time, or merge it with the assignment on the following line (possibly with a single assignment split over two lines). Otherwise later refactoring could result in this variable being uninitialised.

pwithnall:
I copied my review comments to [the upstream merge request](https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/297/diffs), which I think should be dealt with before this one. This should just become a backport.

(Sorry, in sorting through mail I missed the upstream MR initially.)

uajain:
I have closed the upstream MR for this. I have stated reasons [here](https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/297#note_587813).

It's no-longer needed, hence closing.

PR #478, opened 2019-08-13
Rebase translations

wjt:
This includes a cherry-pick of an upstream patch to resolve an issue introduced in ec4c5d4c028aef604c95a3c7a2f10ec394481814.

I followed the process at https://phabricator.endlessm.com/w/desktop/translations/#translation-rebases with the exception of the `.tx/config` step since this was already cherry-picked as c4cfda1a9e53931ec14d35efe149aeaf02567ce4.

https://phabricator.endlessm.com/T27087


po/am.po
```diff
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: gnome-software\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2019-08-13 11:02+0000\n"
+"POT-Creation-Date: 2019-08-13 11:55+0000\n"
```
pwithnall: Perhaps the changes in “Reapply old downstream translations” which just change the `POT-Creation-Date` could be dropped? But there’s not much harm in keeping them either, and it does make this commit less automatable. Up to you.

po/af.po
```diff
@@ -32,21 +31,15 @@ msgstr "Programbestuurder vir GNOME"
 msgid ""
 "Software allows you to find and install new applications and system "
 "extensions and remove existing installed applications."
-msgstr ""
-"Sagteware laat mens nuwe programme en uitbreidings vind en installeer en kan "
-"ook reeds geïnstalleerde programme verwyder."
+msgstr "Sagteware laat mens nuwe programme en uitbreidings vind en installeer en kan ook reeds geïnstalleerde programme verwyder."
```
pwithnall: Many of the changes in “Sync with transifex” appear to just be reformatting the line breaks. Perhaps this commit would also benefit from a `ninja gnome-software-update-po` command to normallise things again, unless that’s going to undo the transifex work?

po/am.po
wjt: This is added by the "update .po files to normalize formatting" step. It would indeed be nice to avoid but I'm not quite sure the best way to glue these tools

po/af.po
wjt: We could do. But – every time we run `tx pull -f` we'll get back Transifex's (lack of) line breaks. At some point in the past this was run automatically by Jenkins. https://ci.endlessm-sf.com/job/transifex_pull/ / https://ci.endlessm-sf.com/job/transifex_update/ – neither of these is run automatically at the moment but it would nice for the automation to return. Personally I think we should treat "what Transifex gives us" as the canonical form for our downstream translations.

It's really annoying that all these tools use different layouts for the same information. The best approach I have found for dealing with this is:

```
$ cat .gitattributes 
*.po	diff=podiff
*.pot	diff=podiff
$ git config diff.podiff.command
podiff -D -u
```

There are many things called `podiff` but the one I've used is http://man.gnu.org.ua/manpage/?1+podiff .

wjt:
I think I'll ship this as-is and open some other tasks for improving the translation rebase process.

PR #481, opened 2019-08-19
Revert "gs-updates-section: Avoid crash when switching modes"

pwithnall:
This reverts commit 12e48ce81af4bce099cd09d2b535404d04cfd41c.

It doesn’t solve the bug and crashes still occur — it is obviously not
the right fix, and has not been accepted upstream.

https://phabricator.endlessm.com/T27333


uajain:
Comment relevant to this revert is [here](https://phabricator.endlessm.com/T27333#746550).

PR #482, opened 2019-08-21
eos-blacklist: Allow multiple locales for g-s apps

mazen-asef:
Make eos-blacklist avoid blacklisting apps that ends with a specific
set of locales dictated by the flatpak user's configuration file.

https://phabricator.endlessm.com/T24213

~** Only merge-able after https://github.com/flatpak/flatpak/pull/3058~
~** Only merge-able after the flatpak rebase~
~** Only merge-able after https://github.com/flatpak/flatpak/pull/3152~

### DIFF
--- a/plugins/eos-blacklist/gs-plugin-eos-blacklist.c
+++ b/plugins/eos-blacklist/gs-plugin-eos-blacklist.c
@@ -35,6 +35,9 @@
 #include <sys/types.h>
 #include <sys/xattr.h>
 
+#define GNOME_DESKTOP_USE_UNSTABLE_API "unstable"
+#include <libgnome-desktop/gnome-languages.h>
+
 #define ENDLESS_ID_PREFIX "com.endlessm."
 
 #define EOS_IMAGE_VERSION_XATTR "user.eos-image-version"
@@ -53,6 +56,7 @@ struct GsPluginData
 	char *personality;
 	char *product_name;
 	gboolean eos_arch_is_arm;
+	FlatpakInstallation *installation;
 };
 
 static char *
@@ -177,6 +181,12 @@ gs_plugin_setup (GsPlugin *plugin,
 			g_warning ("No system product name could be retrieved! %s", local_error->message);
 			g_clear_error (&local_error);
 		}
+
+		priv->installation = flatpak_installation_new_system (cancellable, &local_error);
+		if (local_error != NULL) {
+			g_warning ("No system installation could be retrieved! %s", local_error->message);
+			g_clear_error (&local_error);
+		}
 	}
 
 	return TRUE;
@@ -198,6 +208,7 @@ gs_plugin_destroy (GsPlugin *plugin)
 
 	g_free (priv->personality);
 	g_free (priv->product_name);
+	g_clear_object (&priv->installation);
 }
 
 /* Copy of the implementation of gs_flatpak_app_get_ref_name(). */
@@ -207,19 +218,87 @@ app_get_flatpak_ref_name (GsApp *app)
 	return gs_app_get_metadata_item (app, "flatpak::RefName");
 }
 
+static gchar **
+get_flatpak_default_locales (GsPlugin *plugin)
+{
+	g_auto(GStrv) flatpak_default_locales = NULL;
+	g_autoptr(GError) local_error = NULL;
+
+	GsPluginData *priv = gs_plugin_get_data (plugin);
+	if (priv->installation == NULL)
+		return NULL;
+
+	flatpak_default_locales = flatpak_installation_get_default_locales (priv->installation, &local_error);
+	if (local_error != NULL) {
+		g_warning ("No user locales could be retrieved! %s", local_error->message);
+		return NULL;
+	}
+
+	return g_steal_pointer (&flatpak_default_locales);
+}
+
+static gboolean
+assert_valid_locale_match (const char *app_locale,
+                           const char * const *locale_options)
+{
+	int idx;
+	g_autofree char *app_language_code = NULL;
+	g_autofree char *app_territory_code = NULL;
+	g_autofree char *app_modifier = NULL;
+
+	if (!gnome_parse_locale (app_locale, &app_language_code, &app_territory_code, NULL, &app_modifier))
+		return FALSE;
+
+	for (idx = 0; locale_options[idx] != 0; idx++) {
+		g_autofree char *opt_language_code = NULL;
+		g_autofree char *opt_territory_code = NULL;
+		g_autofree char *opt_modifier = NULL;
+
+		if (!gnome_parse_locale (locale_options[idx], &opt_language_code, &opt_territory_code, NULL, &opt_modifier))
+			return FALSE;
+
+		if (g_strcmp0 (app_language_code, opt_language_code) == 0) {
+			/* If the main is a match, try to match territory eg. US or GB */
+			if (app_territory_code != NULL && opt_territory_code != NULL) {
+				if (g_strcmp0 (app_territory_code, opt_territory_code) == 0) {
+					/* If the territory is a match, try to match the modifier eg. latin/cyrillic */
+					if (app_modifier != NULL && opt_modifier != NULL) {
+						if (g_strcmp0 (app_modifier, opt_modifier) == 0)
+							return TRUE;
+					} else {
+						/* If the [modifier] of the app_locale or from the locale options
+						 * is not defined, this is a desireable app */
+						return TRUE;
+					}
+				}
+			} else {
+				/* If the [territory] of the app_locale or from the locale options
+				 * is not defined, this is a desireable app */
+				return TRUE;
+			}
+		}
+	}
+
+	return FALSE;
+}
+
 static gboolean
 gs_plugin_locale_is_compatible (GsPlugin *plugin,
-				const char *locale)
+                                const char *app_locale)
 {
-	g_auto(GStrv) locale_variants;
+	g_auto(GStrv) plugin_locale_variants = NULL;
+	g_auto(GStrv) flatpak_locales = NULL;
 	const char *plugin_locale = gs_plugin_get_locale (plugin);
-	int idx;
 
-	locale_variants = g_get_locale_variants (plugin_locale);
-	for (idx = 0; locale_variants[idx] != NULL; idx++) {
-		if (g_strcmp0 (locale_variants[idx], locale) == 0)
+	/* Check if a variant of a locale is compatible */
+	plugin_locale_variants = g_get_locale_variants (plugin_locale);
+	if (plugin_locale_variants != NULL && assert_valid_locale_match (app_locale, (const char * const *) plugin_locale_variants))
+		return TRUE;
+
+	/* check if the app's locale is compatible with the languages key on the ostree repo file */
+	flatpak_locales = get_flatpak_default_locales (plugin);
+	if (flatpak_locales != NULL && assert_valid_locale_match (app_locale, (const char * const *) flatpak_locales))
 			return TRUE;
-	}
 
 	return FALSE;
 }
--- a/plugins/eos-blacklist/meson.build
+++ b/plugins/eos-blacklist/meson.build
@@ -13,6 +13,7 @@ shared_module(
   dependencies : [
     plugin_libs,
     flatpak,
+    gnome_desktop,
   ],
   link_with : [
     libgnomesoftware,


plugins/eos-blacklist/gs-plugin-eos-blacklist.c
```diff
@@ -153,6 +154,17 @@ get_product_name (GError **error)
 	return g_strndup (image_version, hyphen_index - image_version);
 }
 
+static FlatpakInstallation *
+get_user_installation (GError **error)
+{
+	g_autoptr(GCancellable) cancellable = NULL;
+	g_autoptr(FlatpakInstallation) installation = NULL;
+
+	installation = flatpak_installation_new_user (cancellable, &local_error);
```
pwithnall: `local_error` isn’t defined here. This won’t compile.

plugins/eos-blacklist/gs-plugin-eos-blacklist.c
```diff
@@ -198,6 +216,7 @@ gs_plugin_destroy (GsPlugin *plugin)
 
 	g_free (priv->personality);
 	g_free (priv->product_name);
+	g_free (priv->inst);
```
pwithnall: `FlatpakInstallation` is a GObject, so you must use `g_object_unref()` with it, rather than `g_free()`.

pwithnall: Although it can be `NULL` if the `get_user_installation()` call failed, so you should use `g_clear_object (&priv->inst)`.

plugins/eos-blacklist/gs-plugin-eos-blacklist.c
```diff
@@ -153,6 +154,17 @@ get_product_name (GError **error)
 	return g_strndup (image_version, hyphen_index - image_version);
 }
 
+static FlatpakInstallation *
+get_user_installation (GError **error)
```
pwithnall: `get_user_installation()` is a very thin wrapper around `flatpak_installation_new_user()` which doesn’t actually do any computation of its own. Why does it exist?

plugins/eos-blacklist/gs-plugin-eos-blacklist.c
```diff
@@ -53,6 +53,7 @@ struct GsPluginData
 	char *personality;
 	char *product_name;
 	gboolean eos_arch_is_arm;
+	FlatpakInstallation *inst;
```
pwithnall: Nitpick: I’d recommend calling the member `installation` rather than `inst`. It’s not much more typing, and `inst` could be misinterpreted as a contraction of `instance`.

plugins/eos-blacklist/gs-plugin-eos-blacklist.c
```diff
@@ -207,11 +226,30 @@ app_get_flatpak_ref_name (GsApp *app)
...
+
+	GsPluginData *priv = gs_plugin_get_data (plugin);
+
+	flatpak_default_locales = flatpak_installation_get_default_languages (priv->inst, &local_error);
+	if (local_error != NULL) {
+		g_warning ("No user locales could be retrieved! %s", local_error->message);
+		return NULL;
+	}
+
+	return g_steal_pointer(&flatpak_default_locales);
```
pwithnall: Nitpick: Missing space before `(`

plugins/eos-blacklist/gs-plugin-eos-blacklist.c
```diff
@@ -207,11 +226,30 @@ app_get_flatpak_ref_name (GsApp *app)
...
+static gchar **
+gs_plugin_get_default_locales (GsPlugin *plugin)
+{
+	g_auto(GStrv) flatpak_default_locales = NULL;
+	g_autoptr(GError) local_error = NULL;
+	g_autoptr(FlatpakInstallation) installations = NULL;
+
+	GsPluginData *priv = gs_plugin_get_data (plugin);
+
+	flatpak_default_locales = flatpak_installation_get_default_languages (priv->inst, &local_error);
```
pwithnall: `priv->inst` could be `NULL` (if setup failed), which would result in a critical warning here.

plugins/eos-blacklist/gs-plugin-eos-blacklist.c
```diff
@@ -221,6 +259,15 @@ gs_plugin_locale_is_compatible (GsPlugin *plugin,
 			return TRUE;
 	}
 
+        /* check if the app's locale is listed in the languages key for the ostree repo file */
+        locale_options = gs_plugin_get_default_locales(plugin);
```
pwithnall: Nitpick: Missing space before `(`

plugins/eos-blacklist/gs-plugin-eos-blacklist.c
```diff
@@ -221,6 +259,15 @@ gs_plugin_locale_is_compatible (GsPlugin *plugin,
...
+        if (locale_options != NULL) {
+		for (idx = 0; locale_options[idx] != NULL; idx++) {
```
pwithnall: Nitpick: You can combine the outer `if` and `for` as:
```
for (idx = 0; locale_options != NULL && locale_options[idx] != NULL; idx++) {
```

pwithnall: That would save one level of nesting.

plugins/eos-blacklist/gs-plugin-eos-blacklist.c
```diff
@@ -207,11 +226,30 @@ app_get_flatpak_ref_name (GsApp *app)
 	return gs_app_get_metadata_item (app, "flatpak::RefName");
 }
 
+static gchar **
+gs_plugin_get_default_locales (GsPlugin *plugin)
```
pwithnall: What was the reasoning for choosing the name `gs_plugin_get_default_locales()`? Typically, the prefix `gs_plugin_` is only used in plugins for the plugin vfuncs which they implement. Internal helper functions should be named differently.

I’d call this `get_default_locales()`.

plugins/eos-blacklist/gs-plugin-eos-blacklist.c
```diff
@@ -207,20 +215,48 @@ app_get_flatpak_ref_name (GsApp *app)
...
 		if (g_strcmp0 (locale_variants[idx], locale) == 0)
 			return TRUE;
 	}
 
+        /* check if the app's locale is listed in the languages key for the ostree repo file */
+	locale_options = get_default_locales (plugin);
+	for (idx = 0; locale_options != NULL && locale_options[idx] != NULL; idx++) {
+		if (g_strcmp0 (locale_options[idx], locale) == 0)
+			return TRUE;
+	}
```
pwithnall: You could actually simplify this further using [`g_strv_contains()`](https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strv-contains):
```
locale_options = get_default_locales (plugin);
if (locale_options != NULL && g_strv_contains (locale_options, locale))
	return TRUE;
```

Sorry I didn’t think of this before. You could also simplify the loop above using `g_strv_contains()` (as a separate commit).

mazen-asef: Oh yeah, that makes sense :) will do!

mazen-asef:
Ok, https://github.com/flatpak/flatpak/pull/3058 has been merged, but after discussion with @wjt this settings will be pulled from the system-side configuration of flatpak, instead of the user's config. I will have to change how `priv->installation` gets the installation object, and will update it as soon as it's done.

mazen-asef:
@pwithnall when adding `g_strv_contains` I had to cast the array properly so it is accepted. Let me know if that's too hacky. 

This is the final version of this PR, I think, as I have switched the installation it fetches (now from the system instead of the user). Also tagging @wjt

pwithnall:
> @pwithnall when adding `g_strv_contains` I had to cast the array properly so it is accepted. Let me know if that's too hacky.

That looks correct (the need to cast it is annoying, but a limitation of the GLib API, and we can’ŧ change it).

From my POV, it would be better for this to wait until the 3.34 rebase of gnome-software has landed (which will happen shortly) and then it can be rebased on that trivially.

pwithnall:
OK, I just landed the 3.34 rebase, so this will need rebasing again (sorry). Then we can give it a final review and hopefully merge it.

plugins/eos-blacklist/gs-plugin-eos-blacklist.c
```diff
@@ -207,19 +215,42 @@ app_get_flatpak_ref_name (GsApp *app)
...
-	for (idx = 0; locale_variants[idx] != NULL; idx++) {
-		if (g_strcmp0 (locale_variants[idx], locale) == 0)
-			return TRUE;
-	}
+	if (locale_variants != NULL && g_strv_contains ((const gchar * const *) locale_variants, locale))
+		return TRUE;
+
+        /* check if the app's locale is listed in the languages key for the ostree repo file */
+	locale_options = get_default_locales (plugin);
+	if (locale_options != NULL && g_strv_contains ((const gchar * const *) locale_options, locale))
```
wjt: Perhaps a stupid question, but… do we not need to go through `g_get_locale_variants` for these locales too, in one or other direction? If one of the default locales is `en`, then I guess we'd expect `com.endlessm.times_of_india.en_IN` to match here?

wjt:
Beside the behavioural question above – looks fine.

plugins/eos-blacklist/gs-plugin-eos-blacklist.c
mazen-asef: I think the direction should be: people with the default locale being `en_IN` and seeing the `com.endlessm.encyclopedia.en`, not the other way around, right? And I think the previous if would match that.

mazen-asef: Oh, of course, you mean the extra-language settings - since as of now this is strictly a configuration of the image personality, this will be pretty much targeted to a specific audience, so I don't think there's a need to derive the languages that way

wjt: I just want to make sure this limitation is understood: as this code is currently implemented, only apps which **exactly** match an element of `xa.extra-languages` will be considered. We can and do use both one-part language suffixes (`.en`, `.pt`) and two-part locale suffixes (`.en_IN`, `.pt_BR`) on apps, so what we're implicitly saying here is that in the cases where we rely on `xa.extra-languages` to cause certain apps to be displayed, we will only use two-character language suffixes for the apps in question, and accept that apps with full locale suffixes will not be shown.

pwithnall: To me, that sounds like a very subtle distinction which is going to get forgotten soon, and is going to cause someone half a day’s debugging of gnome-software to work out why an app isn’t being installed when it should.

Are there any downsides to supporting matching locale variants?

mazen-asef: Hmm no I don't think there are downsides to that - flatpak only allows setting one-part language suffixes (for both `xa.languages` and `xa.extra-languages`) so, the only way to make two-part language suffixes would be to whitelist them if the one-part counterpart is present on the list.

plugins/eos-blacklist/gs-plugin-eos-blacklist.c
```diff
@@ -238,19 +238,18 @@ static gboolean
...
-	locale_variants = g_get_locale_variants (plugin_locale);
-	if (locale_variants != NULL && g_strv_contains ((const gchar * const *) locale_variants, locale))
-		return TRUE;
+	int idx;
 
         /* check if the app's locale is listed in the languages key for the ostree repo file */
 	locale_options = get_default_locales (plugin);
-	if (locale_options != NULL && g_strv_contains ((const gchar * const *) locale_options, locale))
-		return TRUE;
+	for (idx = 0; locale_options[idx] != NULL; idx++) {
```
wjt: This will crash if `locale_options` is `NULL`.

plugins/eos-blacklist/gs-plugin-eos-blacklist.c
```diff
@@ -238,19 +238,18 @@ static gboolean
...
+		/* Check if a variant of a locale is compatible, e.g. 'pt' for 'pt_BR' */
+		locale_variants = g_get_locale_variants (locale_options[idx]);
```
wjt: This is the wrong way round. We know that the entries of `locale_options` will always be two-character language codes, like `pt`. It is the app's suffix, stored in `locale`, which may be either `pt_BR` or `pt` (for example).

mazen-asef: Sorry, this edit was just so I could save my work to switch vms haha

mazen-asef:
This last change should address the derived locales issue. @wjt @pwithnall could you take a look? I had to rollback a few changes to fix the issue.

treitter:
> We need to not change the filtering semantic so that if the user is in es_MX, the es_GT apps show up. If we set a more specific region than just the language, it's intentional - eg regional news, content or significant dialect differences. We need a way to set a filter in Flatpak that can be this specific (es_GT) and have Flatpak be happy, and GNOME Software maintain its current matching semantics of going "up" from regions to languages, but not down again.

In other words, if the setting is `es`, match `es`, `es_MX`, `es_GT`, ...; if the setting is `es_MX`, match only `es` and `es_MX`. If the setting is `es_MX es_GT`, match `es`, `es_MX`, `es_GT`, but not `es_ES`.

> Also, even if this was the desired semantic, use of prefix matching is not entirely safe because there are several three letter language codes which have the first two letters in common with a two letter code. https://www.loc.gov/standards/iso639-2/php/code_list.php

Yes, it's critical we support 3-letter languages since, eg, the dialects/languages of Chatino all have 3-letter codes.

I can't immediately tell from the code changes in this PR if either of those requirements are met but I think they need to be met before we merge this.

plugins/eos-blacklist/gs-plugin-eos-blacklist.c
```diff
@@ -207,17 +215,43 @@ app_get_flatpak_ref_name (GsApp *app)
...
 	locale_variants = g_get_locale_variants (plugin_locale);
-	for (idx = 0; locale_variants[idx] != NULL; idx++) {
-		if (g_strcmp0 (locale_variants[idx], locale) == 0)
+	if (locale_variants != NULL && g_strv_contains ((const gchar * const *) locale_variants, locale))
+		return TRUE;
+
+	/* check if the app's locale is listed in the languages key for the ostree repo file */
+	locale_options = get_default_locales (plugin);
+	for(idx = 0; locale_options[idx] != NULL; idx++) {
+		if (g_str_has_prefix (locale, (const gchar * const) locale_options[idx]))
```
pwithnall: I think [Rob’s comment](https://github.com/endlessm/gnome-software/pull/482#pullrequestreview-295910306) still applies here. We can’t do prefix-matching on locales. I think you need a separate function to compare two locales (`locale` and `locale_options[idx]` or `locale_variants[idx]`) and make a decision based on that. The only way to reliably compare two locales is to parse them completely, following the format described in [`man 3 setlocale`](https://jlk.fjfi.cvut.cz/arch/manpages/man/setlocale.3), i.e. ` language[_territory][.codeset][@modifier]`. Each component can be compared using `g_strcmp0()`.

As I understand it, you want to check that at least one of the user’s locale settings (their current locale, or the results of `get_default_locales()`) is equal to, or less specific than, the `locale` describing the flatpak.

Happy to discuss this on Slack if that would help unblock things. :)

plugins/eos-blacklist/gs-plugin-eos-blacklist.c
```diff
@@ -207,17 +215,43 @@ app_get_flatpak_ref_name (GsApp *app)
 	return gs_app_get_metadata_item (app, "flatpak::RefName");
 }
 
+static gchar **
+get_default_locales (GsPlugin *plugin)
```
pwithnall:
```suggestion
get_flatpak_default_locales (GsPlugin *plugin)
```
(to make it a bit clearer that this gets the preferences from flatpak, rather than the environment, or anywhere else)

plugins/eos-blacklist/gs-plugin-eos-blacklist.c
```diff
@@ -207,17 +215,43 @@ app_get_flatpak_ref_name (GsApp *app)
...
+		g_warning ("No user locales could be retrieved! %s", local_error->message);
+		return NULL;
+	}
+
+	return g_steal_pointer (&flatpak_default_locales);
+}
+
 static gboolean
 gs_plugin_locale_is_compatible (GsPlugin *plugin,
 				const char *locale)
```
pwithnall:
```suggestion
				const char *app_locale)
```

(I’m concerned that almost every variable in this function is named `locale`-something, and it’s getting a bit confusing what each variable is — is `locale` the locale from the app, or from flatpak, or from the user’s environment?? I realise this was a problem before, but we might as well fix it while we’re here.)

mazen-asef:
@pwithnall @ramcq @treitter I *think* now we have what we want - I have implemented the changes to the flatpak code base here * https://github.com/flatpak/flatpak/pull/3152 * so this is blocked until that is merged and I have successfully downstreamed this.

plugins/eos-blacklist/gs-plugin-eos-blacklist.c
```diff
@@ -207,19 +215,71 @@ app_get_flatpak_ref_name (GsApp *app)
...
+
+	return g_steal_pointer (&flatpak_default_locales);
+}
+
+static gboolean
+assert_valid_locale_match (const char *app_locale,
+                           char **locale_options)
+{
+	int idx;
+	g_auto(GStrv) app_locale_parts = g_strsplit (app_locale, "_", -1);
```
wjt: https://gitlab.gnome.org/GNOME/gnome-desktop/blob/master/libgnome-desktop/gnome-languages.h#L39-43 exists to parse locales.

plugins/eos-blacklist/gs-plugin-eos-blacklist.c
```diff
@@ -207,19 +215,71 @@ app_get_flatpak_ref_name (GsApp *app)
...
+		g_warning ("No user locales could be retrieved! %s", local_error->message);
+		return NULL;
+	}
+
+	return g_steal_pointer (&flatpak_default_locales);
+}
+
+static gboolean
+assert_valid_locale_match (const char *app_locale,
+                           char **locale_options)
```
pwithnall:
```suggestion
                           const char * const *locale_options)
```

plugins/eos-blacklist/gs-plugin-eos-blacklist.c
pwithnall: (And we [explicitly enable gnome-desktop support](https://github.com/endlessm/gnome-software/blob/921cae46b8656370dfdc21e97c3cffed3da4eaf4/debian/rules#L10), so gnome-software already has it as a dependency, although you might need to change the `meson.build` for the flatpak plugin to link to it.)

plugins/eos-blacklist/gs-plugin-eos-blacklist.c
```diff
@@ -207,19 +215,71 @@ app_get_flatpak_ref_name (GsApp *app)
...
+{
+	int idx;
+	g_auto(GStrv) app_locale_parts = g_strsplit (app_locale, "_", -1);
+
+	for (idx = 0; locale_options[idx] != 0; idx++) {
+		g_auto(GStrv) locale_parts = g_strsplit (locale_options[idx], "_", -1);
+		if (g_strcmp0 (app_locale_parts[0], locale_parts[0]) == 0) {
+			/* If the main is a match, try to match modifiers: eg. region, alphabet */
```
pwithnall: Please use the same terminology as `man setlocale`, i.e. ‘territory’ and ‘codeset’ — locales are confusing enough already without introducing alternative terminology!

plugins/eos-blacklist/gs-plugin-eos-blacklist.c
```diff
@@ -35,6 +35,9 @@
 #include <sys/types.h>
 #include <sys/xattr.h>
 
+#define GNOME_DESKTOP_USE_UNSTABLE_API "unstable"
```
mazen-asef: There's just this one where I couldn't find the standard practice on how to define this, so I went with `unstable`

wjt:
I have untagged 3 reviewers. @pwithnall since you reviewed the last pass, please could you review this one?

PR #489, opened 2019-09-19
T25941 Fix offline USB app updates

mwleeds:
https://phabricator.endlessm.com/T25941


mwleeds:
I guess this needs rebasing. That diff is hilarious: " +380,980 −102,624 "

mwleeds:
I tested this after rebasing and it seems to still work as intended.

PR #492, opened 2019-09-26
wip: Make icons appear in installed tab (Don't merge)

uajain:
(**Do not merge** this pull request. It's just for demo and discussion holder)

tldr; Desktop files are pushed into the XML silo and
GNOME Software queries the XML silo to know which apps
are installed. There is a .desktop ---> XML mapping
support in appstream which gnome-software used to make
use of until it broke via by introducing a new-er API.

xb_builder_source_add_adapter for "application/x-desktop"
handling is buggy. See https://gitlab.gnome.org/GNOME/gnome-software/issues/811

https://phabricator.endlessm.com/T27779


uajain:
Should be correctly addressed in https://github.com/endlessm/gnome-software/pull/494

PR #493, opened 2019-09-27
Fix NULL pointer dereference in flatpak gs_plugin_download()

wjt:
Unfortunately 68110bf345e38b69f7b4ddf09990be3b53074766 meant that an incorrect early return was replaced by a `NULL` pointer dereference since later in the function, `flatpak` is assumed to be non-`NULL`. This manifested itself in gnome-software segfaulting in the cases where it would previously have silently failed to install any Flatpaks:

```
(gdb) bt 
#0  gs_flatpak_get_installation (self=self@entry=0x0) at ../plugins/flatpak/gs-flatpak.c:3521
#1  0x00007fffdf23d49f in gs_flatpak_has_space_to_update (flatpak=flatpak@entry=0x0, list=list@entry=0x555557120180, 
    is_auto_update=is_auto_update@entry=1) at ../plugins/flatpak/gs-plugin-flatpak.c:546
#2  0x00007fffdf23e306 in gs_plugin_download (plugin=plugin@entry=0x55555592bb20, list=list@entry=0x7fffcc093980, 
    cancellable=cancellable@entry=0x5555570d7720, error=error@entry=0x7fffcae7c9d8) at ../plugins/flatpak/gs-plugin-flatpak.c:598
[…]
(gdb) frame 0
#0  gs_flatpak_get_installation (self=self@entry=0x0) at ../plugins/flatpak/gs-flatpak.c:3521
3521		return self->installation;
(gdb) p self
$11 = (GsFlatpak *) 0x0
(gdb) frame 2
#2  0x00007fffdf23e306 in gs_plugin_download (plugin=plugin@entry=0x55555592bb20, list=list@entry=0x7fffcc093980, 
    cancellable=cancellable@entry=0x5555570d7720, error=error@entry=0x7fffcae7c9d8) at ../plugins/flatpak/gs-plugin-flatpak.c:598
598		if (!gs_flatpak_has_space_to_update (flatpak, list_tmp, is_auto_update)) {
(gdb) p gs_app_get_id (gs_app_list_index (list, gs_app_list_length (list) - 1))
$12 = (const gchar *) 0x7fffcc004c50 "com.endlessm.proxy.EOSUpdatesProxy"
```

Fix this by grouping the elements of the `GsAppList` by the `GsFlatpak` (if any) that they apply to, and processing each group as a separate batch. This is the same fix applied to `gs_plugin_update` by 12c3f6464c9e9ecd0111b35facf931fc28ad91a9.

We have a downstream patch to this function which conflicted with this change, so in this branch I have reverted it, applied the two patches that I submitted upstream, and then un-reverted it, fixing the conflicts. Assuming the upstream patches are accepted in this form, this means that the next rebase can simply drop the old patch + revert old patch pair, and apply the refreshed downstream patch (if it is still needed).

TODO:

- [x] Ensure that `gs_plugin_update` still works after this refactoring. 30 queued updates are currently downloading on my system; I guess we'll see once they're done.

https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/318
https://phabricator.endlessm.com/T27438

### DIFF
--- a/plugins/flatpak/gs-plugin-flatpak.c
+++ b/plugins/flatpak/gs-plugin-flatpak.c
@@ -507,51 +507,55 @@ gboolean
 gs_plugin_download (GsPlugin *plugin, GsAppList *list,
 		    GCancellable *cancellable, GError **error)
 {
-	GsFlatpak *flatpak = NULL;
-	g_autoptr(FlatpakTransaction) transaction = NULL;
-	g_autoptr(GsAppList) list_tmp = gs_app_list_new ();
+	g_autoptr(GHashTable) applist_by_flatpaks = NULL;
+	GHashTableIter iter;
+	gpointer key, value;
 
-	/* not supported */
-	for (guint i = 0; i < gs_app_list_length (list); i++) {
-		GsApp *app = gs_app_list_index (list, i);
-		flatpak = gs_plugin_flatpak_get_handler (plugin, app);
-		if (flatpak != NULL)
-			gs_app_list_add (list_tmp, app);
-	}
-	if (gs_app_list_length (list_tmp) == 0)
-		return TRUE;
+	/* build and run transaction for each flatpak installation */
+	applist_by_flatpaks = _flatpak_group_apps_by_installation (plugin, list);
+	g_hash_table_iter_init (&iter, applist_by_flatpaks);
+	while (g_hash_table_iter_next (&iter, &key, &value)) {
+		GsFlatpak *flatpak = GS_FLATPAK (key);
+		GsAppList *list_tmp = GS_APP_LIST (value);
+		g_autoptr(FlatpakTransaction) transaction = NULL;
 
-	if (!gs_plugin_has_flags (plugin, GS_PLUGIN_FLAGS_INTERACTIVE)) {
-		g_autoptr(GError) error_local = NULL;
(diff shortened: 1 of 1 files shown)


wjt:
As discussed on the upstream MR (which has been accepted with the versions of the patches I pushed to this PR 35 minutes ago) and on our Phab ticket, the test result was not an overwhelming success, but I believe neither of the bugs I hit were introduced by this fix.

plugins/flatpak/gs-plugin-flatpak.c
```diff
@@ -533,6 +630,20 @@ gs_plugin_download (GsPlugin *plugin, GsAppList *list,
...
 		}
 
+		/* Is there enough disk space to download? */
+		if (!gs_flatpak_has_space_to_update (flatpak, list_tmp, is_auto_update)) {
+			g_debug ("Skipping %s: not enough space on disk",
+			 	 (is_auto_update ? "automatic update" : "update"));
+			if (is_auto_update) {
+				/* Do not return an error because we don't want to see a banner
+				 * showing up out of the blue for possibly multiple update failures */
+				continue;
```
uajain: The comment above needs a bit readjusting as we are continuing to the next flatpak group and see if they are auto-updatable or not.

wjt:
I refreshed the comments, and I also made the `g_debug` messages more specific and fixed some whitespace errors.

PR #497, opened 2019-10-04
Fix USB category (again)

mwleeds:
This makes apps show up in the USB category more reliably, and makes only apps show up, rather than various kinds of runtimes showing up also.

https://phabricator.endlessm.com/T25941

https://phabricator.endlessm.com/T23882

### DIFF
--- a/plugins/flatpak/gs-flatpak.c
+++ b/plugins/flatpak/gs-flatpak.c
@@ -485,6 +485,10 @@ gs_flatpak_mark_apps_from_usb_remote (GsFlatpak *self,
 			continue;
 		}
 
+		/* Don't show runtimes, locale extensions, etc. */
+		if (gs_app_get_kind (app) != AS_APP_KIND_DESKTOP)
+			continue;
+
 		gs_app_add_category (app, "usb");
 
 		/* add a keyword so users can search for "usb" */
--- a/src/gs-category-page.c
+++ b/src/gs-category-page.c
@@ -238,6 +238,11 @@ gs_category_page_get_apps_cb (GObject *source_object,
 	/* ensure the filter will show the apps, not the placeholders */
 	self->num_placeholders_to_show = -1;
 	gtk_flow_box_invalidate_filter (GTK_FLOW_BOX (self->category_detail_box));
+
+	if (g_strcmp0 (gs_category_get_id (self->category), "usb") == 0) {
+		gtk_widget_set_visible (self->no_apps_box, gs_category_get_size (self->subcategory) == 0);
+		gtk_widget_set_visible (self->scrolledwindow_category, gs_category_get_size (self->subcategory) != 0);
+	}
 }
 
 static gboolean
--- a/src/gs-shell.c
+++ b/src/gs-shell.c
@@ -609,6 +609,11 @@ on_overview_categories_loaded (GsOverviewPage *page, gpointer user_data)
 					      user_data);
 	if (!gs_overview_page_set_category (page, data->category_id))
 		gs_shell_change_mode (data->shell, GS_SHELL_MODE_OVERVIEW, NULL, TRUE);
+	else {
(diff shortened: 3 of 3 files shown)


uajain:
> However gs_overview_page_reload() only executes GS_PLUGIN_ACTION_GET_CATEGORIES
> which gets the list of categories but does not reload the apps in each
> one. 

I read some code around it and experimented with gdb. `GS_PLUGIN_ACTION_GET_CATEGORY_APPS,` is executed everytime if I switch among categories in the overview mode. Now , as per your commit message, the "USB" category shows up in the overview mode but apps are not loaded in it, hence you try to load them in conjunction with reloading of the overview page.

My understanding of "correctness" is that, clicking on the newly showed up "USB" category should be execute `GS_PLUGIN_ACTION_GET_CATEGORY_APPS`, no? And that's how apps in the drive should show up. My reading goes as:

* `gs_shell_change_mode` calls `gs_category_page_set_category` (with "USB" category)
* `gs_category_page_set_category` calls `gs_category_page_create_filter`
* `gs_category_page_create_filter` has "USB" specific code that does two things:
   -  Call to `gs_category_page_populate_filtered` which eventually reloads the category page( i.e. executes `GS_PLUGIN_ACTION_GET_CATEGORY_APPS`)
   - Sets visibility of widgets similarly to one of the hunks here.

Hence, I think (in theory) , `GS_PLUGIN_ACTION_GET_CATEGORY_APPS` should be executed if I click the USB category(and that it **should not** be required to externally make the category apps reload). Is the formation of category (UI/gnome-software wise) something special that I am missing out on?

mwleeds:
> > However gs_overview_page_reload() only executes GS_PLUGIN_ACTION_GET_CATEGORIES
> > which gets the list of categories but does not reload the apps in each
> > one.
> 
> I read some code around it and experimented with gdb. `GS_PLUGIN_ACTION_GET_CATEGORY_APPS,` is executed everytime if I switch among categories in the overview mode. Now , as per your commit message, the "USB" category shows up in the overview mode but apps are not loaded in it, hence you try to load them in conjunction with reloading of the overview page.

I tested as well and indeed gs_plugin_add_category_apps() is executed when switching categories, although gs_plugin_add_categories() is not.

> 
> My understanding of "correctness" is that, clicking on the newly showed up "USB" category should be execute `GS_PLUGIN_ACTION_GET_CATEGORY_APPS`, no?

Clicking on the USB category should work, but it should also be automatically selected when the drive is inserted, by the call to `gs_overview_page_set_category()` in `on_overview_categories_loaded()`. Unless you're on a recent eos3.7 ostree, you have to manually make that happen by calling e.g. `gnome-software --local-filename=/run/media/mwleeds/flash_drive` after inserting the drive (see https://phabricator.endlessm.com/T23512).

> And that's how apps in the drive should show up. My reading goes as:
> 
>     * `gs_shell_change_mode` calls `gs_category_page_set_category` (with "USB" category)
> 
>     * `gs_category_page_set_category` calls `gs_category_page_create_filter`
> 
>     * `gs_category_page_create_filter` has "USB" specific code that does two things:
>       
>       * Call to `gs_category_page_populate_filtered` which eventually reloads the category page( i.e. executes `GS_PLUGIN_ACTION_GET_CATEGORY_APPS`)
`gs_category_page_populate_filtered` is only executed for the USB category if `gs_category_get_size (s) < 1`, but in case the drive has apps on it, the category size is not zero (because of the `GS_PLUGIN_ACTION_GET_CATEGORIES` in `gs_overview_page_load()`).

Hopefully that makes sense. Thanks for the review by the way!

uajain:
> Clicking on the USB category should work, but it should also be automatically selected when the drive is inserted, 

I wasn't aware of this behavior. If that's the intention, yes, you need to call `GS_PLUGIN_ACTION_GET_CATEGORY_APPS` externally as you have done.

Thanks for the comments. Merging.

PR #499, opened 2019-10-09
T27438 - Flatpak apps autoupdates

uajain:
See https://phabricator.endlessm.com/T27438#760799

These patches seems to me in a pretty good shape but still lacking upstream review as of now. 
These are cherry-picked from upstream's merge-requests as mentioned in the above phabricator link.
Ideally, we should backport them after merging them upstream but that probably will miss the hard-code freeze. 

I have been interactively tested these patches for autoupdates code path and made sure that the apps update reliably, pointing to the latest flatpak commit. This is one of the major issues of  T27438 as [noticed](https://phabricator.endlessm.com/T27438#758602) by me and Will. 

I have included the commit "updates monitor: Don't skip automatic updates that require new permission"  (which is basically a revert of cc6f8df) because I have strong objection with holding back such apps **without** having proper fallback/user-action mechanism in place. This is being c[ooked upstream](https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/329) and we can always backport it in any point releases. Also, we didn't have this mechanism earlier anyways(not a good argument but still keeping fallout risks as low as possible). 
 
https://phabricator.endlessm.com/T27438

### DIFF
--- a/plugins/flatpak/gs-flatpak-transaction.c
+++ b/plugins/flatpak/gs-flatpak-transaction.c
@@ -255,6 +255,7 @@ _transaction_operation_done (FlatpakTransaction *transaction,
 			     const gchar *commit,
 			     FlatpakTransactionResult details)
 {
+	GsFlatpakTransaction *self = GS_FLATPAK_TRANSACTION (transaction);
 	/* invalidate */
 	GsApp *app = _transaction_operation_get_app (operation);
 	if (app == NULL) {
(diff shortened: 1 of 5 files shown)


plugins/flatpak/gs-flatpak-transaction.c
```diff
@@ -348,11 +373,28 @@ _transaction_add_new_remote (FlatpakTransaction *transaction,
 	return FALSE;
 }
 
+static void
+gs_flatpak_transaction_set_property (GObject *object, guint prop_id, const GValue *value, GParamSpec *pspec)
+{
+	FlatpakTransaction *transaction = FLATPAK_TRANSACTION (object);
+
+	switch ((GsFlatpakTransactionProperty) prop_id) {
```
andrunko: nit: the cast should not be needed here.

uajain: Please refer to the @pwithnall  advise [here](https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/331#note_619942)

pwithnall: The cast is not needed, but gives us additional checking from `-Wswitch-enum` when that’s enabled. I think it should be left in.

PR #502, opened 2019-10-14
Fix crash when switching categories on ARM

andrunko:
This PR also fixes another warning when refining an app with <content_rating> without the "type" attr set (note that this commit is not really needed to fix the crash):
```
22:32:03:0210 GLib g_str_has_prefix: assertion 'str != NULL' failed
```

https://phabricator.endlessm.com/T28074

### DIFF
--- a/lib/gs-app-list-private.h
+++ b/lib/gs-app-list-private.h
@@ -42,19 +42,16 @@ typedef enum {
  * @GS_APP_LIST_FILTER_FLAG_PREFER_INSTALLED:	Prefer installed applications
  * @GS_APP_LIST_FILTER_FLAG_KEY_ID_PROVIDES:	Filter using the provides ID
  *
- * Flags to use when filtering. The priority of eash #GsApp is used to choose
+ * Flags to use when filtering. The priority of each #GsApp is used to choose
  * which application object to keep.
  **/
-typedef enum {
-	GS_APP_LIST_FILTER_FLAG_NONE		= 0,
-	GS_APP_LIST_FILTER_FLAG_KEY_ID		= 1 << 0,
-	GS_APP_LIST_FILTER_FLAG_KEY_SOURCE	= 1 << 1,
-	GS_APP_LIST_FILTER_FLAG_KEY_VERSION	= 1 << 2,
-	GS_APP_LIST_FILTER_FLAG_PREFER_INSTALLED= 1 << 3,
-	GS_APP_LIST_FILTER_FLAG_KEY_ID_PROVIDES	= 1 << 4,
-	/*< private >*/
-	GS_APP_LIST_FILTER_FLAG_LAST
-} GsAppListFilterFlags;
+#define GS_APP_LIST_FILTER_FLAG_NONE             ((guint64) 0)
+#define GS_APP_LIST_FILTER_FLAG_KEY_ID           ((guint64) 1 << 0)
+#define GS_APP_LIST_FILTER_FLAG_KEY_SOURCE       ((guint64) 1 << 1)
+#define GS_APP_LIST_FILTER_FLAG_KEY_VERSION      ((guint64) 1 << 2)
+#define GS_APP_LIST_FILTER_FLAG_PREFER_INSTALLED ((guint64) 1 << 3)
+#define GS_APP_LIST_FILTER_FLAG_KEY_ID_PROVIDES  ((guint64) 1 << 4)
+typedef guint64 GsAppListFilterFlags;
 
 GsAppList	*gs_app_list_copy		(GsAppList	*list);
 guint		 gs_app_list_get_size_peak	(GsAppList	*list);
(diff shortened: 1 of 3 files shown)


andrunko:
The other option would be to cast every place setting `dedupe-flags` on `GsPluginJob`.

andrunko:
Related, see also 31c9dd212156afbb2422188cb59ee8acb5a0719f.

lib/gs-app-list-private.h
```diff
@@ -45,16 +45,13 @@ typedef enum {
  * Flags to use when filtering. The priority of eash #GsApp is used to choose
```
wjt: If submitting this change upstream, `s/eash/each/` here (I know it's existing code)

lib/gs-app-list-private.h
```diff
@@ -45,16 +45,13 @@ typedef enum {
...
-	/*< private >*/
-	GS_APP_LIST_FILTER_FLAG_LAST
-} GsAppListFilterFlags;
+#define GS_APP_LIST_FILTER_FLAG_NONE             ((guint64) 0)
+#define GS_APP_LIST_FILTER_FLAG_KEY_ID           ((guint64) 1 << 0)
+#define GS_APP_LIST_FILTER_FLAG_KEY_SOURCE       ((guint64) 1 << 1)
+#define GS_APP_LIST_FILTER_FLAG_KEY_VERSION      ((guint64) 1 << 2)
+#define GS_APP_LIST_FILTER_FLAG_PREFER_INSTALLED ((guint64) 1 << 3)
+#define GS_APP_LIST_FILTER_FLAG_KEY_ID_PROVIDES  ((guint64) 1 << 4)
+typedef guint64 GsAppListFilterFlags;
```
wjt: Another fix might be to add another private enum member, `GS_APP_LIST_FILTER_FLAG_FORCE_UINT64 = G_MAXUINT64` at the end of the enum, to force it to be 64-bits wide? Personally I like enums more than I like preprocessor macros.

plugins/core/gs-appstream.c
```diff
@@ -601,9 +601,17 @@ gs_appstream_refine_app_content_rating (GsPlugin *plugin,
...
 	g_autoptr(GError) error_local = NULL;
 	g_autoptr(GPtrArray) content_attributes = NULL;
+	const gchar *content_rating_kind = NULL;
 
 	/* get kind */
-	as_content_rating_set_kind (cr, xb_node_get_attr (content_rating, "type"));
+	content_rating_kind = xb_node_get_attr (content_rating, "type");
+	/* we only really expect/support OARS 1.0 and 1.1 */
+	if (content_rating_kind == NULL ||
+	    !g_str_has_prefix (content_rating_kind, "oars-1.")) {
```
wjt: This check is wrong – we don't understand a future `oars-1.2` format.

lib/gs-app-list-private.h
wjt: Really the people to ask would be kalev lember or richard hughes, upstream – it's their taste that matters :-)

plugins/core/gs-appstream.c
andrunko: I know, this was also existing code - I just moved the conditionals around to avoid a warning when checking against a `NULL` type. I will fix the conditional though to only check 1.0 and 1.1.

lib/gs-app-list-private.h
andrunko: Given that we dont have much time to integrate this to 3.7.0 I would go with the macros and then push the changes upstream and we can update again in a future version depending on upstream feedback. Note that other flags are also using macros, see https://github.com/endlessm/gnome-software/commit/31c9dd212156afbb2422188cb59ee8acb5a0719f.

I will update the typo though.

wjt: Oh, I see that patch made an analogous change from enum to `#define`. I guess that answers the What Would Upstream Do question!

plugins/core/gs-appstream.c
wjt: Oh! I missed that it's existing code.

andrunko:
Force pushed with the typo and the conditional fixed (to only accept oars-1.0 and 1.1). Care to check again before merging?

plugins/core/gs-appstream.c
pwithnall: Upstreamed as https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/344.

lib/gs-app-list-private.h
pwithnall: Upstreamed using the original `enum` as https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/343. I think that https://github.com/endlessm/gnome-software/commit/31c9dd212156afbb2422188cb59ee8acb5a0719f is a red herring because it was written by Robert Ancell, so it shows us What Upstream Will Accept, not What Upstream Would Do (or What Upstream Should Do, if I’m on my high horse).

lib/gs-app-list-private.h
pwithnall: Done in this PR by Andre, and in the upstream patch (also by Andre) here: https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/343

PR #503, opened 2019-10-15
Show system apps in their relevant categories

uajain:
This fixes the problem of system installed apps not appearing
in their relevant categories as mentioned in their appdata.
These apps are marked as wildcards by gs_appstream_add_category_apps,
hence the refine function for wilcards should also match the real
GsApp using the appdata query.

https://phabricator.endlessm.com/T27899

https://phabricator.endlessm.com/T27505

### DIFF
--- a/plugins/core/gs-appstream.c
+++ b/plugins/core/gs-appstream.c
@@ -1182,22 +1182,24 @@ gs_appstream_add_category_apps (GsPlugin *plugin,
 	}
 	for (guint j = 0; j < desktop_groups->len; j++) {
 		const gchar *desktop_group = g_ptr_array_index (desktop_groups, j);
-		g_autofree gchar *xpath = NULL;
+		g_autoptr(GString) xpath = g_string_new (NULL);
 		g_auto(GStrv) split = g_strsplit (desktop_group, "::", -1);
 		g_autoptr(GPtrArray) components = NULL;
 
 		/* generate query */
 		if (g_strv_length (split) == 1) {
-			xpath = g_strdup_printf ("components/component/categories/"
-						 "category[text()='%s']/../..",
-						 split[0]);
+			xb_string_append_union (xpath,
+						"components/component/categories/"
+						"category[text()='%s']/../..",
+						split[0]);
 		} else if (g_strv_length (split) == 2) {
-			xpath = g_strdup_printf ("components/component/categories/"
-						 "category[text()='%s']/../"
-						 "category[text()='%s']/../..",
-						 split[0], split[1]);
+			xb_string_append_union (xpath,
+						"components/component/categories/"
+						"category[text()='%s']/../"
+						"category[text()='%s']/../..",
+						split[0], split[1]);
 		}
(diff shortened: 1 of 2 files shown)


uajain:
I am not deliberating creating an upstream MR for this, as this is very much dependent on outstanding [upstream](https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/335) [MRs ](https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/337) (which we have merged downstream but still awaiting upstream).

Once those issues are sorted out (I am in regular communication with Richard), I will push this one to upstream for upstream review. Otherwise, it will sit and bit-rot and create more confusion.

plugins/core/gs-appstream.c
```diff
@@ -1182,22 +1182,33 @@ gs_appstream_add_category_apps (GsPlugin *plugin,
...
-			xpath = g_strdup_printf ("components/component/categories/"
-						 "category[text()='%s']/../..",
-						 split[0]);
+			xb_string_append_union (xpath,
+						"components/component/categories/"
+						"category[text()='%s']/../..",
+						split[0]);
+			xb_string_append_union (xpath,
+						"component/categories/"
+						"category[text()='%s']/../../description/..",
```
andrunko: Changes lgtm to me but do we only want to select apps with description set?

plugins/core/gs-appstream.c
```diff
@@ -1182,22 +1182,33 @@ gs_appstream_add_category_apps (GsPlugin *plugin,
...
-						 split[0], split[1]);
+			xb_string_append_union (xpath,
+						"components/component/categories/"
+						"category[text()='%s']/../"
+						"category[text()='%s']/../..",
+						split[0], split[1]);
+			xb_string_append_union (xpath,
+						"component/categories/"
+						"category[text()='%s']/../"
+						"category[text()='%s']/../../description/..",
```
andrunko: Same here.

plugins/core/gs-appstream.c
uajain: So, `<description>` from appdata acts as a differentiator to filter out non-applications.


A "real" system app in gnome-software contains 2 sources of metadata:
1) Desktop file (There is a XML-mapping (appstream-like) that gets pushed into the xmlb silo)
2) The appdata file.

Many non-application stuff have desktop files (for e.g. `ls  /usr/share/applications/ibus*`) 
We need to make sure this doesn't get bubbled up in the UI. Hence, check for the `<description>` tag which only is mentioned in the app's appdata. 

Indirectly, it also confirms that an desktop file has a real appdata file on  disk. I wrote about this at upstream places [here](https://gitlab.gnome.org/GNOME/gnome-software/issues/813) and [here](https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/337)

uajain: Keeping the conversation unresolved if you have more questions.

I think the PR is good to go, so merging it.
Thanks for the review

PR #504, opened 2019-10-15
wip: installed-page: Treat chromium as an actual app

uajain:
While fixing an upstream issue [1], gnome-software behaviour
was changed, to merge metadata from the desktop file and appdata
for an app under a single GsApp object. This made sure that the
"Installed" tab was properly populated with all the description,
icons, screenshots etc.

Chromium only has a desktop file in EOS. Hence, either one can
add an appdata file to the CMS and make sure it ends up in
/usr/share/metainfo or /usr/share/appdata

OR

force-treat chromium as an actual app (like upstream does for
google-chrome).

(I chose the latter approach in short-term although it would be
good to have an appdata file for chromium-browser which should
be kept updated everytime we release a new chromium version)

[1] https://gitlab.gnome.org/GNOME/gnome-software/issues/813

### DIFF
--- a/src/gs-installed-page.c
+++ b/src/gs-installed-page.c
@@ -162,7 +162,8 @@ gs_installed_page_is_actual_app (GsApp *app)
 	if (gs_app_get_description (app) != NULL)
 		return TRUE;
 	/* special snowflake */
-	if (g_strcmp0 (gs_app_get_id (app), "google-chrome.desktop") == 0)
+	if (g_strcmp0 (gs_app_get_id (app), "google-chrome.desktop") == 0 ||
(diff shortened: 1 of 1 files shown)


uajain:
Marking wip because need to check the behavior on 3.6 if web-browsers are restricted via parental-controls. 

With this PR's change, I restricted the Web-browsers, gnome-software still showed me the `Internet` app row in `Installed` tab. However, if I go to it's detail's page, it had (correctly?) hidden the `Launch` button. 

Need to see if that's the behavior on last 3.6 stable too.

uajain:
> Need to see if that's the behavior on last 3.6 stable too.

gaah, I am wrong, no `Internet` app in system apps on 3.6.
This PR is pointless. Apologies for the noise.

PR #506, opened 2019-10-28
odrs: Only add reviews from non-NULL reviewer's id

pwithnall:
Kodi (tv.kodi.Kodi) segfaulted the entire gnome-software when trying
to install it via flatpak. This was due to the fact that one of
the reviewer's id was essentially NULL. Hence, don't add reviews
whose reviewer-id is NULL.

https://phabricator.endlessm.com/T28338


pwithnall:
Already [merged upstream](https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/345#note_632717), so it only needs a trivial review.

PR #508, opened 2019-10-30
gs-plugin-loader: Only return success if task hasn't been cancelled yet

andrunko:
This fixes an issue when switching categories in a quick succession where `gs_plugin_loader_job_process_async()` is invoked for a category, then gets cancelled before completing when the category is changed but the thread running the initial job still returns a valid result, which then confuses GsCategoryPage which ends up re-adding the items.

**Note**: I am not sure if the issue is on Glib itself, given `gs_plugin_loader_job_process_async()` sets `g_task_set_check_cancellable (task, FALSE)` and `g_task_set_return_on_cancel (task, FALSE)`, but doesn't check if the task was cancelled before returning a valid result.

Note also that the thread doing the initial async job (`gs_plugin_loader_process_thread_cb()`) already checks the cancellable in a few places, but then proceeds to do other things (e.g. filter the list) without considering the cancellable, which in turn does extra processing that should not be needed. We should also consider either adding more checks for cancellable in `gs_plugin_loader_process_thread_cb()` or maybe setting `g_task_set_check_cancellable()` to `TRUE` as an alternative.

https://phabricator.endlessm.com/T28198

### DIFF
--- a/lib/gs-plugin-loader.c
+++ b/lib/gs-plugin-loader.c
@@ -1466,6 +1466,7 @@ gs_plugin_loader_job_process_finish (GsPluginLoader *plugin_loader,
 				     GAsyncResult *res,
 				     GError **error)
 {
+	GTask *task;
 	GsAppList *list = NULL;
 
 	g_return_val_if_fail (GS_IS_PLUGIN_LOADER (plugin_loader), NULL);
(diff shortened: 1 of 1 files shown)


andrunko:
It goes somewhat like this:

1. User changes to category A - gs_category_page_reload is called for category A
   1.1. self->category=A
   1.2. self->cancellable=NULL, creates self->cancellable and proceeds to load category
   1.3. invoke gs_plugin_loader_job_process_async with self->cancellable set
2. User changes to category B - gs_category_page_reload is called for category B
   2.1. self->category=B
   2.2. self->cancellable is set, gets cancelled and recreated
   2.3. invoke gs_plugin_loader_job_process_async with new self->cancellable set
3. User changes back to category A - gs_category_page_reload is called for category A
   3.1. self->category=A
   2.2. self->cancellable is set, gets cancelled and recreated
   3.2. gs_plugin_loader_job_process_async returns for category B (as the cancellable was ignored by it)
   3.3. gs_category_page_get_apps_cb is invoked and checks if the cached list for category A matches the returned list (which is for B)
   3.4. gs_category_page_get_apps_cb then removes the cached items from category A **but** doesn't remove the cached tiles for them (**which also seems like a bug**)

Then the user switches again to B and then A, at which point the cached list for A is missing the expected items, which get re-added and new tiles are created (as they didn't get removed when the cached items on step 3 above were removed).

andrunko:
Force pushed to update commit msg.

pwithnall:
I’m not convinced about the need for `g_task_set_return_on_cancel()`, either, but that’s a separate issue. Code should only need to call `g_task_set_return_on_cancel()` if its thread function calls a blocking function which doesn’t accept a `GCancellable` (like `read()` on a pipe or something). If the thread function can use a `GCancellable` throughout to cancel its work, then that’s preferable as it means the thread state can be cleaned up before a result is returned.

If `g_task_set_return_on_cancel()` is used, the worker thread is left to run unattended, and its result is ignored. That’s fine until you run out of worker threads, or unless the thread function changes some global state which affects other ongoing operations. That seems likely with this code.

As I said, though, the use of `g_task_set_return_on_cancel()` is a separate issue.

andrunko:
Force pushed to change the approach and not use `g_task_set_check_cancellable(FALSE)` (also seems to work fine). This is probably better so we don't have to worry about whether the thread code gets changed to return a valid result without checking if the task was cancelled.

andrunko:
Updated PR yet again after upstream review, more details in the upstream PR.

andrunko:
Merging for now and we can update again depending on any new upstream feedback.

PR #511, opened 2019-11-06
Improve caching/performance

andrunko:
https://phabricator.endlessm.com/T28503

### DIFF
--- a/plugins/flatpak/gs-flatpak.c
+++ b/plugins/flatpak/gs-flatpak.c
@@ -28,6 +28,7 @@ struct _GsFlatpak {
 	GObject			 parent_instance;
 	GsFlatpakFlags		 flags;
 	FlatpakInstallation	*installation;
+	GPtrArray		*installed_refs;
 	GHashTable		*broken_remotes;
 	GMutex			 broken_remotes_mutex;
 	GHashTable		*usb_remotes;
@@ -318,6 +319,9 @@ gs_plugin_flatpak_changed_cb (GFileMonitor *monitor,
 		g_warning ("failed to drop cache: %s", error->message);
 		return;
 	}
+
+	/* drop the installed refs cache */
+	g_clear_pointer (&self->installed_refs, g_ptr_array_unref);
 }
 
 static gboolean
@@ -1675,6 +1679,9 @@ gs_flatpak_refresh (GsFlatpak *self,
 		return FALSE;
 	}
 
+	/* drop the installed refs cache */
+	g_clear_pointer (&self->installed_refs, g_ptr_array_unref);
+
 	/* manually do this in case we created the first appstream file */
 	g_rw_lock_reader_lock (&self->silo_lock);
 	if (self->silo != NULL)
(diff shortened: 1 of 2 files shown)


andrunko:
Force pushed to cleanup the cached installed_refs on finalize.

plugins/flatpak/gs-flatpak.c
```diff
@@ -3569,6 +3576,7 @@ gs_flatpak_finalize (GObject *object)
 
 	g_free (self->id);
 	g_object_unref (self->installation);
+	g_ptr_array_unref (self->installed_refs);
```
pwithnall: It might be possible for `self->installed_refs == NULL` at this point (if there are no apps to refine, for example), so this should be `g_clear_pointer (&self->installed_refs, g_ptr_array_unref)`

andrunko: Good catch, updated.

andrunko:
> Looks good. Where’s the upstream version?

Wanted to get a downstream review first, will push tomorrow.

plugins/flatpak/gs-flatpak.c
```diff
@@ -1889,14 +1892,17 @@ gs_flatpak_refine_app_state_unlocked (GsFlatpak *self,
 		return FALSE;
 
 	/* find the app using the origin and the ID */
-	refs = flatpak_installation_list_installed_refs (self->installation,
-							 cancellable, error);
-	if (refs == NULL) {
+	if (self->installed_refs == NULL) {
+		self->installed_refs = flatpak_installation_list_installed_refs (self->installation,
```
uajain: I think arch-wise/upstream-wise, this re-cache-ing bit go in :
1) gs_plugin_setup (i.e. initializing)
2) whenever the cache is invalidated (`gs_plugin_flatpak_changed_cb`)

It feels a bit weird to me to query the installed_refs while refining an app, so we can additionally improve on that too.
+1 on the approach overall though

pwithnall:
> Wanted to get a downstream review first, will push tomorrow.

I think it would be faster to create the upstream and downstream MRs at the same time. That removes a whole day of latency on getting an initial review from upstream.

wjt:
Richard Hughes is in the same room as me and pwithnall today.

andrunko:
> > Wanted to get a downstream review first, will push tomorrow.
> 
> I think it would be faster to create the upstream and downstream MRs at the same time. That removes a whole day of latency on getting an initial review from upstream.

Well, when possible I agree that should be the preferred approach, but I was about to finish the day and had already tested/built the changes downstream, so wanted to get it out for review/opinions asap. Will prepare an upstream MR next.

plugins/flatpak/gs-flatpak.c
andrunko: Thought about that but in the end I tried to keep the current approach of retrying to get the installed refs on refine as it may fail for some reason.

andrunko:
Force pushed to also invalidate the installed refs cache on `gs_flatpak_refresh()` (which also already drops the installaction cache). This fixes an issue with `gs-self-test-flatpak`.

andrunko:
The upstream PR: https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/349

andrunko:
Change was merged upstream, force pushed to update commit msg with a "cherry picked from" comment.

PR #512, opened 2019-11-06
transaction: Sync main app's state if a related ref is installed

uajain:
In T28302, due to encyclopedia split between
com.endlessm.encyclopedia.en and com.endlessm.encyclopedia.en.Content,
old versions of encyclopedia (under one ref) didn't get the
content ref when they tried to upgrade the app. Hence, encyclopedia
broke for those users as there was no content ref.

The content ref is a related ref. Flatpak changes [1] can detect if
such a related ref is missing on the installation and therefore,
can mark the app as updatable again.

We need to make sure that gnome-software can be aware of this state
and sync the main app and related app's states accordingly for UI/UX.

[1] https://github.com/endlessm/flatpak/pull/197

https://phabricator.endlessm.com/T28302

### DIFF
--- a/plugins/flatpak/gs-flatpak-transaction.c
+++ b/plugins/flatpak/gs-flatpak-transaction.c
@@ -62,6 +62,41 @@ gs_flatpak_transaction_set_no_deploy (FlatpakTransaction *transaction, gboolean 
 	g_object_notify (G_OBJECT (self), "no-deploy");
 }
 
+/* Checks if a ref is a related ref to one of the installed ref.
+ * If yes, return the GsApp corresponding to the installed ref,
+ * NULL otherwise.
+ */
+static GsApp *
+get_installed_main_app_of_related_ref (FlatpakTransaction          *transaction,
+                                       FlatpakTransactionOperation *operation)
+{
+	FlatpakInstallation *installation = flatpak_transaction_get_installation (transaction);
+	const gchar *remote = flatpak_transaction_operation_get_remote (operation);
+	const gchar *op_ref = flatpak_transaction_operation_get_ref (operation);
+	GsFlatpakTransaction *self = GS_FLATPAK_TRANSACTION (transaction);
+	g_autoptr(GList) keys = NULL;
+
+	if (g_str_has_prefix (op_ref, "app/"))
+		return NULL;
+
+	keys = g_hash_table_get_keys (self->refhash);
+	for (GList *l = keys; l != NULL; l = l->next) {
+		g_autoptr(GPtrArray) related_refs = NULL;
+		related_refs = flatpak_installation_list_installed_related_refs_sync (installation, remote,
+										      l->data, NULL, NULL);
+		if (related_refs == NULL)
+			continue;
+
+		for (guint i = 0; i < related_refs->len; i++) {
+			g_autofree gchar *rref = flatpak_ref_format_ref (g_ptr_array_index (related_refs, i));
+			if (g_strcmp0 (rref, op_ref) == 0) {
+				return g_hash_table_lookup (self->refhash, l->data);
+			}
+		}
+	}
+	return NULL;
+}
+
 GsApp *
 gs_flatpak_transaction_get_app_by_ref (FlatpakTransaction *transaction, const gchar *ref)
 {
@@ -256,6 +291,8 @@ _transaction_operation_done (FlatpakTransaction *transaction,
 			     FlatpakTransactionResult details)
 {
 	GsFlatpakTransaction *self = GS_FLATPAK_TRANSACTION (transaction);
+	GsApp *main_app = NULL;
+
 	/* invalidate */
 	GsApp *app = _transaction_operation_get_app (operation);
 	if (app == NULL) {
@@ -265,6 +302,26 @@ _transaction_operation_done (FlatpakTransaction *transaction,
 	}
 	switch (flatpak_transaction_operation_get_operation_type (operation)) {
 	case FLATPAK_TRANSACTION_OPERATION_INSTALL:
+		/* Handle special snowflake where "should-download" related refs for an installed ref
+		 * goes missing. In that case, libflatpak marks the main app ref as updatable
+		 * and then FlatpakTransaction resolves one of its ops to install the related ref(s).
+		 *
+		 * We can depend on libflatpak till here. Since, libflatpak returns the main app
+		 * ref as updatable (instead of the related ref), we need to sync the main app's
+		 * state for UI/UX.
+		 *
+		 * Map the current op's ref (which is related ref) to its main app ref (which is
+		 * currently shown in the UI) and set the state of the main GsApp object back
+		 * to INSTALLED here.
+		 *
+		 * This detection whether a related ref belongs to a main ref is quite sub-optimal as
+		 * of now.
+		 */
+		main_app = get_installed_main_app_of_related_ref (transaction, operation);
+		if (main_app != NULL)
+			gs_app_set_state (main_app, AS_APP_STATE_INSTALLED);
+		gs_app_set_state (app, AS_APP_STATE_INSTALLED);
+		break;
 	case FLATPAK_TRANSACTION_OPERATION_INSTALL_BUNDLE:
 		gs_app_set_state (app, AS_APP_STATE_INSTALLED);
 		break;
--- a/plugins/flatpak/gs-plugin-flatpak.c
+++ b/plugins/flatpak/gs-plugin-flatpak.c
@@ -965,6 +965,9 @@ gs_plugin_flatpak_update (GsPlugin *plugin,
 			gs_flatpak_error_convert (error);
 			return FALSE;
 		}
+
+		/* Add the update applist for easier lookup */
+		gs_flatpak_transaction_add_app (transaction, app);
 	}
 
 	/* run transaction */


plugins/flatpak/gs-flatpak-transaction.c
```diff
@@ -265,6 +302,25 @@ _transaction_operation_done (FlatpakTransaction *transaction,
 	}
 	switch (flatpak_transaction_operation_get_operation_type (operation)) {
 	case FLATPAK_TRANSACTION_OPERATION_INSTALL:
+		/* Handle special snowflake where "should-download" related refs for an installed ref
+		 * goes missing. In that case, libflatpak marks the main app ref as updatable
+		 * and then FlatpakTransaction resolves one of it's ops to install the related ref(s).
+		 *
+		 * We can depend on libflatpak till here. Since, libflatpak returns the main app
+		 * ref as updatable (instead of the related ref), we need to sync the main app's
+		 * statefor UI/UX.
```
andrunko: nit: missing space after `state`.

plugins/flatpak/gs-flatpak-transaction.c
```diff
@@ -265,6 +302,25 @@ _transaction_operation_done (FlatpakTransaction *transaction,
...
+		 *
+		 * Map the current op's ref (which is related ref) to it's main app ref(which is currently
```
andrunko: nit: "which is *a* related ref", also pending space before "(which is currently".

plugins/flatpak/gs-flatpak-transaction.c
```diff
@@ -265,6 +302,25 @@ _transaction_operation_done (FlatpakTransaction *transaction,
...
+		 * shown in the UI) and set the state of the main GsApp object back to INSTALLED here.
+		 *
+		 * This detection whether a related ref belongs to a main ref is quite sub-optimal as
+		 * of now.
+		 */
+		main_app = check_if_related_ref_is_installed (transaction, operation);
```
andrunko: Not sure if the question is valid, but what if for some reason the related ref gets installed before the main app? Or is the order guaranteed? If so, do we need to depend on some specific flatpak version for that to work?

plugins/flatpak/gs-plugin-flatpak.c
```diff
@@ -965,6 +965,9 @@ gs_plugin_flatpak_update (GsPlugin *plugin,
 			gs_flatpak_error_convert (error);
 			return FALSE;
 		}
+
+		/* Add the update applist for easier lookup */
+		gs_flatpak_transaction_add_app (transaction, app);
```
andrunko: Could you clarify why this is needed? Not sure I understand it.

plugins/flatpak/gs-flatpak-transaction.c
```diff
@@ -62,6 +62,41 @@ gs_flatpak_transaction_set_no_deploy (FlatpakTransaction *transaction, gboolean
 	g_object_notify (G_OBJECT (self), "no-deploy");
 }
 
+/* Checks if a ref is a related ref to one of the installed ref.
+ * If yes, return the GsApp corresponding to the installed ref,
+ * NULL otherwise.
+ */
+static GsApp *
+check_if_related_ref_is_installed (FlatpakTransaction *transaction,
+				    FlatpakTransactionOperation *operation)
```
andrunko: I think the method name is confusing given it returns the main app ref for a related ref. I would probably rename it to something like `get_main_app_for_related_ref`.

Also nit: missing space before `FlatpakTransactionOperation`.

plugins/flatpak/gs-plugin-flatpak.c
uajain: it's a flatpak transaction local cache which is useful to lookup which GsApp objects are involved in the current Transaction. The cache is populated with whatever `flatpak_transaction_list_installed_refs_for_update` returns. It's useful as in to reduce to scope of which apps we are concerned about(i.e. needing an update, runtime missing?, extensions missing?) rather than look up all installed flatpak apps and iterating over them to query that info.

plugins/flatpak/gs-flatpak-transaction.c
uajain: This is invalid as far as these changes are concerned. We expect the main-app/ref to be installed already, otherwise, how would we know if the app is updatable at the first place :)

Regarding the ordering, I think we should  just let FlatpakTransaction do it's magic

plugins/flatpak/gs-flatpak-transaction.c
```diff
@@ -62,6 +62,41 @@ gs_flatpak_transaction_set_no_deploy (FlatpakTransaction *transaction, gboolean
...
+	FlatpakInstallation *installation = flatpak_transaction_get_installation (transaction);
+	const gchar *remote = flatpak_transaction_operation_get_remote (operation);
+	const gchar *op_ref = flatpak_transaction_operation_get_ref (operation);
+	GsFlatpakTransaction *self = GS_FLATPAK_TRANSACTION (transaction);
+	GList *keys;
+
+	if (g_str_has_prefix (op_ref, "app/"))
+		return NULL;
+
+	keys = g_hash_table_get_keys (self->refhash);
```
andrunko: Missed that during initial review, `keys` is leaking here.

plugins/flatpak/gs-flatpak-transaction.c
```diff
@@ -62,6 +62,41 @@ gs_flatpak_transaction_set_no_deploy (FlatpakTransaction *transaction, gboolean
...
+		g_autoptr(GPtrArray) related_refs = NULL;
+		related_refs = flatpak_installation_list_installed_related_refs_sync (installation, remote,
+										      l->data, NULL, NULL);
+		if (related_refs == NULL)
+			continue;
+
+		for (guint i = 0; i < related_refs->len; i++) {
+			g_autofree gchar *rref = flatpak_ref_format_ref (g_ptr_array_index (related_refs, i));
+			if (g_strcmp0 (rref, op_ref) == 0) {
+				return g_hash_table_lookup (self->refhash, l->data);
```
andrunko: Also I believe you need to ref the result here as you are using g_autoptr(GsApp) below?

uajain: It's already ref-ed when it is inserted in the hashtable

plugins/flatpak/gs-flatpak-transaction.c
uajain: Addressed in latest iteration

plugins/flatpak/gs-flatpak-transaction.c
andrunko: Yes, but it is freed when the hashtable is unrefed on finalize - so you will end up with a double free here. You need to either ref it here or dont use g_autoptr for main_app in `gs_flatpak_transaction_get_app_by_ref()`.

uajain: Yeah, I missed that _finalize bit it seems. Updated

pwithnall:
Is there an upstream version of this MR, or is it not relevant upstream?

plugins/flatpak/gs-flatpak-transaction.c
```diff
@@ -265,6 +302,26 @@ _transaction_operation_done (FlatpakTransaction *transaction,
...
+		 *
+		 * We can depend on libflatpak till here. Since, libflatpak returns the main app
+		 * ref as updatable (instead of the related ref), we need to sync the main app's
+		 * state for UI/UX.
+		 *
+		 * Map the current op's ref (which is related ref) to it's main app ref (which is
+		 * currently shown in the UI) and set the state of the main GsApp object back
+		 * to INSTALLED here.
+		 *
+		 * This detection whether a related ref belongs to a main ref is quite sub-optimal as
```
mwleeds: What's sub-optimal about it?

plugins/flatpak/gs-flatpak-transaction.c
```diff
@@ -265,6 +302,26 @@ _transaction_operation_done (FlatpakTransaction *transaction,
...
 	case FLATPAK_TRANSACTION_OPERATION_INSTALL:
+		/* Handle special snowflake where "should-download" related refs for an installed ref
+		 * goes missing. In that case, libflatpak marks the main app ref as updatable
+		 * and then FlatpakTransaction resolves one of it's ops to install the related ref(s).
+		 *
+		 * We can depend on libflatpak till here. Since, libflatpak returns the main app
+		 * ref as updatable (instead of the related ref), we need to sync the main app's
+		 * state for UI/UX.
+		 *
+		 * Map the current op's ref (which is related ref) to it's main app ref (which is
```
mwleeds: nit: should be "its"

plugins/flatpak/gs-flatpak-transaction.c
```diff
@@ -265,6 +302,26 @@ _transaction_operation_done (FlatpakTransaction *transaction,
 	}
 	switch (flatpak_transaction_operation_get_operation_type (operation)) {
 	case FLATPAK_TRANSACTION_OPERATION_INSTALL:
+		/* Handle special snowflake where "should-download" related refs for an installed ref
+		 * goes missing. In that case, libflatpak marks the main app ref as updatable
+		 * and then FlatpakTransaction resolves one of it's ops to install the related ref(s).
```
mwleeds: nit: should be "its"

plugins/flatpak/gs-flatpak-transaction.c
uajain: I mean it's still querying related refs of the installed app, comparing with operation's current's ref and making sure the extension ref belong to the installed app's ref ... 

Seems like stitched-code.

mwleeds: Yeah I suppose it could be cleaner, perhaps if the app object had the related refs directly associated with it.

PR #513, opened 2019-11-14
eos-updater: reboot when upgrade triggered

wjt:
Untested since I don't have an update pending.

- [ ] Test tomorrow.

https://phabricator.endlessm.com/T28602


wjt:
Unnecessary, the surrounding code in `gs-updates-page.c` has identical logic.

PR #514, opened 2019-11-15
Backport leak fixes from upstream

pwithnall:
Trivial backport of https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/357.

No associated Phabricator task.

PR #515, opened 2019-11-19
Follow up fixes to #514

uajain:
https://phabricator.endlessm.com/T28632


pwithnall:
I’ve proposed an alternative fix in #517. I’ve tested it, but it could do with some close review.

PR #517, opened 2019-11-19
T28632 Fix use of custom CSS quarks

pwithnall:
This is a partial revert of #514, replacing it with https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/366 (trivial backports). I’ve kept the existing `fixup!` commits which apply the CSS API change to our custom widgets, as the CSS API change has stayed the same.

The main difference from #514 is the switch from attaching the `GtkCssProvider` containing custom CSS to the entire screen, to instead attaching it to just the widget which needs the custom CSS. This prevents newer custom CSS (with the same class name as older CSS) from overriding the older stuff.

Supersedes #515.

https://phabricator.endlessm.com/T28632


uajain:
This is already merged upstream. Feel free to merge it OR I can take a look and test/merge tomorrow.

Thanks for the alternative upstream-able fix.

PR #518, opened 2019-11-22
transaction: Sync installed apps' state missing runtime is re-installed

uajain:
Built on top of solution of https://github.com/endlessm/gnome-software/pull/512
(See also base branch)

---
Flatpak changes [1] will start marking apps as updatable if their
runtime component is missing on the installation. These apps will
be shown as updatable in gnome-software updates panel but in the
background they are meant to fetch their missing runtime.
 
Since many apps can target one runtime, the absence of that runtime
can make many apps show up in the updates panel. In that case,
check if the runtime is downloaded → installed and set all
apps' state back to "installed". This solves the case
if users chooses to click "Update All".
    
[1] https://github.com/flatpak/flatpak/pull/3204

https://phabricator.endlessm.com/T27077

---
#### Briefly tested with changes with [1] on following updates navigation paths:
* Single Manual update :heavy_check_mark: 
* Autoupdates behavior - Unistalled various runtimes, turned on autoupdates and used the computer for sometime. All the apps were updated(and functional) and appropriately `Updates` tab showed all updates to be done.  :heavy_check_mark: 
*  Have many apps targeting one `$runtime` :heavy_check_mark: 
   * `flatpak uninstall --force-remove $runtime`
   * Verify all apps targeting `$runtime` show up in updates tab 
   * Click `Update All` in updates tab - All apps goes back to installed state correctly.
* Have many apps targeting one `$runtime` :fish:-y
   * Click on **one** of the app targeting that runtime - Installs the runtime and goes back to installed state. However, other apps still stay on the updates page. They stay until the updates page is refreshed or gnome-software is restarted.
   * Click on **one** of the app(A) targeting that runtime, few seconds later, click on the one of the other app(B) targeting that runtime. A goes back to installed state after runtime install, whereas B is stuck in _INSTALLING state forever (until gnome-software restart) - Not sure what we can do there.

### DIFF
--- a/plugins/flatpak/gs-flatpak-transaction.c
+++ b/plugins/flatpak/gs-flatpak-transaction.c
@@ -62,6 +62,58 @@ gs_flatpak_transaction_set_no_deploy (FlatpakTransaction *transaction, gboolean 
 	g_object_notify (G_OBJECT (self), "no-deploy");
 }
 
+/* Sets installed app(s) back to installed state. Flatpak can return apps as updatable
+ * (for installing a missing runtime); if it is detected that the runtime was missing
+ * at the first place.
+ *
+ * We can determine whether a GsApp is only being updated due to a missing runtime, by checking
+ * if the current operation's ref is the GsApp's runtime and also the GsApp is already deployed.
+ */
+static void
+set_installed_app_state_if_missing_runtime_is_installed (FlatpakTransaction          *transaction,
+							  FlatpakTransactionOperation *operation)
+{
+	GsFlatpakTransaction *self = GS_FLATPAK_TRANSACTION (transaction);
+	FlatpakInstallation *installation = flatpak_transaction_get_installation (transaction);
+	const gchar *op_ref = flatpak_transaction_operation_get_ref (operation);
+	g_autoptr(GList) apps = NULL;
+
+	if (g_str_has_prefix (op_ref, "app/"))
+		return;
+
+	apps = g_hash_table_get_values (self->refhash);
+	for (GList *l = apps; l != NULL; l = l->next) {
+		GsApp *app = l->data;
+		GsApp *app_runtime;
+		g_autofree gchar *app_runtime_ref = NULL;
+
+		app_runtime = gs_app_get_runtime (app);
+		if (app_runtime == NULL)
+			continue;
+
+		app_runtime_ref = gs_flatpak_app_get_ref_display (app_runtime);
+		if (app_runtime_ref != NULL &&
+		    g_strcmp0 (app_runtime_ref, op_ref) == 0) {
+			g_autoptr(FlatpakInstalledRef) app_ref = NULL;
+			g_autoptr(GBytes) metadata = NULL;
+
+			app_ref = flatpak_installation_get_installed_ref (installation,
+									  FLATPAK_REF_KIND_APP,
+									  gs_flatpak_app_get_ref_name (app),
+									  gs_flatpak_app_get_ref_arch (app),
+									  gs_flatpak_app_get_ref_branch (app),
+									  NULL, NULL);
+			if (app_ref == NULL)
+				continue;
+
+			metadata = flatpak_installed_ref_load_metadata (app_ref, NULL, NULL);
+			/* This makes sure that the app is already deployed. */
+			if (metadata != NULL)
+				gs_app_set_state (app, AS_APP_STATE_INSTALLED);
+		}
+	}
+}
+
 /* Checks if a ref is a related ref to one of the installed ref.
  * If yes, return the GsApp corresponding to the installed ref,
  * NULL otherwise.
@@ -320,6 +372,14 @@ _transaction_operation_done (FlatpakTransaction *transaction,
 		main_app = get_installed_main_app_of_related_ref (transaction, operation);
 		if (main_app != NULL)
 			gs_app_set_state (main_app, AS_APP_STATE_INSTALLED);
+
+		/* Do the same as above but if the main app is missing its runtime.
+		 * Multiple GsApp can depend on one (missing) runtime, hence set state to "installed"
+		 * state for all those apps too.
+		 */
+		set_installed_app_state_if_missing_runtime_is_installed (transaction, operation);
+
+		/* For all other trivial cases. */
 		gs_app_set_state (app, AS_APP_STATE_INSTALLED);
 		break;
 	case FLATPAK_TRANSACTION_OPERATION_INSTALL_BUNDLE:


mwleeds:
Oh! This is targeting the branch of https://github.com/endlessm/gnome-software/pull/512. I didn't understand that at first despite your comment at the top.

plugins/flatpak/gs-flatpak-transaction.c
```diff
@@ -62,6 +62,52 @@ gs_flatpak_transaction_set_no_deploy (FlatpakTransaction *transaction, gboolean
 	g_object_notify (G_OBJECT (self), "no-deploy");
 }
 
+/* Sets installed app(s) back to installed state. Since libflatpak can return apps as updatable
+ * for re-installing a runtime if its is detected that the runtime was missing at the first place.
```
mwleeds: Typo here

plugins/flatpak/gs-flatpak-transaction.c
```diff
@@ -62,6 +62,52 @@ gs_flatpak_transaction_set_no_deploy (FlatpakTransaction *transaction, gboolean
...
+		return;
+
+	apps = g_hash_table_get_values (self->refhash);
+	for (GList *l = apps; l != NULL; l = l->next) {
+		GsApp *app = l->data;
+		GsApp *app_runtime;
+		g_autofree gchar *app_runtime_ref = NULL;
+
+		app_runtime = gs_app_get_runtime (app);
+		app_runtime_ref = gs_flatpak_app_get_ref_display (app_runtime);
```
mwleeds: Don't we need `if (app_runtime == NULL) continue;` here?

plugins/flatpak/gs-flatpak-transaction.c
```diff
@@ -62,6 +62,52 @@ gs_flatpak_transaction_set_no_deploy (FlatpakTransaction *transaction, gboolean
...
+			g_autoptr(FlatpakInstalledRef) app_ref = NULL;
+			g_autoptr(GBytes) metadata = NULL;
+
+			app_ref = flatpak_installation_get_installed_ref (installation,
+									  FLATPAK_REF_KIND_APP,
+									  gs_flatpak_app_get_ref_name (app),
+									  gs_flatpak_app_get_ref_arch (app),
+									  gs_flatpak_app_get_ref_branch (app),
+									  NULL, NULL);
+			metadata = flatpak_installed_ref_load_metadata (app_ref, NULL, NULL);
```
mwleeds: Need to check if `app_ref == NULL` here

plugins/flatpak/gs-flatpak-transaction.c
```diff
@@ -62,6 +62,52 @@ gs_flatpak_transaction_set_no_deploy (FlatpakTransaction *transaction, gboolean
...
+			/* This makes sure that the app is already deployed. */
+			if (metadata != NULL)
+				gs_app_set_state (app, AS_APP_STATE_INSTALLED);
+		}
+	}
+	return;
```
mwleeds: nit: don't really need this `return` right?

plugins/flatpak/gs-flatpak-transaction.c
```diff
@@ -62,6 +62,52 @@ gs_flatpak_transaction_set_no_deploy (FlatpakTransaction *transaction, gboolean
 	g_object_notify (G_OBJECT (self), "no-deploy");
 }
 
+/* Sets installed app(s) back to installed state. Since libflatpak can return apps as updatable
```
mwleeds: Could say "back to installed state when the associated runtime has been installed"

plugins/flatpak/gs-flatpak-transaction.c
```diff
@@ -62,6 +62,52 @@ gs_flatpak_transaction_set_no_deploy (FlatpakTransaction *transaction, gboolean
...
+ * for re-installing a runtime if its is detected that the runtime was missing at the first place.
+ *
+ * We can determine whether is GsApps is only being updated due to a missing runtime, by checking
```
mwleeds: What about the case where an installed app has an update available, and its runtime also has an update available? Are we prematurely setting the app to the installed state in that case?

plugins/flatpak/gs-flatpak-transaction.c
```diff
@@ -320,6 +366,14 @@ _transaction_operation_done (FlatpakTransaction *transaction,
 		main_app = get_installed_main_app_of_related_ref (transaction, operation);
 		if (main_app != NULL)
 			gs_app_set_state (main_app, AS_APP_STATE_INSTALLED);
+
+		/* Do the same as above but if the main app is missing it's runtime.
```
mwleeds: nit: should be "its"

plugins/flatpak/gs-flatpak-transaction.c
uajain: Good to think more on cases like this. 
On top of my head, runtime is missing at the first place in these scenarios, hence having runtime-update seems unlikely.

plugins/flatpak/gs-flatpak-transaction.c
uajain: There is a `app_runtime !=NULL ` just below this

mwleeds: But is it safe to call `gs_flatpak_app_get_ref_display(NULL)`?

plugins/flatpak/gs-flatpak-transaction.c
mwleeds: Oh, this function is only called if the operation is an install, not an update, so the situation I asked about isn't relevant. But what about when an app and its runtime are being installed? Wouldn't we prematurely set the app to `AS_APP_STATE_INSTALLED`, since the runtime operation is ordered before the app one?

uajain: It should not set the app state prematurely for  (app+runtime) install 
a) runtimes are installed before the app, no? 
b) we only set the state iff the app is deployed at the first place. `flatpak_installed_ref_load_metadata` checks if the ref has a deploy file on the installation or not

mwleeds: Good point, I forgot about the `flatpak_installed_ref_load_metadata()` call.

plugins/flatpak/gs-flatpak-transaction.c
```diff
@@ -62,6 +62,58 @@ gs_flatpak_transaction_set_no_deploy (FlatpakTransaction *transaction, gboolean
...
+			g_autoptr(GBytes) metadata = NULL;
+
+			app_ref = flatpak_installation_get_installed_ref (installation,
+									  FLATPAK_REF_KIND_APP,
+									  gs_flatpak_app_get_ref_name (app),
+									  gs_flatpak_app_get_ref_arch (app),
+									  gs_flatpak_app_get_ref_branch (app),
+									  NULL, NULL);
+			if (app_ref == NULL)
+				return;
```
mwleeds: Shouldn't this be `continue;`? In case `app` is not installed.

uajain: Right, addressed.

PR #519, opened 2019-11-22
eos-blacklist: remove Eclipse and 3 other apps

wjt:
https://phabricator.endlessm.com/T28707


wjt:
> I'm assuming from the comment on the Phabricator task that you ran each of these!

I ran Eclipse, and indeed the new Flatpak is more functional than the old one!

None of these 4 apps have ever been published on Flathub under these app IDs, and only Eclipse was published in eos-apps. So I haven't run the other 3 since they don't exist :-)

PR #521, opened 2019-11-27
flatpak: Only ignore estimated progress if >10%

ptomato:
Flatpak may set the "estimating" flag on a progress report if the value
is bogus, or it may set it if it is downloading metadata which it
arbitrarily decides comprises the first 5% of an install operation.
Ignoring estimated progress therefore means we get no progress changes
for the first 10% of an app install (installing the app and its locale
extension). Instead, only ignore the estimated progress if it is an
obviously bogus value of >10%.

https://phabricator.endlessm.com/T28764

### DIFF
--- a/plugins/flatpak/gs-flatpak-transaction.c
+++ b/plugins/flatpak/gs-flatpak-transaction.c
@@ -175,8 +175,18 @@ _transaction_progress_changed_cb (FlatpakTransactionProgress *progress,
 {
 	GsApp *app = GS_APP (user_data);
 	guint percent = flatpak_transaction_progress_get_progress (progress);
-	if (flatpak_transaction_progress_get_is_estimating (progress))
-		return;
+	if (flatpak_transaction_progress_get_is_estimating (progress)) {
+		/* "Estimating" happens while fetching the metadata, which
+		 * flatpak arbitrarily decides happens during the first 5% of
+		 * each operation. Often there are two install operations,
+		 * for the flatpak and its locale data.
+		 * However, "estimating" may also mean bogus values. We have to
+		 * arbitrarily decide whether to show this value to the user. */
+		if (percent > 10) {
+			g_debug ("Ignoring estimated progress of %u%%", percent);
+			return;
+		}
+	}
 	if (gs_app_get_progress (app) != 100 &&
 	    gs_app_get_progress (app) > percent) {
 		g_warning ("ignoring percentage %u%% -> %u%% as going down...",


plugins/flatpak/gs-flatpak-transaction.c
```diff
@@ -175,8 +175,18 @@ _transaction_progress_changed_cb (FlatpakTransactionProgress *progress,
...
-		return;
+	if (flatpak_transaction_progress_get_is_estimating (progress)) {
+		/* "Estimating" happens while fetching the metadata, which
+		 * flatpak arbitrarily decides happens during the first 5% of
+		 * each operation. Often there are two install operations,
+		 * for the flatpak and its locale data.
+		 * However, "estimating" may also mean bogus values. We have to
+		 * arbitrarily decide whether to show this value to the user. */
+		if (percent > 10) {
+			g_warning("ignoring estimated progress of %u%%", percent);
```
pwithnall:
```suggestion
			g_warning ("Ignoring estimated progress of %u%%", percent);
```
(nitpick)

ptomato:
> Is there an accompanying upstream MR for this?
>
> Could we improve flatpak so it doesn’t produce bogus progress values during the ‘estimating’ phase? What causes them?

To be honest I'm not sure this should go upstream without improving flatpak in the way you suggest. Upstream gnome-software have presumably made a choice to prefer reporting only accurate progress information from flatpak, at the cost of snappiness in the UI. I think we should make the opposite choice and I'm not opposed to it being a hacky solution like this if it has a clear benefit for users' experience.

Making these improvements to flatpak seems like a large, multi-week effort even for someone already familiar with flatpak's codebase; to quote myself on the Phabricator ticket, the reason why these things happen is probably due to the difficulty in calculating a simple number for the progress when extra-data, deltas, and remote appstream processing are involved, which leads to these hacks in flatpak: https://github.com/endlessm/flatpak/blob/master/common/flatpak-utils.c#L6184-L6405 To improve this, I think it would be necessary to refactor flatpak to be able to calculate one simple number (total bytes to download) and make the progress bar climb according to that. Or if that's really not possible, then split the progress reporting so it works per operation (fetch metadata, _N_ × install, cleanup) and let the UI sort out how to present it to the user.

pwithnall:
> To be honest I'm not sure this should go upstream without improving flatpak in the way you suggest. …

I think we should ask. I suspect upstream will be open to a similar compromise as us, given that they likely have the same usability issues. :)

> Making these improvements to flatpak seems like a large, multi-week effort even for someone already familiar with flatpak's codebase; …

Sure, but we could at least file an issue about it because it’s likely affecting others, and maybe someone else will choose to work on it in the mean time.

ptomato:
Upstream merge request: https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/376

It seems the problem with progress reporting is just closed whenever someone reports it, so I don't think there's any point in reporting it again: https://github.com/flatpak/flatpak/issues/2768 https://github.com/flatpak/flatpak/issues/3175

Also relevant but depends on ostree: https://github.com/flatpak/flatpak/issues/404

mwleeds:
> It seems the problem with progress reporting is just closed whenever someone reports it, so I don't think there's any point in reporting it again: [flatpak/flatpak#2768](https://github.com/flatpak/flatpak/issues/2768) [flatpak/flatpak#3175](https://github.com/flatpak/flatpak/issues/3175)
> 

The reasoning on closing those was that it should be an ostree issue if ostree is the one sending bad progress information. But if Flatpak's progress estimation is bad, and if it could be improved in areas that are not dependent on ostree like extra-data, that seems like a legitimate Flatpak issue.

ptomato:
That's certainly true, but on the other hand I don't think that having more accurate extra-data information would help that much if we didn't have accurate ostree pull information.

Mainly my opinion (admittedly after not working on these tools for very long) is what I posted in the Phabricator ticket:
> Note that the reason why those things happen is probably due to the difficulty in calculating a simple number for the progress when extra-data, deltas, and remote appstream processing are involved, which leads to these hacks in flatpak: https://github.com/endlessm/flatpak/blob/master/common/flatpak-utils.c#L6184-L6405
> To improve this, I think it would be necessary to refactor flatpak to be able to calculate one simple number (total bytes to download) and make the progress bar climb according to that.

I think flatpak even already does calculate a total number of extra data bytes (although no idea how accurate it is) but it's not useful unless we can get a number of bytes from ostree.

ptomato:
I've updated this following the upstream review comments, it looks like Richard didn't have a problem with the patch itself, just a minor comment. As discussed with @andrunko, therefore merging this.

PR #523, opened 2020-01-23
T29074 flatpak bundle install fix

mwleeds:
https://phabricator.endlessm.com/T29074


plugins/flatpak/tests/build.py
```diff
@@ -70,6 +71,13 @@ def copy_repo(srcdir, destdir):
         shutil.rmtree(destdir_repo)
     shutil.copytree(srcdir_repo, destdir_repo)
 
+# use git master where available
+local_checkout = '/home/hughsie/Code/flatpak'
+if os.path.exists(local_checkout):
+    flatpak_cmd = os.path.join(local_checkout, 'flatpak')
+else:
+    flatpak_cmd = 'flatpak'
```
andrunko: I dont like the fact that `flatpak_cmd` is a global var used by `build_flatpak` and `build_flatpak_bundle`, but details.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2659,11 +2677,129 @@ gs_flatpak_refine_appstream_release (XbNode *component, GsApp *app)
...
+			     GS_PLUGIN_ERROR,
+			     GS_PLUGIN_ERROR_INVALID_FORMAT,
+			     "unable to decompress appstream data");
+		return FALSE;
+	}
+	stream_data = g_converter_input_stream_new (stream_gz,
+						    G_CONVERTER (decompressor));
+
+	appstream = g_input_stream_read_bytes (stream_data,
+					       0x100000, /* 1Mb */
```
andrunko: ooc, why 1Mb and what happens if the appstream is bigger than that? I know the code was already there but just wondering.

plugins/flatpak/tests/build.py
mwleeds: In the upstream PR there's a commit that deletes these few lines.

plugins/flatpak/gs-flatpak.c
mwleeds: I'm not sure, there's no explanation in the commit that introduced it: https://gitlab.gnome.org/GNOME/gnome-software/commit/f4fee54408

But 1 Mb would be a lot of text.

mwleeds:
I updated this so it matches the current state of the upstream MR. The only conflicts during cherry-picking were in "flatpak: Fix searching for installed bundles".

mwleeds:
I added a commit to fix loading icons for apps installed from bundles, and this is now ready for review.

plugins/flatpak/gs-plugin-flatpak.c
```diff
@@ -728,6 +728,14 @@ gs_plugin_app_remove (GsPlugin *plugin,
 		gs_flatpak_error_convert (error);
 		return FALSE;
 	}
+
+	/* If the app is still in REMOVING it's probably because it was installed
+	 * from a .flatpak bundle and is therefore no longer available (unless the
+	 * bundle is still around).
+	 */
+	if (gs_app_get_state (app) == AS_APP_STATE_REMOVING)
```
andrunko: Maybe also check whether `app` comes from a bundle here?

plugins/flatpak/gs-flatpak.c
```diff
@@ -2755,6 +2756,19 @@ gs_flatpak_refine_appstream_from_bytes (GsFlatpak *self,
...
+	if (installed_ref != NULL) {
+		g_autoptr(XbBuilderNode) info = NULL;
+		g_autofree char *icon_prefix = NULL;
+
+		info = xb_builder_node_insert (NULL, "info", NULL);
+		xb_builder_node_insert_text (info, "scope", as_app_scope_to_string (self->scope), NULL);
+		icon_prefix = g_build_filename (flatpak_installed_ref_get_deploy_dir (installed_ref),
+						"files", "share", "app-info", "icons", "flatpak", NULL);
+		xb_builder_node_insert_text (info, "icon-prefix", icon_prefix, NULL);
+		xb_builder_source_set_info (source, info);
```
andrunko: Is it possible that the appstream data already has icon set?

mwleeds: Yes, the appstream data will have an `<icon>` but not an icon prefix, which is machine specific.

plugins/flatpak/gs-plugin-flatpak.c
mwleeds: Agree that's a good idea

mwleeds: But sadly we can't check if `gs_flatpak_app_get_file_kind (app) == GS_FLATPAK_APP_FILE_KIND_BUNDLE` because that's only true for the `GsApp` created when the bundle is opened, not the `GsApp` created for the installed app.

One thing we could check is if the associated remote has an empty `url`. But then there's the question of what would happen for bundles which include a URL. I think those should have the app picked up by `gs_flatpak_add_apps_from_xremote()` but it depends on if they serve appstream data so I'm not sure.

andrunko: I see - just wondering if this could break uninstallation for other cases, have you tested it? Does the Install button shows up right after an app (not bundle) gets uninstalled? Can you reinstall an app right after uninstalling it?

What happens in the UI when uninstalling a bundle?

mwleeds: This code is after the uninstall transaction completes and the remotes have been scanned again for apps (`gs_flatpak_refresh()`) so it shouldn't apply to apps that were not from bundles. Even so, I'll do some more testing and get back to you.

mwleeds: For an app installed from a bundle, there's a spinner while it's being removed, then that goes away and after a few seconds the details page is replaced by the message "Unable to find org.gnome.Fractal" which seems like reasonable behavior.

For an app installed from the Internet, the "Removing" spinner goes away and after a few seconds a "Download" button appears as you'd expect. (There is an upstream MR to make the spinner spin longer: https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/367)

mwleeds: > Can you reinstall an app right after uninstalling it?

Yes, for an app installed from the Internet, I was able to use the Download button successfully immediately after uninstalling.

andrunko: > For an app installed from a bundle, there's a spinner while it's being removed, then that goes away and after a few seconds the details page is replaced by the message "Unable to find org.gnome.Fractal" which seems like reasonable behavior.
> 
I guess that is reasonable, the other option would be to go back to the main page I guess, but that would probably make the user even more confused. Lets go with this as is and we can iterate later if needed.

PR #524, opened 2020-01-24
Set spinner only on app copying to usb

edward-mwangi:
https://phabricator.endlessm.com/T26219

### DIFF
--- a/lib/gs-plugin-loader.c
+++ b/lib/gs-plugin-loader.c
@@ -40,6 +40,7 @@ typedef struct
 	SoupSession		*soup_session;
 	GPtrArray		*file_monitors;
 	GsPluginStatus		 global_status_last;
+	GHashTable		*global_plugin_statuses;
 
 	GMutex			 pending_apps_mutex;
 	GsAppList		*pending_apps;
@@ -2262,6 +2263,7 @@ gs_plugin_loader_status_changed_cb (GsPlugin *plugin,
 		}
 		return;
 	}
+	g_hash_table_insert (priv->global_plugin_statuses, g_strdup (gs_app_get_id (app)), GUINT_TO_POINTER (status));
 
 	/* a specific app */
 	g_debug ("emitting %s(%s)",
@@ -3108,6 +3110,7 @@ gs_plugin_loader_finalize (GObject *object)
 	g_ptr_array_unref (priv->file_monitors);
 	g_hash_table_unref (priv->events_by_id);
 	g_hash_table_unref (priv->disallow_updates);
+	g_hash_table_unref (priv->global_plugin_statuses);
 
 	g_mutex_clear (&priv->pending_apps_mutex);
 	g_mutex_clear (&priv->events_by_id_mutex);
@@ -3274,6 +3277,7 @@ gs_plugin_loader_init (GsPluginLoader *plugin_loader)
 	for (i = 0; projects[i] != NULL; i++)
 		g_debug ("compatible-project: %s", projects[i]);
 	priv->compatible_projects = projects;
+	priv->global_plugin_statuses = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
 }
 
 /**
@@ -4193,4 +4197,26 @@ gs_plugin_loader_get_locale (GsPluginLoader *plugin_loader)
 {
 	GsPluginLoaderPrivate *priv = gs_plugin_loader_get_instance_private (plugin_loader);
 	return priv->locale;
+}
+
+gboolean
+gs_plugin_loader_app_copying (GsPluginLoader *plugin_loader, GsApp *app)
+{
+	GsPluginLoaderPrivate *priv = gs_plugin_loader_get_instance_private (plugin_loader);
+	if (app == NULL) {
+		return FALSE;
+	}
+	return GPOINTER_TO_UINT (g_hash_table_lookup (priv->global_plugin_statuses,
+				gs_app_get_id (app))) == GS_PLUGIN_STATUS_COPYING;
+}
+
+gboolean
+gs_plugin_loader_copy_queue_empty (GsPluginLoader *plugin_loader)
+{
+	GsPluginLoaderPrivate *priv = gs_plugin_loader_get_instance_private (plugin_loader);
+	g_autoptr(GList) statuses = g_hash_table_get_values (priv->global_plugin_statuses);
+	if (g_list_find (statuses, GUINT_TO_POINTER (GS_PLUGIN_STATUS_COPYING)) != NULL) {
+		return FALSE;
+	}
+	return TRUE;
 }
--- a/lib/gs-plugin-loader.h
+++ b/lib/gs-plugin-loader.h
@@ -109,4 +109,7 @@ void            gs_plugin_loader_set_max_parallel_ops  (GsPluginLoader *plugin_l
 const gchar	*gs_plugin_loader_get_locale		(GsPluginLoader *plugin_loader);
 GPtrArray	*gs_plugin_loader_dup_copy_dests	(GsPluginLoader *plugin_loader);
 
+gboolean	 gs_plugin_loader_app_copying 		(GsPluginLoader *plugin_loader,
+		                                         GsApp		*app);
+gboolean	 gs_plugin_loader_copy_queue_empty 	(GsPluginLoader *plugin_loader);
 G_END_DECLS
--- a/src/gs-details-page.c
+++ b/src/gs-details-page.c
@@ -42,7 +42,6 @@ struct _GsDetailsPage
 	GsPage			 parent_instance;
 
 	GsPluginLoader		*plugin_loader;
-	GsPluginStatus		 plugin_status;
 	GtkBuilder		*builder;
 	GCancellable		*cancellable;
 	GCancellable		*app_cancellable;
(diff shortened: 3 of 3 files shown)


lib/gs-plugin-loader.c
```diff
@@ -3274,6 +3276,9 @@ gs_plugin_loader_init (GsPluginLoader *plugin_loader)
 	for (i = 0; projects[i] != NULL; i++)
 		g_debug ("compatible-project: %s", projects[i]);
 	priv->compatible_projects = projects;
+
+	priv->global_plugin_statuses = g_hash_table_new (g_str_hash,
```
treitter: there should be a corresponding disposal of `priv->global_plugin_statuses` in `gs_plugin_loader_dispose()`. And please make sure that you free the keys (if necessary) as well as the structure

lib/gs-plugin-loader.c
```diff
@@ -2262,6 +2263,7 @@ gs_plugin_loader_status_changed_cb (GsPlugin *plugin,
 		}
 		return;
 	}
+	g_hash_table_insert(priv->global_plugin_statuses, gs_app_get_id(app), (guint)status);
```
treitter: note that you shouldn't use integers as values in a `GHashTable` since all values are treated as pointers.

You need to use the `GUINT_TO_POINTER()` and `GPOINTER_TO_UINT()` macros when inserting and retrieving integer values.

Here's a good explanation (from a former Endlesser!): https://stackoverflow.com/a/7124255

Reference docs here: https://developer.gnome.org/glib/stable/glib-Type-Conversion-Macros.html

src/gs-details-page.c
```diff
@@ -2726,9 +2729,8 @@ gs_details_page_setup (GsPage *page,
 	self->plugin_loader = g_object_ref (plugin_loader);
 	self->builder = g_object_ref (builder);
 	self->cancellable = g_object_ref (cancellable);
-
-	self->plugin_status = GS_PLUGIN_STATUS_FINISHED;
-
+	self->plugin_statuses = g_hash_table_new (g_str_hash,
```
treitter: Likewise, this needs to be disposed and its values packed and unpacked from pointers using the macros described above.

pwithnall:
Is there a task associated with this?

lib/gs-plugin-loader.c
edward-mwangi: Is this different from doing it in gs_plugin_loader_finalize? How do i determine which one's appropriate. ie between gs_plugin_loader_dispose and gs_plugin_loader_finalize.

lib/gs-plugin-loader.c
edward-mwangi: Thanks for the reference. Made the change.

lib/gs-plugin-loader.c
treitter: Please see the discussion here on dispose vs. finalize in GObject: https://mail.gnome.org/archives/gtk-devel-list/2004-July/msg00151.html

treitter: This might also be useful; in short, `dispose()` can be called multiple times whereas `finalize()` only happens when the memory will actually be freed: https://developer.gnome.org/gobject/stable/howto-gobject-destruction.html

treitter:
> Is there a task associated with this?

https://phabricator.endlessm.com/T26219

mwleeds:
nit: ideally the commiter name should be your whole name not "ed":
```
Author:     Edward Mwangi <edward.mwangi@endlessm.com>
AuthorDate: Mon Jan 20 12:18:51 2020 +0300
Commit:     ed <edward.mwangi@endlessm.com>
CommitDate: Wed Jan 29 20:10:01 2020 +0300
```

src/gs-details-page.c
mwleeds: No, I think we need to use `gs_plugin_loader_get_state()`. Why would we make a separate hash table here?

src/gs-details-page.c
```diff
@@ -2413,7 +2418,7 @@ gs_details_page_app_copied (GsPage *page, GsApp *app, const GError *error)
 {
 	GsDetailsPage *self = GS_DETAILS_PAGE (page);
 
-	if (error == NULL) {
+	if (error == NULL && g_strcmp0 (gs_app_get_id (self->app), gs_app_get_id (app)) == 0) {
```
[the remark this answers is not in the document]
mwleeds: Also would be better to do a pointer comparison as in `gs_details_page_plugin_status_changed_cb` (for performance and for correctness since `gs_app_get_id()` doesn't uniquely identify an app).

src/gs-details-page.c
```diff
@@ -2990,3 +2992,22 @@ gs_details_page_new (void)
...
 	return GS_DETAILS_PAGE (self);
 }
+
+gboolean
+gs_details_page_app_copying (GsDetailsPage *self)
+{
+	if (self->app == NULL) {
+		return FALSE;
+	}
+	return GPOINTER_TO_UINT(g_hash_table_lookup(self->plugin_statuses, gs_app_get_id (self->app))) == GS_PLUGIN_STATUS_COPYING;
```
mwleeds: Would be better style to do this in more than one line to improve readability

src/gs-details-page.c
```diff
@@ -2990,3 +2992,22 @@ gs_details_page_new (void)
...
+}
+
+gboolean
+gs_details_page_copy_queue_empty (GsDetailsPage *self)
+{
+	GList *statuses = g_hash_table_get_values (self->plugin_statuses);
+	if (g_list_find (statuses, (gpointer) GS_PLUGIN_STATUS_COPYING) != NULL) {
```
mwleeds: Don't you want `GUINT_TO_POINTER` here, not a cast?

src/gs-details-page.h
```diff
@@ -24,5 +24,6 @@ void		 gs_details_page_set_local_file(GsDetailsPage		*self,
 void		 gs_details_page_set_url	(GsDetailsPage		*self,
 						 const gchar		*url);
 GsApp		*gs_details_page_get_app	(GsDetailsPage		*self);
-
+gboolean 	 gs_details_page_copy_queue_empty (GsDetailsPage *self);
+gboolean     gs_details_page_app_copying (GsDetailsPage *self);
```
mwleeds: the indentation is off here, and should use tabs instead of spaces

src/gs-details-page.c
```diff
@@ -2689,15 +2694,13 @@ gs_details_page_plugin_status_changed_cb (GsPluginLoader *plugin_loader,
                                           GsPluginStatus status,
                                           GsDetailsPage *self)
 {
-	self->plugin_status = status;
-
 	if (app == NULL || app != self->app)
 		return;
 
+	g_hash_table_insert(self->plugin_statuses, gs_app_get_id(app), GUINT_TO_POINTER(status));
```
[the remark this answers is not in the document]
edward-mwangi: Loading state directly from `gs-plugin-loader.c`

mwleeds:
Looks like the latest version does not address https://github.com/endlessm/gnome-software/pull/524#discussion_r373596948

src/gs-details-page.c
treitter: This seems to be resolved in the latest version

treitter:
@edward-mwangi I resolved the comments you addressed but a few trivial fixes remain. I think, once those are done, this will be ready to merge

lib/gs-plugin-loader.c
```diff
@@ -4194,3 +4198,25 @@ gs_plugin_loader_get_locale (GsPluginLoader *plugin_loader)
...
+	}
+	return GPOINTER_TO_UINT (g_hash_table_lookup (priv->global_plugin_statuses,
+				gs_app_get_id (app))) == GS_PLUGIN_STATUS_COPYING;
+}
+
+gboolean
+gs_plugin_loader_copy_queue_empty (GsPluginLoader *plugin_loader)
+{
+	GsPluginLoaderPrivate *priv = gs_plugin_loader_get_instance_private (plugin_loader);
+	GList *statuses = g_hash_table_get_values (priv->global_plugin_statuses);
```
mwleeds: Should be `g_autoptr(GList)` since `g_hash_table_get_values` has its return value documented as `(transfer container)`

src/gs-details-page.c
```diff
@@ -267,9 +266,6 @@ gs_details_page_update_copy_button (GsDetailsPage *self)
 	gtk_widget_set_sensitive (self->button_copy, FALSE);
 	gtk_widget_set_visible (self->button_copy, FALSE);
 
-	if (self->plugin_status == GS_PLUGIN_STATUS_COPYING)
```
mwleeds: I don't think you want to remove this, since it would be wrong to update the copy button when the app is copying. Don't you want to change it to `gs_plugin_loader_app_copying()`?

src/gs-details-page.c
```diff
@@ -2689,15 +2692,12 @@ gs_details_page_plugin_status_changed_cb (GsPluginLoader *plugin_loader,
...
-	self->plugin_status = status;
-
-	if (app == NULL || app != self->app)
+	if (app == NULL)
 		return;
 
 	/* Various bits of UI state depend on the plugin status, so refresh
 	 * everything. */
 	gs_details_page_refresh_all (self);
-
```
mwleeds: Seems like a stray line deletion

lib/gs-plugin-loader.h
```diff
@@ -109,4 +109,6 @@ void            gs_plugin_loader_set_max_parallel_ops  (GsPluginLoader *plugin_l
 const gchar	*gs_plugin_loader_get_locale		(GsPluginLoader *plugin_loader);
 GPtrArray	*gs_plugin_loader_dup_copy_dests	(GsPluginLoader *plugin_loader);
 
+gboolean	gs_plugin_loader_app_copying (GsPluginLoader *plugin_loader, GsApp *app);
```
mwleeds: Alignment is still off here, the `gs_` needs to align with the ones above, as does the `(Gs`. And `GsApp`... should be on the next line.

mwleeds: (This is still true with the latest version, ef75c8f)

mwleeds: Looks good now

PR #525, opened 2020-02-15
Fix proxy apps updates support

andrunko:
Tested locally by downgrading both `EknServicesMultiplexer` and `quote_of_the_day`. The updates are showed in the `Updates` tab grouped in the `Endless Platform` proxy app and clicking `Update` on it actually updates the proxied apps as expected.

https://phabricator.endlessm.com/T28672
https://phabricator.endlessm.com/T29298

### DIFF
--- a/plugins/eos/gs-plugin-eos.c
+++ b/plugins/eos/gs-plugin-eos.c
@@ -914,7 +914,6 @@ static void
 process_proxy_updates (GsPlugin *plugin,
 		       GsAppList *list,
 		       GsApp *proxy_app,
-		       gboolean install_missing,
 		       const char **proxied_apps)
 {
 	g_autoptr(GsAppList) proxied_updates = gs_app_list_new ();
@@ -923,18 +922,6 @@ process_proxy_updates (GsPlugin *plugin,
 						      gs_app_get_unique_id (proxy_app));
 	if (cached_proxy_app != NULL)
 		gs_app_list_remove (list, cached_proxy_app);
-
-	/* add all the apps if we should install the missing ones; the real sorting for
-	 * whether they're already installed is done in the refine method */
-	if (install_missing) {
-		for (guint i = 0; proxied_apps[i] != NULL; ++i) {
-		        g_autoptr(GsApp) app = gs_app_new (proxied_apps[i]);
-			gs_app_add_quirk (app, GS_APP_QUIRK_IS_WILDCARD);
-			gs_app_set_bundle_kind (app, AS_BUNDLE_KIND_FLATPAK);
-
-			gs_app_list_add (proxied_updates, app);
-		}
-	}
 
 	for (guint i = 0; i < gs_app_list_length (list); ++i) {
 		GsApp *app = gs_app_list_index (list, i);
@@ -1002,7 +989,6 @@ add_updates (GsPlugin *plugin,
 
 	process_proxy_updates (plugin, list,
 			       framework_proxy_app,
-			       FALSE,
 			       framework_proxied_apps);
 
 	return TRUE;
--- a/plugins/eos/gs-plugin-eos.c
+++ b/plugins/eos/gs-plugin-eos.c
@@ -35,7 +35,6 @@
 
 #define ENDLESS_ID_PREFIX "com.endlessm."
 #define METADATA_SYS_DESKTOP_FILE "EndlessOS::system-desktop-file"
-#define METADATA_REPLACED_BY_DESKTOP_FILE "EndlessOS::replaced-by-desktop-file"
 #define EOS_PROXY_APP_PREFIX ENDLESS_ID_PREFIX "proxy"
 
 /*
--- a/plugins/eos/gs-plugin-eos.c
+++ b/plugins/eos/gs-plugin-eos.c
@@ -993,7 +993,6 @@ add_updates (GsPlugin *plugin,
 
 	const char *framework_proxied_apps[] = {"com.endlessm.Platform",
 						"com.endlessm.apps.Platform",
-						"com.endlessm.CompanionAppService.desktop",
 						"com.endlessm.EknServicesMultiplexer.desktop",
 						"com.endlessm.quote_of_the_day.en.desktop",
 						"com.endlessm.word_of_the_day.en.desktop",
--- a/plugins/eos/gs-plugin-eos.c
+++ b/plugins/eos/gs-plugin-eos.c
@@ -1009,13 +1009,6 @@ add_updates (GsPlugin *plugin,
 						_("Endless Platform"),
 						/* TRANSLATORS: this is the summary of the Endless Platform app */
 						_("Framework for applications"));
-	g_autoptr(GsApp) hack_proxy_app =
-		gs_plugin_eos_create_proxy_app (plugin,
-						EOS_PROXY_APP_PREFIX ".EOSHackUpdatesProxy",
-						/* TRANSLATORS: this is the name of the Endless Hack Platform app */
-						_("Endless Hack Platform"),
-						/* TRANSLATORS: this is the summary of the Endless Hack Platform app */
-						_("Endless Hack system components"));
 
 	const char *framework_proxied_apps[] = {"com.endlessm.Platform",
 						"com.endlessm.apps.Platform",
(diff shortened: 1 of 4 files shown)


andrunko:
Force pushed to fix some styling issues.

plugins/flatpak/gs-plugin-flatpak.c
```diff
@@ -274,6 +274,9 @@ gs_plugin_flatpak_get_handler (GsPlugin *plugin, GsApp *app)
 	GsPluginData *priv = gs_plugin_get_data (plugin);
 	const gchar *object_id;
 
+	if (gs_app_has_quirk (app, AS_APP_QUIRK_IS_PROXY))
+		goto select_by_scope;
```
wjt: :scream_cat:

andrunko: https://github.com/endlessm/gnome-software/commit/ec25ff8c5001a0d9923c4b1a7ab9f906504d2338#diff-1023bde5962b1305d7ead31178586dd8R281

andrunko: Alternatively we could set the management plugin for the proxy apps created on `gs_plugin_eos_create_proxy_app` to `flatpak`, but I am not sure if that will cause other issues.

andrunko: Or just keep this change but not use `goto`, but I think the change is so small/simple that I wouldn't bother.

plugins/eos/gs-plugin-eos.c
```diff
@@ -1024,32 +1017,13 @@ add_updates (GsPlugin *plugin,
...
-					   "com.endlessm.OperatingSystemApp.desktop",
-					   "com.endlessm.Sidetrack.desktop",
-					   NULL};
 
 	GsPluginData *priv = gs_plugin_get_data (plugin);
-	gboolean is_hack_product = g_strcmp0 (priv->product_name, "hack") == 0;
 
 	process_proxy_updates (plugin, list,
 			       framework_proxy_app,
 			       FALSE,
```
wjt: This parameter is now always `FALSE` so should be removed along with the code for the `TRUE` path.

andrunko: Yes, thought about it but decided to not remove this code as I thought this codepath could be useful (although currently unused) in case we keep using proxy apps.

I can still remove it if you prefer but I think this is independent of the fixes on this PR.

wjt: It was introduced as a last-minute hack to install new Hack 1 apps only on Hack 1 systems. We should remove it.

andrunko: Done.

plugins/flatpak/gs-plugin-flatpak.c
```diff
@@ -992,13 +989,11 @@ gs_plugin_flatpak_update (GsPlugin *plugin,
 	}
 	for (guint i = 0; i < gs_app_list_length (list_tmp); i++) {
 		GsApp *app = gs_app_list_index (list_tmp, i);
-		g_autofree gchar *ref = NULL;
 
-		ref = gs_flatpak_app_get_ref_display (app);
```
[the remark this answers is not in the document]
mwleeds: Andre said this change is needed because the proxy app is used in this code path, but the question remains of whether we should instead skip the call to `gs_flatpak_refine_app()`

plugins/flatpak/gs-flatpak-transaction.c
```diff
@@ -280,7 +335,34 @@ _transaction_progress_changed_cb (FlatpakTransactionProgress *progress,
...
+		guint64 total_expected_size_installed = 0;
+		guint64 total_size_installed = 0;
+
+		proxied_apps = gs_app_get_related (proxy_app);
+		for (guint i = 0; i < gs_app_list_length (proxied_apps); ++i) {
+			GsApp *proxied_app = gs_app_list_index (proxied_apps, i);
+			guint64 proxied_app_size_installed;
+			guint proxied_app_progress;
+
+			proxied_app_size_installed = gs_app_get_size_installed (proxied_app);
```
mwleeds: would be good to check for if proxied_app_size_installed is 0 or GS_APP_SIZE_UNKNOWABLE just in case

plugins/eos/gs-plugin-eos.c
```diff
@@ -1009,47 +977,19 @@ add_updates (GsPlugin *plugin,
...
-						"com.endlessm.EknServicesMultiplexer.desktop",
-						"com.endlessm.quote_of_the_day.en.desktop",
-						"com.endlessm.word_of_the_day.en.desktop",
+						"com.endlessm.EknServicesMultiplexer",
+						"com.endlessm.quote_of_the_day.en",
+						"com.endlessm.word_of_the_day.en",
 						NULL};
-	const char *hack_proxied_apps[] = {"com.endlessm.Clubhouse.desktop",
-					   "com.endlessm.Fizzics.desktop",
-					   "com.endlessm.GameStateService.desktop",
```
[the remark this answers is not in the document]
andrunko: I checked with Will and he asked to drop all hack proxy apps and that these apps were no longer supported (hack1 vs hack2 apps). New hack apps have prefix `com.hack_computer.` iiuc.

plugins/eos/gs-plugin-eos.c
```diff
@@ -1009,47 +977,19 @@ add_updates (GsPlugin *plugin,
...
-						_("Endless Hack system components"));
 
 	const char *framework_proxied_apps[] = {"com.endlessm.Platform",
 						"com.endlessm.apps.Platform",
-						"com.endlessm.CompanionAppService.desktop",
-						"com.endlessm.EknServicesMultiplexer.desktop",
-						"com.endlessm.quote_of_the_day.en.desktop",
-						"com.endlessm.word_of_the_day.en.desktop",
+						"com.endlessm.EknServicesMultiplexer",
+						"com.endlessm.quote_of_the_day.en",
```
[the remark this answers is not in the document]
andrunko: Yep, what is happening is that the apps received on `gs_plugin_add_updates` don't have the `.desktop` suffix and that was making the code on `process_proxy_updates` not find any update as the available apps ids didn't match the proxied apps.

plugins/flatpak/gs-plugin-flatpak.c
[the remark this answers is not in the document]
andrunko: I think we should still call gs_flatpak_refine_app on the proxy app given the eos plugin has some specific code to refine proxy apps (although I am not 100% sure it is reached by this call here).

plugins/flatpak/gs-flatpak-transaction.c
andrunko: You mean check and warn if 0? It wouldn't make any diff to the logic given this variable is only used on addition/multiplication. It could give a very slight performance improvement as we could skip some computations but then again we would gain an extra condition check in every call on this loop if the size is not 0.

I am already protecting against `total_expected_size_installed` when computing the progress for the proxy app.

plugins/flatpak/gs-plugin-flatpak.c
andrunko: Yep, agreed.

plugins/eos/gs-plugin-eos.c
```diff
@@ -283,7 +266,6 @@ gs_plugin_setup (GsPlugin *plugin,
 							      g_free, g_free);
 
 	image_version = get_image_version ();
```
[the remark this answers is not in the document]
andrunko: Done and also added another commit to remove an unused define.

plugins/flatpak/gs-flatpak-transaction.c
andrunko: Ok, I see now GS_APP_SIZE_UNKNOWABLE is not 0. I will update.

andrunko: Forced pushed with this update, didn't add any warning though.

Also checked and many other places are invoking `gs_app_get_size_installed (app)` without considering `GS_APP_SIZE_UNKNOWABLE` which in reality is only returned if `GS_IS_APP(app)` fails.

plugins/eos/gs-plugin-eos.c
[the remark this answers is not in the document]
andrunko: See also https://phabricator.endlessm.com/T28672#799768.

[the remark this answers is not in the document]
wjt: @erikos please confirm that these Hack 1 apps no longer need special treatment

plugins/flatpak/gs-plugin-flatpak.c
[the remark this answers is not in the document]
mwleeds: Ah looks like `gs_flatpak_refine_app_unlocked()` is a no-op if the app is a proxy, so never mind.

mwleeds:
LGTM except for the unresolved discussion [here](https://github.com/endlessm/gnome-software/pull/525#discussion_r381567987)

plugins/eos/gs-plugin-eos.c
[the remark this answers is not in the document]
wjt: I had forgotten my own comment. We have removed these apps from hack-apps. Those which are still relevant are added to eos-apps with a new name. Those which are not there are not currently relevant.

PR #529, opened 2020-03-09
gs-details-page: Avoid a NULL object dereference

pwithnall:
When checking for updates, the plugin status will change. If that
happens before an app detail page is first viewed by the user (which is
a common situation if gnome-software is started in the background and
does a background update check), a load of `gs_app_*()` methods will be
called on `self->app` when it is `NULL`, giving a load of critical
warnings.

Avoid that by doing what every other call site of
`gs_details_page_refresh_all()` does, and checking whether `self->app ==
NULL` first.


pwithnall:
Note: This is a downstream-only fix needed for USB app support. The bug doesn’t exist upstream as far as I can tell.

PR #536, opened 2020-04-08
T29286 Fix duplicate apps when refreshing category

pwithnall:
Revert two patches from https://phabricator.endlessm.com/T22807 whose caching behaviour was somehow causing duplicate apps to appear when refreshing a category. The same performance improvements should be gained by improving upstream performance (but this PR doesn’t do that, so it causes a performance regression, adding about 2s to each category load).

https://phabricator.endlessm.com/T29286

### DIFF
--- a/src/gs-category-page.c
+++ b/src/gs-category-page.c
@@ -138,21 +138,6 @@ sort_button_clicked (GtkButton *button, gpointer data)
 }
 
 static void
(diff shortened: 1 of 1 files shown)


mwleeds:
Sounds like you've already tested this and followed it up with performance improvements, thanks for that. LGTM

PR #539, opened 2020-04-30
T29392 Drop proxy apps support

pwithnall:
https://phabricator.endlessm.com/T29392


pwithnall:
There’s an unrelated fix for the USB app copying code in here, to fix a crash I saw when running `gnome-software --details-pkg` (see the commit message). Dropping proxy apps support itself is implemented as the big stack o’ reversions, which can all be dropped together with the commits they revert, when we next rebase.

pwithnall:
> Changes lgtm but I think we should probably revert commits [f5a7f7f](https://github.com/endlessm/gnome-software/commit/f5a7f7f9adfc270899b284cf67aa777745330bd8) and [3f87953](https://github.com/endlessm/gnome-software/commit/3f8795393cf15ae88c669ca98a260f91c2a0d334) as part of this PR as well.

Oops, of course. That is, in fact, the entire point of doing this task now. :facepalm: Done.

> I also would prefer to merge this once the issues with the proxied apps not getting displayed as updates (with this change applied) are sorted.

I don’t think it makes any difference, as long the metadata for the proxied apps is updated before they are next published.

andrunko:
> > I also would prefer to merge this once the issues with the proxied apps not getting displayed as updates (with this change applied) are sorted.
> 
> I don’t think it makes any difference, as long the metadata for the proxied apps is updated before they are next published.

I mean, I would prefer to not release with this change until the proxied apps get updated and released such that we guarantee users will still get the updates accordingly, but yeah not a hard requirement as apps can be released separately from the OS.

pwithnall:
The proxied apps’ metadata has been updated (see https://phabricator.endlessm.com/T29392), so I’m merging this.

PR #541, opened 2020-05-14
fixup! transaction: Sync main app's state if a related ref is installed

pwithnall:
Use GHashTableIter rather than g_hash_table_get_keys()

It’s a lot faster, as it doesn’t have to iterate through the table to
get the keys (and allocate a linked list for them), then the caller
iterate through that and query the hash table for *each* of them, and
then iterate through the list again to free it.

Not that this is a performance bottleneck, but the use of
`g_hash_table_get_{keys,values}()` irks me.

This introduces no functional changes.


pwithnall:
Minor cleanup, unrelated to any particular task. Just spotted it while reading the code.

andrunko:
Also landed this on eos3.8 fwiw.

@pwithnall Should we send this change upstream as well?

pwithnall:
Not specifically, it’s for a fix which isn’t itself upstream yet (and iirc there were some complications about upstreaming it which means that something else will land upstream eventually, or something like that).

PR #548, opened 2020-07-29
Bring back flatpak bundle fixes

mwleeds:
https://phabricator.endlessm.com/T30587


mwleeds:
> Changes lgtm but before merging just wanted to confirm whether the changes from this PR are cherry-picks from [upstream](https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/405/commits) or from our [prev downstream PR](https://github.com/endlessm/gnome-software/pull/523/commits) (iotw, whether this is in sync with the changes landed upstream).
> 
These are cherry-picks from our eos3.7 branch, because that was the route with fewer merge conflicts. With the exception of the txn cache commit (see below). Based on reading the upstream MR, I believe the differences between here and there are from the different branch histories.

> Related, this PR seems to have [one extra commit](https://github.com/endlessm/gnome-software/pull/548/commits/866bbf9e20a722634623e1d49186960484d9368c) compared to the upstream PR above, is this commit still needed and if so should we fwd it upstream as well?

Good catch. Looks like that commit was replaced by the commit "flatpak: Add app to be uninstalled to txn cache" per the discussion [here](https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/405#note_696635). And we already have that txn cache commit, so I dropped the one setting the state to UNKNOWN.

andrunko:
Ack, let's merge this on master and test it there, if all good we can go ahead with the backport.

PR #369, opened 2018-05-15
T21809

joaquimrocha:
https://phabricator.endlessm.com/T21809

PR #549, opened 2020-08-10
build: Enable unit tests for CI

mwleeds:
Make the CI execute unit tests rather than merely building
gnome-software for downstream PRs. In T29985 I landed fixes for the
flatpak extension unit tests and that brought this issue to light.

This commit can be squashed into "build: Override meson's default
options" during the next rebase.

https://phabricator.endlessm.com/T21809

### DIFF
--- a/lib/gs-metered.c
+++ b/lib/gs-metered.c
@@ -132,6 +132,13 @@ gs_metered_block_on_download_scheduler (GVariant      *parameters,
                                         GCancellable  *cancellable,
                                         GError       **error)
 {
+	/* FIXME: Use a mock service instead of skipping the check here.
+	 * https://phabricator.endlessm.com/T30646 */
+	if (g_getenv ("GS_UNIT_TESTS_SKIP_MOGWAI") != NULL) {
+		g_debug ("%s: Allowed to download (Skipping check in unit tests)", G_STRFUNC);
+		return TRUE;
+	}
+
 #ifdef HAVE_MOGWAI
 	g_autoptr(MwscScheduler) scheduler = NULL;
 	g_autoptr(MwscScheduleEntry) schedule_entry = NULL;
--- a/lib/gs-plugin-loader.c
+++ b/lib/gs-plugin-loader.c
@@ -2080,7 +2080,6 @@ remove_app_from_install_queue (GsPluginLoader *plugin_loader, GsApp *app)
 	g_mutex_unlock (&priv->pending_apps_mutex);
 
 	if (ret) {
-		gs_app_set_state (app, AS_APP_STATE_AVAILABLE);
 		id = g_idle_add (emit_pending_apps_idle, g_object_ref (plugin_loader));
 		g_source_set_name_by_id (id, "[gnome-software] emit_pending_apps_idle");
 		save_install_queue (plugin_loader);
@@ -3340,6 +3339,7 @@ gs_plugin_loader_app_installed_cb (GObject *source,
 						   res,
 						   &error);
 	if (!ret) {
+		gs_app_set_state (app, AS_APP_STATE_AVAILABLE);
 		remove_app_from_install_queue (plugin_loader, app);
 		g_warning ("failed to install %s: %s",
 			   gs_app_get_unique_id (app), error->message);
--- a/lib/gs-plugin-loader.c
+++ b/lib/gs-plugin-loader.c
@@ -2024,8 +2024,8 @@ pending_app_state_changed_cb (GsApp *app,
 			      GParamSpec *pspec,
 			      GsPluginLoader *plugin_loader)
 {
-	if (gs_app_get_state (app) != AS_APP_STATE_INSTALLING ||
-	    gs_app_get_state (app) != AS_APP_STATE_QUEUED_FOR_INSTALL)
+	if (gs_app_get_state (app) == AS_APP_STATE_INSTALLING ||
+	    gs_app_get_state (app) == AS_APP_STATE_QUEUED_FOR_INSTALL)
 		return;
 
 	remove_app_from_install_queue (plugin_loader, app);
--- a/lib/gs-app.c
+++ b/lib/gs-app.c
@@ -1017,6 +1017,7 @@ gs_app_set_state_internal (GsApp *app, AsAppState state)
 	case AS_APP_STATE_AVAILABLE_LOCAL:
 		/* local has to go into an action state */
 		if (state == AS_APP_STATE_UNKNOWN ||
+		    state == AS_APP_STATE_QUEUED_FOR_INSTALL ||
 		    state == AS_APP_STATE_INSTALLING)
 			state_change_ok = TRUE;
 		break;
--- a/lib/gs-self-test.c
+++ b/lib/gs-self-test.c
@@ -800,7 +800,11 @@ gs_app_list_related_func (void)
 int
 main (int argc, char **argv)
 {
-	g_test_init (&argc, &argv, NULL);
+	g_test_init (&argc, &argv,
+#if GLIB_CHECK_VERSION(2, 60, 0)
+		     G_TEST_OPTION_ISOLATE_DIRS,
+#endif
+		     NULL);
 	g_setenv ("G_MESSAGES_DEBUG", "all", TRUE);
 
 	/* only critical and error are fatal */
--- a/lib/gs-test.c
+++ b/lib/gs-test.c
@@ -35,3 +35,36 @@ gs_test_flush_main_context (void)
 	if (cnt > 0)
 		g_debug ("cleared %u events", cnt);
 }
+
+/**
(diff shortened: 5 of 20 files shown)


mwleeds:
This supercedes #369

mwleeds:
> Could “fedora-langpacks: Skip self tests under non-fedora distros” be submitted upstream? Would mean one fewer downstream patch.

Yeah, I was thinking the same thing. https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/495
> 
> Wrt “lib: Skip mogwai checks in unit tests”, see https://gitlab.gnome.org/GNOME/gnome-software/-/issues/825 for the current upstream status of this. I’m not going to get time to work on that in the near future. :(
> 

Ah, good to know
> Seems [the CI](https://ci.endlessm-sf.com/job/obs-gnome-software-pr/572/console) is not yet entirely happy though :(
> 
> ```
> --- stderr ---
> 
> (/build/src/_build/plugins/repos/gs-self-test-repos:25694): Gs-WARNING **: 01:17:17.208: failed to create an app for */*/*/*/chiron.desktop/*
> ```
> 
> Does this depend on the 1.8.x rebase having landed first — would that fix the tests?

Not that I'm aware of but I haven't checked

mwleeds:
bot, retest this please

mwleeds:
bot, retest this please

mwleeds:
Marking this as a draft despite the tests passing because I want to understand better if the commits I reverted are okay to revert.

mwleeds:
> Marking this as a draft despite the tests passing because I want to understand better if the commits I reverted are okay to revert.

I ended up adding patches instead. I think the downstream behavior with respect to the install queue is correct and should be upstreamed; see https://phabricator.endlessm.com/T30642

mwleeds:
I sent the patch "Skip saving install queue during unit tests" upstream.

lib/gs-plugin-loader.c
```diff
@@ -1915,6 +1915,11 @@ save_install_queue (GsPluginLoader *plugin_loader)
 	g_autoptr(GString) s = NULL;
 	g_autofree gchar *file = NULL;
 
+	/* Saving the install queue during unit tests could trip up other tests as
+	 * they are run in parallel */
+	if (g_getenv ("GS_UNIT_TESTS_SKIP_SAVING_INSTALL_QUEUE") != NULL)
+		return;
```
pwithnall, line 1921: I don’t think this is the right solution. Unit tests should be testing the code which is going to be run in production, not skipping various parts of it.

Would it work to allow the install queue path to be set in the test harness, either by passing it into libgnome-software as an argument, or by creating a per-test context directory using `G_TEST_OPTION_ISOLATE_DIRS`?

lib/gs-metered.c
```diff
@@ -132,6 +132,11 @@ gs_metered_block_on_download_scheduler (GVariant      *parameters,
                                         GCancellable  *cancellable,
                                         GError       **error)
 {
+	if (g_getenv ("GS_UNIT_TESTS_SKIP_MOGWAI") != NULL) {
+		g_debug ("%s: Allowed to download (Skipping check in unit tests)", G_STRFUNC);
+		return TRUE;
+	}
```
pwithnall, line 138: I’m not particularly enthusiastic about this or the parental controls environment variable. They mean we’re not actually testing the code we’re shipping, we’re testing some other code. Particularly with Mogwai and parental controls, this could result in the tests making it look like app installation works, but on actual systems it’s always disallowed through some interaction with Mogwai or parental controls.

The way to fix both of them is to run the gnome-software unit tests in a private D-Bus session bus, where the various daemons that it uses have been started by the test framework and configured to use that bus (setting `DBUS_SESSION_BUS_ADDRESS` in their environment). A custom D-Bus service directory might be needed, with a `.service` file for each service gnome-software uses, to allow service activation during the tests. Then `mogwai-scheduled` can be run as one of those services, and `accounts-daemon` as another, and that should get metered data and parental controls checks working.

A second phase of work on that would probably swap out `mogwai-scheduled` and `accounts-daemon` for mock implementations of their D-Bus interfaces, provided by the test harness. That would allow the results returned by them to be controlled by the tests. But I guess that for the first pass, just using the system-installed daemons and assuming that the configuration they load from the system is appropriate, will work.

If you don’t have time to do the first pass on that then I guess these environment variables will have to do. In that case I think they should definitely get a big `FIXME` comment each, pointing to tasks like https://phabricator.endlessm.com/T25340

pwithnall:
> I ended up adding patches instead. I think the downstream behavior with respect to the install queue is correct and should be upstreamed; see https://phabricator.endlessm.com/T30642

How much work do you think that would be, having looked at the code?

> I sent the patch "Skip saving install queue during unit tests" upstream.

For future reference: https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/496
plus also https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/495

lib/gs-metered.c
mwleeds: Completely agree that would be better but I don't think there's time for it right now. I'll add FIXME comments.

mwleeds:
> > I ended up adding patches instead. I think the downstream behavior with respect to the install queue is correct and should be upstreamed; see https://phabricator.endlessm.com/T30642
> 
> How much work do you think that would be, having looked at the code?
> 

I put an estimate of three days on the ticket. It's quite a few patches so I don't think it's something you could do in an afternoon unless everything goes incredibly smoothly.

lib/gs-plugin-loader.c
mwleeds: Yeah testing part of the code is sub-optimal but it's better than testing none at all which is the status quo. I'll look further into `G_TEST_OPTION_ISOLATE_DIRS`; I think you're right that would solve it and maybe it's not too much work to include it here. I happen to have also just filed [an issue about it](https://gitlab.gnome.org/GNOME/gnome-software/-/issues/1043).

mwleeds: Actually it looks like `G_TEST_OPTION_ISOLATE_DIRS` isn't a perfect solution for this: it doesn't take effect until `g_test_run()` is called so in `gs_plugin_loader_setup()` the install queue would be loaded from the system location, and then we would write the queue to the per-test isolated directory while the tests are running. So it would only work out fine in the normal case where the system install queue is empty.

mwleeds: So I guess instead we should add a way to override the install queue directory with a new `gs_plugin_loader_...` function (and separately someone should make use of `G_TEST_OPTION_ISOLATE_DIRS`).

mwleeds: Actually using an env var would be more in line with what we're doing with e.g. `GS_SELF_TEST_CACHEDIR`

mwleeds: Did that in the latest version of the branch

lib/gs-metered.c
mwleeds: Done

lib/gs-plugin-loader.c
pwithnall: > Actually using an env var would be more in line with what we're doing with e.g. `GS_SELF_TEST_CACHEDIR`

I don’t think `GS_SELF_TEST_CACHEDIR` is the right approach to copy. We need fewer env vars, not more (they’re not threadsafe, are global state where local state is more useful, and are hard to debug and log). The only reason that env var still exists is because I ran out of time to rearchitect the appstream plugin when I last poked the tests. (See the comment on, for example, line 1860 of `plugins/flatpak/gs-self-test.c`.)

See also commits 1536efc75bd4c554a2615a8e2eb56ce25e17ce60 and 9f1f5ba2c3e9fff25bac20512523f29a502a6c7c, which I just dug up.

I think it should be possible to copy the `load_install_queue()` call from `gs_plugin_loader_setup()` to `gs_plugin_loader_setup_again()`, which means it will be called at the start of each unit test, to re-load the install queue from that test’s tmp directory. It might need modifying to correctly free and drop the old install queue, but I guess that should be fairly straightforward.

mwleeds: I dropped the patch that set the environment variable to control the install queue location, and cherry-picked most of your patches from [460](https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/460), which is sufficient to stop the tests from interfering with each other's install queues.

mwleeds:
The last few commits I added are just cherry-picks from upstream, so I think it's reasonable to merge this without waiting for another round of review. They will be dropped anyway in the looming rebase. For the record, here they are:
```
cae2dcd42 tests: Fix racy assertion in flatpak test
d61a5ec58 core: Add an override for icon lookup paths for tests
21327de8b tests: Enable G_TEST_OPTION_ISOLATE_DIRS for all tests
```

PR #553, opened 2020-08-20
Fix ARM crash when switching categories

andrunko:
This reverts an upstream commit that was based on one of our downstream patches for 3.7 but that still causes the crash and re-applied the original patch from 3.7 which fixed the issue for me.

**NOTE: Do not merge this just yet as we are on freeze.**

https://phabricator.endlessm.com/T30658


andrunko:
Closing as the PR should be on top of eos3.8, will open a new one.

PR #554, opened 2020-08-20
Fix ARM crash when switching categories

andrunko:
This reverts an upstream commit that was based on one of our downstream patches for 3.7 but that still causes the crash and re-applied the original patch from 3.7 which fixed the issue for me.

NOTE: Do not merge this just yet as we are on freeze.

https://phabricator.endlessm.com/T30658


mwleeds:
Why is this PR not against the master branch?

andrunko:
> Why is this PR not against the master branch?

Because if the upstream PR lands we will get the fix during the rebase.

andrunko:
> > Why is this PR not against the master branch?
> 
> Because if the upstream PR lands we will get the fix during the rebase.

We are also dropping ARM 32 bits support on master...

wjt:
Approved by release team.

PR #555, opened 2020-09-04
T30171 Rebase on 3.38.0

pwithnall:
Rebase the gnome-software code on the upstream 3.37.92 release (specifically, directly onto the `3.37.92` tag).

Not too much to point out in this rebase.
 * I haven’t dropped the Speedwagon/Desktop shortcut/Discovery feed support commits yet, as it was too much to do in a single rebase pass. I have highlighted the relevant commits in the report below, though, so they can easily be dropped in the next rebase pass (likely onto 3.38.0), or can be `git revert`ed separately.
 * I squashed in a lot of small fixup patches to the commits they are fixing.
 * The only significant code changes I remember having to make were CSS parsing performance improvements in “Add the new GsBackgroundTile” and “Add GsSideFilterRow widget”, so those two commits will need to be re-reviewed in detail.
 * Everything else is noted as “(Rebase 3.38: blah blah)” in the commit messages.
 * We’ve gone from 238 to 133 patches.
 * Packaging changes in https://github.com/endlessm/gnome-software/pull/556.

---

Full commit-by-commit notes:
 * *pick* `62f342587 Revert "Remove the now-unused code that supported the colorful category buttons"` — keep, EOS-specific
 * *pick* `a5e7450c4 core: Add USB category` — USB support, pending upstreaming in https://phabricator.endlessm.com/T26487
 * *pick* `72f35f15e flatpak: Select USB category when loading a folder that has a removable remote` — USB support, pending upstreaming in https://phabricator.endlessm.com/T26487
 * *pick* `faa4b9414 flatpak: Allow offline installation of apps that have the USB category` — USB support, pending upstreaming in https://phabricator.endlessm.com/T26487
 * *pick* `ba63bb2de Set gnome-software as default handler for x-content/ostree-repository` — USB support, pending upstreaming in https://phabricator.endlessm.com/T26487
 * *pick* `6cc764376 flatpak: Emit updates-changed when a USB remote is loaded` — USB support, pending upstreaming in https://phabricator.endlessm.com/T26487
 * *pick* `18558a09d lib: Make gs_app_{get,set}_priority() public` — USB support, pending upstreaming in https://phabricator.endlessm.com/T26487
 * *pick* `50a82b988 flatpak: Add collection ID support to GsFlatpakApp` — USB support, pending upstreaming in https://phabricator.endlessm.com/T26487
 * *pick* `09696a6e9 flatpak: Add apps to the USB category` — USB support, pending upstreaming in https://phabricator.endlessm.com/T26487
[...]

PR #556, opened 2020-09-04
debian: GNOME 3.38 rebase

pwithnall:
Not a rebase on the upstream Debian packaging as it hasn’t been done
yet. This is a rebase on 3.37.92.

No significant changes from 3.36 packaging.

https://phabricator.endlessm.com/T30171

---

Code changes in https://github.com/endlessm/gnome-software/pull/555

### DIFF
--- a/debian/gnome-software-dev.install
+++ b/debian/gnome-software-dev.install
@@ -1,2 +1,3 @@
 usr/include/gnome-software/
+usr/lib/*/libgnomesoftware.a
 usr/lib/*/pkgconfig
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,10 @@
+gnome-software (3.38.0-1endless0) eos; urgency=medium
+
+  * Rebase on 3.38.0 (T30171)
+    - Install libgnomesoftware.a in gnome-software-dev
+
+ -- Philip Withnall <withnall@endlessm.com>  Tue, 15 Sep 2020 11:27:00 +0100
+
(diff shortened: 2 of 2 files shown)


mwleeds (commented):
Looks fine except the build failure I noted [here](https://github.com/endlessm/gnome-software/pull/555#issuecomment-689179091)

pwithnall:
Fixed the build failure

mwleeds (approved):

debian/gnome-software-dev.install
```diff
@@ -1,2 +1,3 @@
 usr/include/gnome-software/
+usr/lib/*/libgnomesoftware.a
```
mwleeds, line 2: I guess this is a product of [this commit](https://gitlab.gnome.org/GNOME/gnome-software/-/commit/d3fae3b356edc59b53e476de19549c118a6bdc78), and we don't need or want to support out-of-tree plugins, so that's why we're putting it in gnome-software-dev instead of gnome-software?

mwleeds: Might be worth explaining that in the commit message

pwithnall: We’re putting it in gnome-software-dev rather than gnome-software because aiui `.a` static libraries always go in the `-dev` package. They’re only needed when building things which depend on them, not when running those things.

I’ve mentioned it in the commit message, thanks

pwithnall:
(merged manually)

PR #557, opened 2020-09-08
flatpak: Fix offline installation

mwleeds:
This commit fixes offline installation of .flatpak bundles, and fixes
offline installation of .flatpakref files that point to a file:// URL.
Currently such installations would be forever stuck in the "Pending
installation..." state. The check for a file:// prefix on the
gs_app_get_origin_hostname() return value is wrong because in case a
file:// URL is set, "localhost" is returned. See gs_app_func() in
lib/gs-self-test.c.

One might wonder if this could be implemented by checking if
gs_app_get_local_file() returns a non-NULL value, but
gs_app_set_local_file() is called for flatpakref files which are not
always offline friendly.

https://phabricator.endlessm.com/T30749

### DIFF
--- a/plugins/flatpak/gs-plugin-flatpak.c
+++ b/plugins/flatpak/gs-plugin-flatpak.c
@@ -781,8 +781,18 @@ static gboolean
 app_has_local_source (GsApp *app)
 {
 	const gchar *url = gs_app_get_origin_hostname (app);
-	return gs_app_has_category (app, "usb") ||
-		(url != NULL && g_str_has_prefix (url, "file://"));
+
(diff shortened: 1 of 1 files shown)


mwleeds:
Even though this PR is against the master branch, I tested the patch on top of the rebase branch in #555, and it should be picked onto that branch.

mwleeds:
@pwithnall if this commit looks good you could pull it into #555

mwleeds:
(Fixed whitespace)

mwleeds:
This was merged upstream.

PR #560, opened 2020-09-17
T30446 Drop desktop shortcuts support

pwithnall:
Follow-up from https://phabricator.endlessm.com/T30171.

https://phabricator.endlessm.com/T30446


andrunko (approved):
LGTM - lets follow-up on the "ReplaceApplication" support separately - we will probably need some update on the shell for that.

PR #571, opened 2021-03-10
eos: Remove eos-link-* special handling from blocklist

andrunko:
We are removing all web links from the OS, let's remove the special handling here also.

Depends on whether we are also dropping image personalities' specific web links - see https://github.com/endlessm/eos-image-builder/pull/975.

https://phabricator.endlessm.com/T31560


pwithnall (approved):
LGTM iff https://github.com/endlessm/eos-image-builder/pull/975 is merged.

wjt:
As mentioned in https://phabricator.endlessm.com/T31560#878405 I think we can remove this special handling anyway (unless I am missing something).

PR #577, opened 2021-12-03
[DROP-ME] tests: Temporarily disable failing tests

jprvita:
This is currently breaking ostree builds. Let's temporarily disable
failing tests so we can continue working on other high priority tasks
while they get fixed properly.

https://phabricator.endlessm.com/T32793


wjt:
Are you sure only these tests are failing out of the whole suite?

wjt:
Let's find out!

jprvita:
> Are you sure only these tests are failing out of the whole suite?

Yep, I've tested with jenkins-obs and it build on my OBS branch.

PR #578, opened 2021-12-03
Better flatpak skip parental controls

dbnicholson:
This uses upstream flatpak's environment variable to skip parental controls rather than our downstream environment variable that was recently removed. I tested this in OBS and it succeeded. I believe the 2nd commit is a candidate for upstream gnome-software (at least until malcontent mocking is written or there's some other non-dbus method to skip parental controls).

https://phabricator.endlessm.com/T32793

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -237,6 +237,9 @@ test_env = [
   'GSETTINGS_BACKEND=memory',
   'MALLOC_CHECK_=2',
   'GS_UNIT_TESTS_SKIP_MOGWAI=1',
+  # Make flatpak skip parental controls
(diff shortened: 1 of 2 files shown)


jprvita (approved):

pwithnall:
> I believe the 2nd commit is a candidate for upstream gnome-software (at least until malcontent mocking is written or there's some other non-dbus method to skip parental controls).

Yes, please submit upstream when you get a minute

pwithnall:
Ah, Dan already had and I hadn’t got to it in my inbox yet.

=> https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/1126

PR #582, opened 2022-01-24
debian: Stop disabling Snap in EOS build

pwithnall:
While we don’t want the snap plugin to be installed or enabled on EOS,
the Debian packaging has now been changed (upstream in Debian) to build
the snap plugin as a separate package, which we can just not install on
EOS.

Explicitly disabling the plugin causes that package to fail to build.

This can be squashed into the commit to apply Endless-specific packaging
changes next time we rebase.

https://phabricator.endlessm.com/T32854

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -55,7 +55,6 @@ GS_CONFIGURE_FLAGS += \
 	-Dmogwai=true \
 	-Dostree=true \
 	-Dpackagekit=false \
-	-Dsnap=false \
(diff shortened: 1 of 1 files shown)


pwithnall:
This should hopefully fix the [failed OBS build](https://obs-master.endlessm-sf.com/package/live_build_log/eos:master:endless/gnome-software/endless/x86_64)

PR #584, opened 2022-01-24
Move 3 dpkg/apt-related packages to Suggests:

wjt:
- libgtk3-perl: Needed for debconf
- software-properties-gtk: Debian-specific apt repository configuration tool
- apt-config-icons: Needed to show icons for apps to be installed with apt
- packagekit: It's packagekit!

None of these are relevant on Endless OS where apps cannot be installed
with apt.

This is spiritually a cherry-pick of
683c20c33345a8fd9fe8620e42b35fec0e75987c.

https://phabricator.endlessm.com/T32854
￼

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -52,18 +52,18 @@ X-Ubuntu-Use-Langpack: yes
 Package: gnome-software
 Architecture: any
 Depends: appstream,
-         apt-config-icons,
          gnome-software-common (= ${source:Version}),
          gsettings-desktop-schemas (>= 3.18),
-         libgtk3-perl,
-         packagekit (>= 1.1.11),
-         software-properties-gtk,
          ${misc:Depends},
          ${shlibs:Depends}
 Conflicts: sessioninstaller
 Recommends: fwupd [linux-any], ${plugin:Recommends}
-Suggests: apt-config-icons-hidpi,
+Suggests: apt-config-icons,
+          apt-config-icons-hidpi,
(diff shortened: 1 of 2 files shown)


debian/control
```diff
@@ -6,7 +6,7 @@ Source: gnome-software
 Section: gnome
 Priority: optional
 Maintainer: Debian GNOME Maintainers <pkg-gnome-maintainers@lists.alioth.debian.org>
-Uploaders: Laurent Bigonville <bigon@debian.org>, Matthias Klumpp <mak@debian.org>
```
pwithnall, line 9: Is the change to `Uploaders:` intentional?

pwithnall (approved):
Sigh, I was hoping to reduce the packaging delta, but I guess that’s still not possible with these :(

debian/control
wjt: I generate `debian/control` from `debian/control.in` by running `debian/rules clean` which I believe is the intended process. I didn't intentionally change this, it's just what the computer did. Happy to remove this hunk.

wjt:
I think the delta could potentially be avoided by moving some of the offending bits to separate packages, just like `-flatpak` and `-snap` already are, and/or putting some:


```

ifeq (yes,$(shell dpkg-vendor --derives-from Endless && echo yes))
```

into `rules`. But perhaps one for another day.

debian/control
wjt: I removed this by hand. It is generated by `/usr/share/gnome-pkg-tools/1/rules/uploaders.mk`, which derives the list using a gnarly heuristic of `debian/changelog` entries.

GeorgesStavracas (approved):

PR #597, opened 2022-09-22
T33843 (Debian)

GeorgesStavracas:
A pretty direct update. Only minor rebase conflicts.

https://phabricator.endlessm.com/T33843


GeorgesStavracas:
Retest this please

GeorgesStavracas:
retest this please

debian/control
```diff
@@ -16,15 +16,15 @@ Build-Depends: debhelper-compat (= 13),
                gtk-doc-tools,
                libadwaita-1-dev (>= 1.0.1),
                libappstream-dev (>= 0.14.0),
-               libflatpak-dev (>= 1.14.0) [amd64 arm64 armel armhf i386 mipsel mips64el ppc64el s390x hppa powerpc ppc64 riscv64 x32],
+               libflatpak-dev [amd64 arm64 armel armhf i386 mipsel mips64el ppc64el s390x hppa powerpc ppc64 riscv64 x32],
```
wjt, line 19: I was a little sceptical of this patch so I went looking for why Debian raised the dependency.

Here's the commit: https://salsa.debian.org/gnome-team/gnome-software/-/commit/562eb7ea73e362e561e154deefe5d5bf954519df

> ### Update dependency versions
> in preparation for the libsoup3 transition

Flatpak 1.14 introduced a curl backend…

debian/control
```diff
@@ -16,15 +16,15 @@ Build-Depends: debhelper-compat (= 13),
...
                libfwupd-dev (>= 1.5.6) [linux-any],
                libglib2.0-dev (>= 2.70.0),
                libglib-testing-0-dev,
                libgnome-desktop-3-dev (>= 3.18.0),
                libgtk-4-dev (>= 4.6),
                libgudev-1.0-dev [linux-any],
                libjson-glib-dev (>= 1.6.0),
-               libostree-dev (>= 2022.5-2~) [linux-any],
+               libostree-dev [linux-any],
```
wjt, line 27: and Debian's 2022.5-2 ostree package switched to the libcurl backend. OK!

wjt (approved):

debian/rules
```diff
@@ -43,6 +43,19 @@ ifeq ($(DEB_HOST_ARCH_OS), linux)
 	endif
 endif
 
+# EOS-specific configuration
+GS_CONFIGURE_FLAGS += \
+	-Deos_updater=true \
+	-Dexternal_appstream=true \
+	-Dfwupd=false \
+	-Dgudev=false \
+	-Dhardcoded_popular=false \
```
wjt, line 52: This will fail to build.

See 0983f7392c204ed1ca6dba69061761401d1796aa.

debian/rules
```diff
@@ -43,6 +43,19 @@ ifeq ($(DEB_HOST_ARCH_OS), linux)
...
+	-Dmalcontent=true \
+	-Dmogwai=true \
+	-Dpackagekit=false \
+	-Dsoup2=true \
+	-Dtests=true
```
wjt, line 57: and this will override d922fd66794c4b318c2965db62f0942ea77f2a9b on OBS and presumably again fail the build.

wjt (changes requested):

debian/control
GeorgesStavracas: Correct. I've added this to the commit message. We will be able to drop this patch later.

debian/rules
GeorgesStavracas: In theory tests are disabled a few lines below:

```
# DISABLED
override_dh_auto_test:
```

But removed nonetheless

wjt (approved):

wjt:
Pushed a patch to fix:

```
dpkg-gencontrol: warning:     debian/changelog(l5): ignoring invalid week day 'Qui'
```

wjt:
The resulting package builds but fails to install:

```
The following packages have unmet dependencies:
 gnome-software-plugin-flatpak : Depends: flatpak (>= 1.14.0) but 1.12.7+dev7.438e811-0bem1 is to be installed
E: Unable to correct problems, you have held broken packages.
```

PR #598, opened 2022-09-28
Drop Snap

GeorgesStavracas:
We don't plan on enabling Snap on Endless OS.

https://phabricator.endlessm.com/T33843


jprvita (approved):
This should solve the OBS building breakage we are seeing, but perhaps we should report to Debian that libsnapd-glib-dev is missing from the build-dependencies list so we can drop this commit in a future rebase.

PR #599, opened 2022-09-29
plugin-loader: Refine before installing pending apps

GeorgesStavracas:
The installation of pending apps is handled by GsInstalledPage usually, which refines the app list before actually installing them. However, there is an alternative code path in GsPluginLoader that doesn't refine, and yet tries to install the pending apps queue.

Not refining the pending apps list is risky, since the Flatpak plugin needs to add metadata to the app in order to proceed with installation.

Refine the pending apps queue before installing it.

https://phabricator.endlessm.com/T33188


GeorgesStavracas:
The patch applied cleanly, no conflicts.

PR #605, opened 2022-12-19
T34222 Rebase on 43.2

pwithnall:
Fairly straightforward rebase on upstream 43.2 (actually the `gnome-43` branch, which is a couple of commits ahead of 43.2).

Packaging changes are in #606.

These two commits were automatically dropped in the rebase:
 * 48f179b66cf3d90a6d9872691da74bdffb430220 `plugin-loader: Refine before installing pending apps`
 * 251f3c1916d6eed25b978f765aeee6dad41915b7 `gs-plugin-job-list-apps: Filter out invalid apps`

And the other commits in the rebase list were:
 * pick `0aaa98751 src: Expose origin hostname in app details for blocklisting`
 * pick `c0b67f8d0 eos: Add eos-blocklist plugin`
 * pick `a4987ecca gs-prefs-dialog: Hide automatic updates setting when using Mogwai`
 * pick `7ded7d2ea Add the top bar about auto updates to the Updates page`
 * pick `a1fc07d61 Add EOS desktop plugin`
 * pick `318a92a88 eos: Mark all non-flatpak apps as compulsory if installed`
 * pick `219968988 eos: Fix launching Google Chrome`
 * pick `fc1dff0ea eos: Add parental controls metrics on update poll`
 * pick `678397d6c Change user-facing name from Software to App Center`
 * pick `6e9cf9cbd Limit the number of certain ops in parallel to 1`
 * pick `a488e14aa build: Override meson's default options`
 * pick `8e65e4322 flatpak: Set the apps' update priority`
 * pick `1e67331e2 Do not invalidate OS upgrades in the Updates page`
 * pick `fc935725f flatpak: Introduce success/failure heuristics for app install and autoupdates`
 * pick `9a63310ed updates monitor: Don't skip automatic updates that require new permissions`
 * pick `35d23f7f4 appstream: Override <id> with <launchable type="desktop-id"> while parsing appdata files`
 * pick `2d0dea9fe Integrate Transifex translation infrastructure`
 * pick `c7e01b380 Disable all tests`
 * **modified** `ef29475c4 appdata: Remove release notes` — reworked to drop latest release notes
 * pick `b742b8145 eos: Don't hide web-apps or mark them as unremovable`
 * pick `a6f08d4f1 Migrate transifex configuration for use with new tx client`
 * **modified** `4150fdf98 Add translation support for several new languages` — fixed trivial rebase conflicts
 * **modified** `2861f7ff7 Update .po files` — regenerated using `ninja`
[...]

PR #606, opened 2022-12-19
T34222 Rebase on 43.2 (packaging changes)

pwithnall:
Packaging changes to go with #605. Basically nothing has changed.

https://phabricator.endlessm.com/T34222


GeorgesStavracas (approved):
Seems like I left we all hanging with my commit message on commit 15ba84e

`Debian raised the dependency so that they could do the switch to libsoup3, and`

Other than that, LGTM

PR #607, opened 2022-12-22
T34082 Backport ‘Learn More’ link changes for upgrades from upstream

pwithnall:
Trivial backport of https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/1567 to EOS `master`.

This will need fast-forwarding to on `eos5.0` once landed, and then backporting to `eos4.0`. I haven’t done that yet as there are a couple of other backports waiting review and it’s getting to the point where a merge train would be needed if I added more PRs.


wjt:
Fast-forwarded eos5.0 as well

PR #608, opened 2023-01-05
eos-updater: Use the OS logo for minor update rows (backport to eos4.0)

pwithnall:
Just like using an app icon to brand the updates for an app, this makes it a bit more obvious what the updates are.

(Backport 4.0: Rework code in eos-updater plugin as the control flow is significantly different in eos4.0 vs eos5.0.)

Upstream: https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/1556 https://phabricator.endlessm.com/T33143


pwithnall:
This is a not-so-trivial backport of #605 to `eos4.0`.

PR #611, opened 2023-01-12
gs-updates-page: Ensure correct section is used for minor OS updates

pwithnall:
Image based OS updates, which need a restart to be applied, should always be labelled as needing a restart.

While that would happen OK if gnome-software added them to the updates page when eos-updater is polling or ready, if they were added to the page when already being fetched (or in subsequent states of eos-updater), the `GsApp` for the update would be in the `GS_APP_STATE_INSTALLING` state, and this would trigger the updates page to add the update to the wrong section.

Fix that by always putting OS updates which need a reboot into the `OFFLINE` section.

Upstream: https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/1588

Fixes: #2028, #2038

### DIFF
--- a/src/gs-updates-page.c
+++ b/src/gs-updates-page.c
@@ -157,6 +157,10 @@ gs_updates_page_invalidate (GsUpdatesPage *self)
 static GsUpdatesSectionKind
 _get_app_section (GsApp *app)
 {
+	if (gs_app_get_kind (app) == AS_COMPONENT_KIND_OPERATING_SYSTEM &&
+	    gs_app_has_quirk (app, GS_APP_QUIRK_NEEDS_REBOOT))
+		return GS_UPDATES_SECTION_KIND_OFFLINE;
+
 	if (gs_app_get_state (app) == GS_APP_STATE_UPDATABLE_LIVE ||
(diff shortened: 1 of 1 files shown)


pwithnall:
Trivial backport of https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/1588 to EOS gnome-software. There is no corresponding Phab issue, since this is an opportunistic quick backport before we actually encounter the bug in EOS.

It’ll need landing on EOS 5 once it’s reviewed for master. Probably less important to backport to EOS 4 since the only update people will be seeing there is the major update to EOS 5.

wjt:
> Probably less important to backport to EOS 4 since the only update people will be seeing there is the major update to EOS 5.

Well… they will see an update from 4.0.13 (which predates our recent work in this area) to 4.0.14 (which includes the changes to relegate minor updates). And if we have to push 4.0.15 to fix some migration bugs we find after the 4.0.14 release (this happened at the eos3 -> eos3a checkpoint, many times; I don't know about eos3a -> latest1) then they will see a subsequent minor update on eos4.

wjt:
LGTM, I'll leave the backporting in your hands!

PR #613, opened 2023-03-03
eos-updater: Ignore WrongState errors from eos-updater daemon

pwithnall:
These imply that the state tracking in gnome-software has got out of sync with the state tracking in the eos-updater daemon. This happens as a result of two client processes racing to apply updates, typically gnome-software racing with eos-autoupdater.

This is actually harmless, as both gnome-software and eos-autoupdater will re-sync their state with the eos-updater daemon at the start of each state machine cycle.

The user doesn’t need to know about this, and can’t do anything about it, so just ignore the error.

Fixes: #2078
Upstream: https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/1639

(Backport 5.1: Significantly reworked as the upstream code is asynchronous, but this code is still synchronous.)

https://phabricator.endlessm.com/T31618

### DIFF
--- a/plugins/eos-updater/gs-plugin-eos-updater.c
+++ b/plugins/eos-updater/gs-plugin-eos-updater.c
@@ -1050,6 +1050,19 @@ cancelled_cb (GCancellable *ui_cancellable,
 	g_cond_broadcast (&self->state_change_cond);
 }
 
+static gboolean
+is_wrong_state_error (const GError *error)
+{
+	g_autofree gchar *remote_error = NULL;
+
+	if (!g_dbus_error_is_remote_error (error))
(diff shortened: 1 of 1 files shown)


pwithnall:
Compile tested only due to running out of time this week. I think the changes are straightforward enough to merge and then test on an image next week, but I can instead do some more in-depth pre-merge testing next week if desired.

pwithnall:
@dbnicholson, fancy reviewing this?

dbnicholson (approved):
Makes sense and seems like it should work correctly. Indeed, the upstream async code using function pointers is much nicer.

PR #615, opened 2023-03-20
T33138 flatpak: Don’t signal success for automatic updates if not enough space

pwithnall:
Previously the code tried to not return an error to the user if
automatic updates failed in the background due to a lack of disk space.

However, by returning `TRUE` from `gs_plugin_flatpak_update()`, this
signalled success for the updates, which would result in a “123
Applications Updated” notification banner which was completely
incorrect.

Instead, it’s better to return an error, and then let the UI layer
determine whether to display that as a massive warning dialogue, a small
in-app warning notification, or not at all.

https://phabricator.endlessm.com/T33138

### DIFF
--- a/plugins/flatpak/gs-plugin-flatpak.c
+++ b/plugins/flatpak/gs-plugin-flatpak.c
@@ -915,13 +915,11 @@ remove_schedule_entry (gpointer schedule_entry_handle)
 }
 
 static gboolean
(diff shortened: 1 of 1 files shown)


pwithnall:
I’ve tested this briefly locally, but it could do with more testing. @wjt, can you still reproduce the original failure case easily? If so, that’ll do for testing. If not, I’ll set up some outdated apps in an installation and run through some scenarios more thoroughly.

None of these changes are upstreamable, as all of this code is currently downstream-only.

PR #617, opened 2023-04-04
debian: Disable hardcoded webapps

dbnicholson:
We're going to consume the current webapp lists into our own external appstream and don't want duplicates from the hard coded lists.

https://phabricator.endlessm.com/T34569
(cherry picked from commit 0e8949841daab47f395548885757baafbe63d50a)
https://phabricator.endlessm.com/T34678


wjt:
I fast-forwarded    23c1040cb..3154585f9  instead because this was the only change.

dbnicholson:
Oops, I did it again

Clearly out of practice after vacation.

PR #619, opened 2023-10-20
Upstream and drop some old downstream patches

pwithnall:
Here are a few commits based on me going through the list of pending downstream gnome-software patches today, and looking at what I could upstream in half a day.

The results are presented as a series of reverts of the affected commits, with either no replacement, or with a linked upstream merge request.

See the commit messages for details.

Ideally this is ready to incorporate into our next rebase of gnome-software, but due to time pressures I haven’t actually tested it, so whoever lands it should do that and be prepared to abandon parts of it if I’ve forgotten something.


dylanmccall:
The proposed changes here have all been triaged appropriately now. Kudos for sorting through this ahead of time, @pwithnall! :)

PR #621, opened 2024-03-06
Rebase on GNOME Software 45

dylanmccall:

-----

Please merge this along with:

- https://github.com/endlessm/gnome-software/pull/622

using a force push so commit hashes are unchanged.

-----

```
# EOS infrastructure stuff

pick 59af59f5fd4860f1fb4880c8c42b17daceda09d9 Integrate Transifex translation infrastructure
pick 1ee82adc0f8266ee96a98083d0ccdc80fc82528f Migrate transifex configuration for use with new tx client
pick 25181a0c4886b40048c62b33478e0b9e728fb0e0 Disable all tests
pick f551791fc867db27dda2df0b0fa7724da34f3729 build: Override meson's default options


# Desktop integration

pick db878736163215f6bfa43ada01cc94b0eecc5790 appdata: Remove release notes
pick 9af695c88d7d9c63ee7b1690716333465b724cd8 Change user-facing name from Software to App Center
pick 8cfb4c69f2830cb62d0b6f86d7bf0f3f1e667b04 src: Expose origin hostname in app details for blocklisting
pick 0bde12368269eb0f08e9231619d99ae667685d3a eos: Add eos-blocklist plugin
pick e710fdbcfa1b672ab355f6a91f724941e92e0d82 Add EOS desktop plugin
pick d001d2881a81938ff9b0804a42d75769210d98bd gs-prefs-dialog: Hide automatic updates setting when using Mogwai
pick 937c27f4f606616eebb45723f56eec750ce44d1a Add the top bar about auto updates to the Updates page
pick 1fe3d2c629ecadd66a298071c14e2b2fcfd00734 eos: Mark all non-flatpak apps as compulsory if installed
pick 029358f71dd0070603937bb63a674058af8ffa59 eos: Fix launching Google Chrome
pick 54828e8d0501bf1ebaac1f5e7a4535efe901aca7 eos: Add parental controls metrics on update poll
pick 723be30e36014af8f4eb75e6548269cec556070b Limit the number of certain ops in parallel to 1
pick d8eecfd5af782caa03479a37c3d729ca89bc4710 Do not invalidate OS upgrades in the Updates page
pick 2d9fa3015f3b8e1c89be011971a2b9c332de1d6c flatpak: Introduce success/failure heuristics for app install and autoupdates
squash 325a46872c31e6fa8671d7a2771fe649987e87a3 flatpak: Don’t signal success for automatic updates if not enough space
pick 52167ff2f2c2a097a284a6e748bbf392743e0174 updates monitor: Don't skip automatic updates that require new permissions
pick 6f8441aa1946fff6c4c6ea359bbcfd0b4afb84e8 eos: Don't hide web-apps or mark them as unremovable
pick 7c9a7418937db48d4c3e70b2ff104e8fc9adb8df flatpak: Simplify internal function to get installation’s free space
pick 0a6c9282575591efe849f8fb532cf9261029fced flatpak: Ignore space requirements for apps which are already downloaded

drop 8ab012b0db0a052d3f29e39f63830ac8ac73222d appstream: Override <id> with <launchable type="desktop-id"> while parsing appdata files
#- We believe this is no longer necessary.

drop 35fa3a70f08c11603639c4ed576fe4b1a72e14df flatpak: Set the apps' update priority
#- This is no longer useful.

drop 36733834765c58d56ca9bf123a284d722289c73b eos-updater: Update D-Bus interface description from upstream
#- Applied upstream in <https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/1555>
drop b22a8962c59ec154aeefef595457f55abf1fdc83 eos-updater: Update python-dbusmock template for eos-updater
#- Applied upstream in <https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/1555>
drop 859522207d16a218147e28ea6e6e78337de2ec52 eos-updater: Clear download size once update is fetched
#- Applied upstream in <https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/1555>
drop 457442cb5dee4ce69e9263cf7339fceda46259d4 eos-updater: Change icon used for eos-updater updates
#- Applied upstream in <https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/1555>
drop e76c8f03adf4f677031cd92a8457f085ded6fded eos-updater: Rename gs_plugin_app_upgrade_download() implementation
#- Applied upstream in <https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/1555>
drop 0f3d5b33876210019add2137c9f11095691215d3 eos-updater: Support major vs minor OSTree updates in the UI
#- Applied upstream in <https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/1555>

drop b1d16cbbbf310e73cc3bc2f97052b4b539a717de gs-updates-page: Ensure correct section is used for minor OS updates
#- Applied upstream in <https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/1588>
drop 930335bc909071227f4ecc14de6ad25ca331d963 eos-updater: Ignore WrongState errors from eos-updater daemon
#- Applied upstream in <https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/1639>
drop c2e615f18e522c3b94bfc41b4279c64305934bda eos-updater: Use the OS logo for minor update rows
#- Applied upstream in <https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/1556>


# Translations

drop baee429961bb72527510a993c0d1167d8a50b64d Add translation support for several new languages
#- Based on the PO files in the eos5.1 branch, we do not have new strings for any of these languages.

drop 94e133739820c6c887b1362ff6f6d5700dc63a31 Update .po files
drop 378ed125200e46227bab38ddac1341527dd4a253 New translations from Transifex
drop 7673f07393d5c06ee082c37fe7ba828dafe13640 New translations from Transifex
drop 3d4643c9dc1bc2c370609508c165388e41beb6ab New translations from Transifex
drop ca3eebfaecde8ae0e6564954cc0b9b7582f72ac9 New translations from Transifex
drop 8ae9036d67b539df8b3b7c6697dbb728ab651096 New translations from Transifex
```

### DIFF
--- a/plugins/flatpak/gs-plugin-flatpak.c
+++ b/plugins/flatpak/gs-plugin-flatpak.c
@@ -1124,7 +1124,8 @@ gs_flatpak_has_space_to_update (GsFlatpak *flatpak, GsAppList *list, gboolean in
 
 			size_type = gs_app_get_size_installed (app_temp, &installed_size);
 
-			if (size_type == GS_SIZE_TYPE_VALID)
+			if (size_type == GS_SIZE_TYPE_VALID ||
+			    !gs_app_get_is_update_downloaded (app_temp))
 				space_required += installed_size;
 		}
 	}


wjt:
> drop db878736163215f6bfa43ada01cc94b0eecc5790 appdata: Remove release notes
> #- Is this relevant?

Yes it is. It [remains the case](https://gitlab.gnome.org/Infrastructure/damned-lies/-/issues/473) that GNOME's translation infrastructure has a patched version of the [metainfo ITS rules](https://gitlab.gnome.org/Infrastructure/damned-lies/-/blob/4a653d8408d451ccc5c327fc866c795c513b3884/common/static/metainfo.its) so as not to extract release notes for translation. We don't have that patch in our script to extract strings for translation. As a result, we would find ourselves pushing many years of Software release notes up to Transifex without this patch. And then if we want to buy translations for other strings in the application, Transifex doesn't have a way to exclude certain strings from the order, so we would buy translations for many years of release notes that are just not interesting.

However we could imagine addressing this in a less destructive way:

1. Repeat 7dc447128c7ab5d85aa46d8d3c73bbf5abbf4a9a but for all release notes. The following XSLT program will do that:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">

  <!-- Identity template: copies everything as-is -->
  <xsl:template match="@*|node()">
    <xsl:copy>
      <xsl:apply-templates select="@*|node()"/>
    </xsl:copy>
  </xsl:template>

  <!-- Match <description> elements within <release> -->
  <xsl:template match="/component/releases/release/description[not(@translate) and not(@translatable)]">
    <xsl:copy>
      <!-- Add the translate="no" attribute -->
      <xsl:attribute name="translate">no</xsl:attribute>
      <xsl:apply-templates select="@*|node()"/>
    </xsl:copy>
  </xsl:template>

</xsl:stylesheet>
```

You can apply it with

```
xsltproc foo.xslt data/metainfo/org.gnome.Software.xml.in \
  > data/metainfo/org.gnome.Software.xml.in.tmp \
  && \
  mv data/metainfo/org.gnome.Software.xml.in.tmp data/metainfo/org.gnome.Software.xml.in
```

I have [proposed that Damned Lies could do something similar](https://gitlab.gnome.org/Infrastructure/damned-lies/-/issues/473#note_2041738) to extract at least the last few releases' notes for translations.

2. Or we could do the same thing as GNOME Damned Lies does, using a patched copy of the `metainfo.its` rule. [This](https://github.com/endlessm/eos-build/blob/master/ci-integration/jenkins-obs-buildscript#L233) is where strings get extracted. We would have to place the file somewhere in `xgettext`s search path. I think this is more work than carrying forward (a spiritual successor of) the "delete em all" patch.

wjt:
> # Translations
> # Did the new languages we need make it upstream?
> 
> drop baee429961bb72527510a993c0d1167d8a50b64d Add translation support for several new languages

According to a quick check on eos5.1 the overwhelming majority of these are 0% translated. This is the same thing I found in endlessm/gnome-initial-setup:

```
In [13]: for filename in sorted(glob.glob("po/*.po")):
    ...:     pofile = polib.pofile(filename)
    ...:     if pofile.percent_translated() == 0:
    ...:         print(os.path.splitext(os.path.basename(filename))[0])
    ...: 
am
an
ast
az
bn_IN
bo
br
crh
csb
cy
dz
et
fil
fy
gu
ha
hy
ia
ig
io
kn
ks
ku
ky
la
li
ln
mai
mg
mi
mk
mn
mr
my
nds
nn
nso
or
ps
rw
si
sq
ta
tg
tk
ug
uz
wa
xh
yi
yo
zu
```

I suggest preserving only the additions to LINGUAS that are not on this list (if any).

wjt:
pick 8ab012b0db0a052d3f29e39f63830ac8ac73222d appstream: Override `<id>` with `<launchable type="desktop-id">` while parsing appdata files

I would be curious if this is still needed. Looking at the ticket linked in the commit message it was related to apps on the Installed tab. But since 4.0.0 (eos-ostree-builder commit d1a3d633d57e04271cf7f2e57153488be35115a9) we have generated an OS appstream data XML which I think should make this unnecessary.

meson_options.txt
```diff
@@ -1,4 +1,4 @@
-option('tests', type : 'boolean', value : true, description : 'enable tests')
+option('tests', type : 'boolean', value : false, description : 'enable tests')
```
wjt, line 1: I still find this quite sad and feel like we ought to be able to do better. But the upstream `.gitlab-ci/run-tests.sh` does the following which also feels like a lot:

```bash
# FIXME: The tests should be isolated and use mock services so they do not
# require a functioning system bus. This will have to do for now though.
sudo mkdir -p /run/dbus
sudo mkdir -p /var
sudo ln -s /var/run /run
sudo dbus-daemon --system --fork
#sudo /usr/lib/polkit-1/polkitd --no-debug &
#sudo /usr/libexec/fwupd/fwupd --verbose &

# FIXME: Running the flatpak tests as root means the system helper doesn’t
# need to be used, which makes them run a lot faster.
sudo \
dbus-run-session -- \
meson test \
        -C _build \
        --timeout-multiplier ${MESON_TEST_TIMEOUT_MULTIPLIER} \
        --no-suite flaky \
        --print-errorlogs \
        "$@"
```

meson_options.txt
```diff
@@ -1,25 +1,25 @@
...
 option('gtk_doc', type : 'boolean', value : true, description : 'enable API reference')
-option('hardcoded_curated', type : 'boolean', value : true, description : 'enable installation of default curated apps list')
+option('hardcoded_curated', type : 'boolean', value : false, description : 'enable installation of default curated apps list')
 option('default_featured_apps', type : 'boolean', value : true, description : 'enable installation of default featured apps list')
-option('mogwai', type : 'boolean', value : false, description : 'enable metered data support using Mogwai')
+option('mogwai', type : 'boolean', value : true, description : 'enable metered data support using Mogwai')
 option('sysprof', type : 'feature', value : 'auto', description : 'enable sysprof-capture support for profiling')
 option('profile', type : 'string', value : '', description : 'Build with specified application ID')
-option('soup2', type : 'boolean', value : false, description : 'build with libsoup2')
+option('soup2', type : 'boolean', value : true, description : 'build with libsoup2')
```
wjt, line 25: We have soup3 on master, we should drop this hunk.

meson_options.txt
```diff
@@ -1,25 +1,25 @@
...
-option('eos_updater', type : 'boolean', value : false, description : 'enable eos-updater support')
-option('fwupd', type : 'boolean', value : true, description : 'enable fwupd support')
+option('eos_updater', type : 'boolean', value : true, description : 'enable eos-updater support')
+option('fwupd', type : 'boolean', value : false, description : 'enable fwupd support')
 option('flatpak', type : 'boolean', value : true, description : 'enable Flatpak support')
 option('malcontent', type : 'boolean', value : true, description : 'enable parental controls support using libmalcontent')
 option('rpm_ostree', type : 'boolean', value : false, description : 'enable rpm-ostree support')
 option('webapps', type : 'boolean', value : true, description : 'enable webapps support')
 option('hardcoded_foss_webapps', type : 'boolean', value : true, description : 'enable inclusion of a default set of installable FOSS webapps')
 option('hardcoded_proprietary_webapps', type : 'boolean', value : true, description : 'enable inclusion of a default set of installable proprietary webapps')
```
wjt, line 14: I don't think we want this one to be `true`, do we flip it to False in a future patch?

plugins/eos-blocklist/gs-plugin-eos-blocklist.c
```diff
@@ -0,0 +1,547 @@
...
+				return TRUE;
+			}
+		}
+	}
+
+	return FALSE;
+}
+
+static gboolean
+locale_is_compatible (GsPluginEosBlocklist *self,
```
wjt, line 295: I wonder if we could take a different approach here:

1. Remove all our unmaintained locale-specific knowledge apps from https://github.com/endlessm/gnome-software-data/blob/master/app-info/eos-extra.txt so that they don't appear in the carousel or Editor's Choice
2. Then just remove this entire plugin and allow 10 different encyclopedias in 10 different languages to appear

Maybe that's a separate discussion.

plugins/eos/gs-plugin-eos.c
```diff
@@ -0,0 +1,94 @@
...
+}
+
+gboolean
+gs_plugin_launch (GsPlugin *plugin,
+		  GsApp *app,
+		  GCancellable *cancellable,
+		  GError **error)
+{
+	/* if the app is one of the system ones, we simply launch it through the
+	 * plugin's app launcher */
```
wjt, line 186: I'm curious what this plugin actually does. If we remove it, can we still launch built-in apps from their page in GNOME Software?

plugins/eos/gs-plugin-eos.c
```diff
@@ -67,6 +67,56 @@ gs_plugin_adopt_app (GsPlugin *plugin, GsApp *app)
...
+		return;
+
+	/* Hide non-installed apt packages, as they can’t actually be installed.
+	 * The installed ones are pre-installed in the image, and can’t be
+	 * removed. We only allow flatpaks to be removed. */
+	if (!gs_app_is_installed (app)) {
+		gs_app_add_quirk (app, GS_APP_QUIRK_HIDE_EVERYWHERE);
+	} else {
+		gs_app_add_quirk (app, GS_APP_QUIRK_COMPULSORY);
+	}
```
wjt, line 120: OK, this plugin does end up doing something more interesting than "launch an app".

plugins/eos-updater/gs-plugin-eos-updater.c
```diff
@@ -767,6 +769,50 @@ gs_plugin_eos_updater_dispose (GObject *object)
...
+static void
+report_parental_controls_metric (GCancellable *cancellable)
+{
+	g_autoptr(GDBusConnection) system_bus = NULL;
+	g_autoptr(MctManager) manager = NULL;
+	g_autoptr(MctAppFilter) filter = NULL;
+	g_autoptr(GError) metrics_error = NULL;
+	gboolean parental_controls_enabled = FALSE;
+
+#define GS_PARENTAL_CONTROLS_USAGE_EVENT "c227a817-808c-4fcb-b797-21002d17b69a"
```
wjt, line 786: I'd love to move this somewhere else that runs on a timer. But we can carry this forward for now.

plugins/flatpak/gs-flatpak.c
```diff
@@ -2139,6 +2190,18 @@ gs_flatpak_add_updates (GsFlatpak *self,
...
+		 * extension whose update should get prioritized, then we need to
+		 * set its main app priority to represent it) */
+		if (main_app != app) {
+			UpdatePriority main_app_update_priority =
+				get_app_update_priority (main_app);
+			if (update_priority < main_app_update_priority)
+				update_priority = main_app_update_priority;
+		}
+		set_app_update_priority (main_app, update_priority);
+
```
wjt, line 2204: I think we could drop this patch. Looking at the rationale in the commit message:

> For EOS, the priority for updates should be, from more important to less important:
>
> - Apps that have a "Content" runtime, which means they are likely news apps and need to be updated often;

We no longer update such apps regularly, particularly not news apps

> - OS updates;

OK I guess...

> - Apps from Endless' domain (since those are usually Knowledge Apps that also get updates with new content);

We no longer regularly update these apps.

> - Any other apps;
> - Runtimes.

Personally I would be happy with OS > app > runtime. But updating an app will, in practice, update its runtime(s), so…

plugins/eos-blocklist/gs-plugin-eos-blocklist.c
wjt: https://phabricator.endlessm.com/T33450 is the separate discussion I guess. OK let's carry this forwards.

plugins/flatpak/gs-flatpak.c
dylanmccall: Yeah, I was feeling skeptical about this one as well, seeing as, indeed, the first half of its rationale is no longer a thing :) I'll drop it.

dylanmccall:
> pick [8ab012b](https://github.com/endlessm/gnome-software/commit/8ab012b0db0a052d3f29e39f63830ac8ac73222d) appstream: Override `<id>` with `<launchable type="desktop-id">` while parsing appdata files
> 
> I would be curious if this is still needed. Looking at the ticket linked in the commit message it was related to apps on the Installed tab. But since 4.0.0 (eos-ostree-builder commit d1a3d633d57e04271cf7f2e57153488be35115a9) we have generated an OS appstream data XML which I think should make this unnecessary.

Yep, this seems like it should be unnecessary. I'll drop this patch here and I'll add a note in the Phabricator task to make sure we haven't regressed https://phabricator.endlessm.com/T27779.

dylanmccall:
> > drop [baee429](https://github.com/endlessm/gnome-software/commit/baee429961bb72527510a993c0d1167d8a50b64d) Add translation support for several new languages
> 
> According to a quick check on eos5.1 the overwhelming majority of these are 0% translated.
> […]
> I suggest preserving only the additions to LINGUAS that are not on this list (if any).

Neat! The only one of these is `en_CA`, which you'd alluded to earlier with gnome-tour :b I'll remove it this time. (I checked and there is one single occurrence of the word "humor" instead of "humour", which is terribly upsetting but really doesn't need to be fixed by us). Happily, there _is_ a [Burmese translation upstream](https://gitlab.gnome.org/GNOME/gnome-software/-/commit/ea7924d67a9bcabc9774edce9b0d5a6d19ab517d) now.

meson_options.txt
dylanmccall: This looks the same as we have for [meson_options.txt in eos5.1](https://github.com/endlessm/gnome-software/blob/eos5.1/meson_options.txt) :thinking:

wjt: It is flipped to false in `debian/rules`, which is used on OBS but not Jenkins. https://github.com/endlessm/gnome-software/commit/0e8949841daab47f395548885757baafbe63d50a

wjt:
I was about to make a joke about the en_GB translation existing only to make that same change but I checked and actually it also changes “dialog” to “dialogue” (yes please) and “email” to “e-mail” (come on).

meson_options.txt
dylanmccall: Oh, that line! Github's had me looking at line 7 instead :b I see we did that in https://phabricator.endlessm.com/T34569. It does seem a bit odd, so I went ahead and set the option here while we have the opportunity.

dylanmccall:
I'm feeling a bit uneasy about my rebase of our changes to plugins/flatpak/gs-plugin-flatpak.c here, so I'm leaving those fixup commits as they are for now.

plugins/flatpak/gs-plugin-flatpak.c
```diff
@@ -1825,7 +1825,7 @@ gs_plugin_flatpak_update (GsPlugin *plugin,
 	}
 
 	/* build and run transaction */
-	transaction = _build_transaction (plugin, flatpak, interactive, cancellable, error);
+	transaction = _build_transaction (plugin, flatpak, GS_FLATPAK_ERROR_MODE_STOP_ON_FIRST_ERROR, interactive, cancellable, error);
```
wjt, line 1828: I think this would be a regression. We really want Software to apply as many updates as it can, even if some of them fail.

Otherwise, you can get into a state where you have a vast number of app updates pending; but the first one happens to always fail, and so you never get updates. But each time, Software will download them all before applying them, wasting bandwidth and disk space. Particularly when the reason that some of the updates fail is “you don't have enough disk space” you really want to apply some rather than none, because updating apps temporarily uses additional disk space.

https://gitlab.gnome.org/GNOME/gnome-software/-/issues/1929

wjt:
> drop 8ab012b0db0a052d3f29e39f63830ac8ac73222d appstream: Override <id> with <launchable type="desktop-id"> while parsing appdata files
> #- We believe this is no longer necessary.

https://github.com/endlessm/gnome-software/pull/619/commits/e3b56dd0eabe4ceccf48573a5489ffc8bbc25ae2 agrees we can drop this.

> pick d001d2881a81938ff9b0804a42d75769210d98bd gs-prefs-dialog: Hide automatic updates setting when using Mogwai

https://github.com/endlessm/gnome-software/pull/619/commits/5f5a97cb03b7d0844f17a72623ca0ac87d60c4cb suggests dropping this one.

> pick d8eecfd5af782caa03479a37c3d729ca89bc4710 Do not invalidate OS upgrades in the Updates page

https://github.com/endlessm/gnome-software/pull/619/commits/36ffc64474c63dec8915317ce577ef618f6a50f1 also discusses cherry-picking an upstream patch that  will be in 46, then reapplying some subset of this change.

I suggest we land this branch as-is while looking into the details of those.

PR #622, opened 2024-03-15
Rebase on GNOME Software 45 (packaging)

dylanmccall:

-----

Please merge this along with:

- https://github.com/endlessm/gnome-software/pull/621


dylanmccall:
bot, rebuild this please

debian/patches/Fix-loading-OS-AppStream-catalog-data-if-it-is-available-.patch
```diff
@@ -0,0 +1,30 @@
+From: Matthias Klumpp <matthias@tenstral.net>
```
wjt, line 1: Please drop this patch – we do not ship YAML-formatted AppStream catalogues.

debian/patches/debian/use-software-properties-gtk.patch
```diff
@@ -0,0 +1,25 @@
...
+--- a/src/gs-shell.c
++++ b/src/gs-shell.c
+@@ -2339,6 +2339,10 @@ gs_shell_show_sources (GsShell *shell)
+ {
+ 	GtkWidget *dialog;
+ 
++	/* use if available */
++	if (g_spawn_command_line_async ("software-properties-gtk", NULL))
++		return;
++
```
wjt, line 22: It's not available. Please remove this patch.

debian/patches/series
```diff
@@ -1 +1,2 @@
-01_details-Use-custom-icon-for-verified-developers.patch
+debian/use-software-properties-gtk.patch
+Fix-loading-OS-AppStream-catalog-data-if-it-is-available-.patch
```
wjt, line 2: Remove this file together with the patches

debian/rules
```diff
@@ -33,20 +33,6 @@ ifeq ($(DEB_HOST_ARCH_OS), linux)
...
-	-Dexternal_appstream=true \
-	-Dfwupd=false \
-	-Dgudev=false \
-	-Dhardcoded_curated=false \
-	-Dmalcontent=true \
-	-Dmogwai=true \
-	-Dpackagekit=false \
-	-Dsoup2=true \
-	-Dhardcoded_foss_webapps=false \
-	-Dhardcoded_proprietary_webapps=false
```
wjt, line 48: I don't think it's correct to drop these flags. Higher up in the file, packagekit support (for example) is enabled, which we do not want.

Some of them may be redundant with the code-branch change to `meson_options.txt`, so it could perhaps be slimmed down.

wjt (changes requested):

dylanmccall:
Okay, I addressed those things. I tweaked the rules so we're still only specifying the configure flags which _need_ to be specified, but putting things back that have been changed to something else.

wjt (approved):

PR #623, opened 2024-04-02
gs-app-context-bar: Hide the Safety and Hardware tiles

dylanmccall:
These tend to be either alarming or confusing for Endless OS users.

https://phabricator.endlessm.com/T35252


starnight (approved):
Thanks!

PR #624, opened 2024-04-08
Revert "New translations from Transifex"

manuq:
This reverts commit 6fcd33f967ca712a5a1197ee02b1a599312e1d93.

https://phabricator.endlessm.com/T35218


wjt:
We don't need to merge the revert. We just need to revert locally so we can push the pristine GNOME translations into Transifex.

PR #625, opened 2024-04-18
flatpak: Count all references to task in block_scheduler_schedule_cb

dylanmccall:
In `block_scheduler_schedule_cb`, a `GCancellable::cancelled` signal handler was connected to a tasks's cancellable, with a pointer to the task itself as data. However, this was done without incrementing the reference count for the task, resulting in an error due to the task being prematurely freed when `g_cancellable_disconnect` was called.

In addition to handling that case, we will change `block_check_cb` to increment the task's reference count while it is in use.

https://phabricator.endlessm.com/T35306

### DIFF
--- a/lib/gs-metered.c
+++ b/lib/gs-metered.c
@@ -224,7 +224,7 @@ block_scheduler_schedule_cb (GObject      *source_object,
                              gpointer      user_data)
 {
 	MwscScheduler *scheduler = MWSC_SCHEDULER (source_object);
-	g_autoptr(GTask) task = g_steal_pointer (&user_data);
+	g_autoptr(GTask) task = G_TASK (user_data);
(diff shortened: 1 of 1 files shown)


wjt:
OK, interesting that this didn't work – this is roughly what I had in mind too!

dylanmccall:
Oh never mind, it works fine! Just that the gnome-software which autostarts on said real system is `/usr/bin/gnome-software`. Unfortunately I haven't been able to run the tests for gnome-software, and I rather wish I could, but I think I'll make a separate Phabricator task about that one.

wjt (approved):
As I said yesterday, I wrote a semantically identical but untested  patch, so I'm predisposed to accept this. :)

Please submit upstream too.

PR #626, opened 2024-04-19
flatpak: Count all references to task in block_scheduler_schedule_cb (eos6.0)

dylanmccall:
https://phabricator.endlessm.com/T35306


wjt:
We should be able to fast-forward and keep the two branches level

wjt:
OK, I see that we can't fast-forward, my mistake! I merged an equivalent change and pushed the result to both eos6.0 and master so that the two branches are level. Apart from general tidiness, the reason this is desirable is actually the same reason the branches had diverged: if they are level, then new translations [get pushed to both](https://github.com/endlessm/eos-build/blob/master/ci-integration/jenkins-transifex-pull#L85-L960), rather than just to the stable branch.

